PR #1, opened 2019-01-07
A basic Play app

tomrf1:
A new app that will eventually provide the tools for configuration support-frontend and frontend tests.
No client-side code yet.

1. Loads config from parameter store, or from disk if running locally.
2. /index is behind google auth

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -10,7 +10,7 @@ resolvers += "Guardian Platform Bintray" at "https://dl.bintray.com/guardian/pla
 libraryDependencies ++= Seq(
   "com.typesafe.scala-logging" %% "scala-logging" % "3.7.2",
   "com.gu" %% "play-googleauth" % "0.7.7",
-  "com.gu" %% "simple-configuration-s3" % "1.4.1",
+  "com.gu" %% "simple-configuration-ssm" % "1.4.1",
   ws
 )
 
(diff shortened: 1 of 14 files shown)


cloudformation.yaml
```diff
@@ -0,0 +1,194 @@
...
+          Action:
+          - sts:AssumeRole
+      Policies:
+      - PolicyName: PrivateBucket
+        PolicyDocument:
+          Version: 2012-10-17
+          Statement:
+          - Effect: Allow
+            Action: s3:GetObject
+            Resource: !Sub arn:aws:s3:::membership-private/${Stage}/support-admin-console.private.conf
```
joelochlann: personally I'd much rather use secrets manager/parameter store rather than a private config file in S3 - the parameter store has worked very well for us in the Payment API, and secrets manager is even better I believe (allows for auto-rotation and such)

build.sbt
```diff
@@ -0,0 +1,28 @@
+
+name := "support-admin-console"
+
+version := "1.0-SNAPSHOT"
+
+scalaVersion := "2.12.8"
+
+resolvers += "Guardian Platform Bintray" at "https://dl.bintray.com/guardian/platforms"
```
joelochlann: oh interesting, are we using Bintray by default these days? or is this just for certain dependencies?

tomrf1: It's for `simple-configuration`, but not sure why bintray was chosen

PR #3, opened 2019-01-17
Webpack typescript react

joelochlann:
And some routing


.editorconfig
```diff
@@ -0,0 +1,13 @@
+; This file is for unifying the coding style for different editors and IDEs
```
tomrf1: 💯

PR #4, opened 2019-01-18
Endpoints for getting and updating support-frontend config in s3

tomrf1:
The commit history here shows a bit of a journey (play forms, elm frontend, and finally just a json api). We realised that elm is really overkill for such a simple page, and are going with a [react frontend ](https://github.com/guardian/support-admin-console/pull/3)for now (but not ruling out elm forever!).

This PR introduces GET and POST endpoints for the config in s3. The json is validated in both directions against the scala model.
The server also checks the version of the s3 file before updating it, to ensure it doesn't overwrite new updates.
This app is going to do a lot more updating of json files in s3, so this logic will be used quite a bit.


app/services/S3.scala
```diff
@@ -0,0 +1,94 @@
...
+      val version = s3Object.getObjectMetadata.getVersionId
+
+      val stream = s3Object.getObjectContent
+      val raw: String = scala.io.Source.fromInputStream(stream).mkString
+      stream.close()
+
+      Right[String,RawVersionedS3Data](VersionedS3Data(raw, version))
+
+    } catch {
+      case NonFatal(e) =>
```
johnduffell: if you use the scala Try rather than the try, it handles the non fatal stuff for you

johnduffell: also I'm pretty sure it has a toEither as well

app/services/S3.scala
```diff
@@ -0,0 +1,94 @@
...
+
+object S3 extends S3Client with StrictLogging {
+
+  private val s3Client: AmazonS3 = AmazonS3ClientBuilder
+    .standard()
+    .withRegion(Aws.region)
+    .withCredentials(Aws.credentialsProvider)
+    .build()
+
+  def get(bucket: String, key: String)(implicit ec: ExecutionContext): Future[Either[String,RawVersionedS3Data]] = Future {
```
johnduffell: wrapping a sync client in a future blocks an extra thread so can cause perf issues, this is a low traffic app so it won't cause anything in reality but it's better practice to use the async and complete a promise to avoid that.

tomrf1: it seems the only way to do async s3 access is [TransferManager](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/index.html?com/amazonaws/services/s3/transfer/TransferManager.html), which looks a bit tedious.
Anyway, thread use is never going to be an issue for this app, but if it was I guess we could [configure the thread pool](https://www.playframework.com/documentation/2.6.x/ThreadPools#Highly-synchronous) differently

app/wiring/AppLoader.scala
```diff
@@ -14,19 +14,22 @@ class AppLoader extends ApplicationLoader with StrictLogging {
...
+    val loadedConfig = ConfigurationLoader.load(identity) {
+      case AwsIdentity(app, stack, stage, _) => SSMConfigurationLocation(s"/$app/$stage")
+      case DevIdentity(app) =>
+        FileConfigurationLocation(new File(s"/etc/gu/support-admin-console.private.conf"))  //assume conf is available locally
+    }
+
+    val newContext = context.copy(initialConfiguration = context.initialConfiguration ++ Configuration(loadedConfig))
 
-      context.copy(initialConfiguration = context.initialConfiguration ++ Configuration(loadedConfig))
+    val stage = identity match {
```
johnduffell: since stage is used in so many places  it would be worth making into a value class to avoid mistakes that can't be found by the type checker
```scala
case class Stage(value: String) extends AnyVal
```

johnduffell:
thanks for that, my comments are quite general so feel free to bear them in mind or otherwise in future! 👍

PR #6, opened 2019-01-18
Render and save switches with server data

joelochlann:
Using lenses could make the all object getting/setting nicer - there will be lots more of that in this application so a generic approach could be worth it.

But it works!

### DIFF
--- a/public/src/components/switchboard.tsx
+++ b/public/src/components/switchboard.tsx
@@ -4,12 +4,61 @@ import set from 'lodash/set';
 
 import FormLabel from '@material-ui/core/FormLabel';
 import FormControl from '@material-ui/core/FormControl';
-import FormGroup from '@material-ui/core/FormGroup';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
 import Switch from '@material-ui/core/Switch';
 import Button from '@material-ui/core/Button';
 import SaveIcon from '@material-ui/icons/Save';
 
+
+interface AmountsFromServer {
+  // TBC, just a placeholder for now
+  US: number[],
+  GB: number[],
+  AU: number[],
+  // ... and so on
+}
+
+enum ContributionType {
+  OneOff,
+  Monthly,
+  Annual
+}
+
+interface ContributionTypesFromServer {
+  // just a placeholder for now
+  US: ContributionType[],
+  GB: ContributionType[],
+  AU: ContributionType[],
+}
+
+enum SwitchState {
+  On, Off
+}
+
+interface SwitchesFromServer {
+  oneOffPaymentMethods: {
+    stripe: SwitchState,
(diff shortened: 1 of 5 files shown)


public/src/components/switchboard.tsx
```diff
@@ -0,0 +1,147 @@
...
+    stripeOneOff: safelyExtractSwitchValue(json, 'value.switches.oneOffPaymentMethods.stripe'),
+    payPalOneOff: safelyExtractSwitchValue(json, 'value.switches.oneOffPaymentMethods.payPal'),
+    stripeRecurring: safelyExtractSwitchValue(json, 'value.switches.recurringPaymentMethods.stripe'),
+    payPalRecurring: safelyExtractSwitchValue(json, 'value.switches.recurringPaymentMethods.payPal'),
+    directDebitRecurring: safelyExtractSwitchValue(json, 'value.switches.recurringPaymentMethods.directDebit')
+  };
+}
+
+export class Switchboard extends React.Component {
+  state: Switches;
```
tomrf1: nice, does this mean it'll fail to compile if you assign another type to `this.state`?

public/src/components/switchboard.tsx
```diff
@@ -0,0 +1,147 @@
...
+  saveSwitches = () => {
+    const newState = {...this.previousStateFromServer};
+
+    // actually I'm not sure I like this technique since it bypasses the type system
+    // though in this case it doesn't matter since it's just an object type
+    set(newState, 'value.switches.oneOffPaymentMethods.stripe', this.state.stripeOneOff ? 'On' : 'Off');
+    set(newState, 'value.switches.oneOffPaymentMethods.payPal', this.state.payPalOneOff ? 'On' : 'Off');
+    set(newState, 'value.switches.recurringPaymentMethods.stripe', this.state.stripeRecurring ? 'On' : 'Off');
+    set(newState, 'value.switches.recurringPaymentMethods.payPal', this.state.payPalRecurring ? 'On' : 'Off');
+    set(newState, 'value.switches.recurringPaymentMethods.directDebit', this.state.directDebitRecurring ? 'On' : 'Off');
```
tomrf1: is this form going to allow us to set other switches? Currently the file also has `optimize` and `experiments`
If this component is going to update all settings in this file then it might be worth defining the type of this object properly

public/src/components/switchboard.tsx
joelochlann: yep. it also enforces that you set a value for it in the constructor

public/src/components/switchboard.tsx
tomrf1: I may now disagree with my statement from yesterday - if it means a load of boilerplate code then the server-side validation is enough!

PR #7, opened 2019-01-22
[MERGING INTO A BRANCH] Update switches in a typesafe way

joelochlann:
I had a look at Ramda's lenses but they're no more typesafe than the lodash `set` function I was using before (though they have the advantage of composability - e.g. we could pass around a lens to zoom into `value` from the server data)

One nice thing about this `immutability-helper` is it will produce type errors if you try and set things that don't correspond to the object types. Also its syntax for drilling into an object it pretty intuitive, I think

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,5 @@
 {
   "devDependencies": {
-    "@types/lodash": "^4.14.120",
     "@types/material-ui": "^0.21.5",
     "@types/react-dom": "^16.0.11",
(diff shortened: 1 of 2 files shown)


joelochlann:
https://github.com/JohnWeisz/TypedJSON looks really interesting - we could try adding that later if we want to make the actual JSON deserialisation safer (though I think the real safety probably lies simply in making sure our Scala and TypeScript models stay in sync)

PR #12, opened 2019-02-01
Require specific Google Groups for login

joelochlann:
To Do

- [x] Upload new dev config file to S3
- [x] Add new config to parameter store

### DIFF
--- a/app/controllers/Login.scala
+++ b/app/controllers/Login.scala
@@ -1,12 +1,18 @@
 package controllers
 
-import com.gu.googleauth.{GoogleAuthConfig, LoginSupport}
+import com.gu.googleauth.{GoogleAuthConfig, GoogleGroupChecker, LoginSupport}
 import play.api.libs.ws.WSClient
 import play.api.mvc._
 
 import scala.concurrent.ExecutionContext
 
-class Login(val authConfig: GoogleAuthConfig, val wsClient: WSClient, components: ControllerComponents)(implicit executionContext: ExecutionContext)
+class Login(
+  val authConfig: GoogleAuthConfig,
+  val wsClient: WSClient,
+  requiredGoogleGroups: Set[String],
+  googleGroupChecker: GoogleGroupChecker,
+  components: ControllerComponents
+)(implicit executionContext: ExecutionContext)
   extends AbstractController(components) with LoginSupport {
 
   /**
@@ -28,7 +34,7 @@ class Login(val authConfig: GoogleAuthConfig, val wsClient: WSClient, components
    * Looks up user's identity via Google
    */
   def oauth2Callback: Action[AnyContent] = Action.async { implicit request =>
-    processOauth2Callback()
+    processOauth2Callback(requiredGoogleGroups, googleGroupChecker)
   }
 
   def logout: Action[AnyContent] = Action { implicit request =>
--- a/app/wiring/AppComponents.scala
+++ b/app/wiring/AppComponents.scala
@@ -1,6 +1,8 @@
 package wiring
 
-import com.gu.googleauth.{AntiForgeryChecker, AuthAction, GoogleAuthConfig}
+import com.amazonaws.services.s3.AmazonS3URI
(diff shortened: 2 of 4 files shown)


app/wiring/AppComponents.scala
```diff
@@ -21,13 +24,33 @@ class AppComponents(context: Context, stage: String) extends BuiltInComponentsFr
     GoogleAuthConfig(clientId, clientSecret, redirectUrl, domain, antiForgeryChecker = AntiForgeryChecker.borrowSettingsFromPlay(httpConfiguration))
   }
 
+  private val googleGroupChecker = {
```
joelochlann: worth noting we have a try/catch in `AppLoader` which catches and logs any app startup errors which occur in `AppComponents`

app/controllers/Login.scala
```diff
@@ -1,12 +1,18 @@
...
 import scala.concurrent.ExecutionContext
 
-class Login(val authConfig: GoogleAuthConfig, val wsClient: WSClient, components: ControllerComponents)(implicit executionContext: ExecutionContext)
+class Login(
+  val authConfig: GoogleAuthConfig,
+  val wsClient: WSClient,
+  requiredGoogleGroups: Set[String],
+  googleGroupChecker: GoogleGroupChecker,
+  components: ControllerComponents
+)(implicit executionContext: ExecutionContext)
```
johnduffell: this class has a lot of deps, and the two new ones are only used by a single method.  It would make more sense to separate as necessary so in general things only get what they need.
We could have a Login class with everything else, and a OAuth2 class with a def callback.

not essential for the PR as it won't break anything though!

app/wiring/AppComponents.scala
```diff
@@ -21,13 +24,33 @@ class AppComponents(context: Context, stage: String) extends BuiltInComponentsFr
...
+      .getObject(s3URI.getBucket, s3URI.getKey)
+      .getObjectContent
+
+    val googleServiceAccountCredential = GoogleCredential.fromStream(stream)
+    stream.close()
+
+    val serviceAccount = GoogleServiceAccount(
+      googleServiceAccountCredential.getServiceAccountId,
+      googleServiceAccountCredential.getServiceAccountPrivateKey,
+      configuration.get[String]("googleAuth.serviceAccount.impersonatedUser"),
```
johnduffell: what's an impersonatedUser ?

joelochlann: https://github.com/guardian/play-googleauth#implement-googlegroups-based-access-control-using-the-directory-api

app/wiring/AppComponents.scala
```diff
@@ -21,13 +24,33 @@ class AppComponents(context: Context, stage: String) extends BuiltInComponentsFr
...
   }
 
+  private val googleGroupChecker = {
+    val s3URI = new AmazonS3URI(configuration.get[String]("googleAuth.serviceAccount.certificateS3Path"))
+    val stream = S3.s3Client
+      .getObject(s3URI.getBucket, s3URI.getKey)
+      .getObjectContent
+
+    val googleServiceAccountCredential = GoogleCredential.fromStream(stream)
+    stream.close()
```
johnduffell: will this be closed if there's an exception (and if not does it matter or will the app fail the healthcheck anyway?)

app/wiring/AppComponents.scala
joelochlann: added a link in a comment

johnduffell:
looks good to me! 👍 good work!

app/wiring/AppComponents.scala
joelochlann: ah good point. if anything goes wrong in this class, the app won't start so it will fail healthcheck. we'll get logging on the exception in CloudWatch to diagnose

johnduffell:
re your TODO list, it's worth having an IT test for downloading the config and validating it in all environments.  This does require credentials and it's hard to automate, but at least it doesn't rely on not forgetting to add the config in the right place.

PR #16, opened 2019-03-26
Contribution types page

tomrf1:
Allows us to manage the configured contribution types (in s3)

Screenshot 2019-03-28 at 12 43 06

### DIFF
--- a/public/src/components/contributionTypes.tsx
+++ b/public/src/components/contributionTypes.tsx
@@ -24,19 +24,31 @@ interface ContributionTypeSetting {
   isDefault?: boolean
 }
 
-interface ContributionTypes {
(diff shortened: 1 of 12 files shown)


public/src/components/contributionTypes.tsx
```diff
@@ -0,0 +1,293 @@
...
+      NZDCountries: []
+    };
+    this.previousStateFromServer = null;
+  }
+
+  componentWillMount(): void {
+    this.fetchStateFromServer();
+  }
+
+  fetchStateFromServer(): void {
```
joelochlann: before we add a third page we should refactor this duplication

tomrf1: I've just moved the common GET/POST logic to /utils/

PR #17, opened 2019-04-02
Make amounts configurable

tomrf1:
Screenshot 2019-04-18 at 11 33 03

### DIFF
--- a/app/models/Amounts.scala
+++ b/app/models/Amounts.scala
@@ -1,0 +1,15 @@
+package models
+
+case class Amount(value: String, isDefault: Option[Boolean])
+
+case class Amounts(ONE_OFF: List[Amount], MONTHLY: List[Amount], ANNUAL: List[Amount])
+
+case class AmountsRegions(
+  GBPCountries: Amounts,
+  UnitedStates: Amounts,
+  EURCountries: Amounts,
+  AUDCountries: Amounts,
+  International: Amounts,
+  NZDCountries: Amounts,
+  Canada: Amounts
+)
--- a/public/src/components/amounts/amount.tsx
+++ b/public/src/components/amounts/amount.tsx
@@ -26,6 +26,7 @@ interface Props extends WithStyles<typeof styles> {
 
 interface AmountState {
   amount: Amount,
+  //anchor determines where on the page the right-click menu should pop-up
   anchor: HTMLElement | null
 }
 
--- a/public/src/components/amounts/amount.tsx
+++ b/public/src/components/amounts/amount.tsx
@@ -60,13 +60,11 @@ class AmountComponent extends React.Component<Props, AmountState> {
         <Button
           className={this.props.classes.amount + (this.props.amount.isDefault ? ` ${this.props.classes.isDefault}` : '') }
           onClick={this.onButtonClick}
-          aria-owns={this.state.anchor ? 'simple-menu' : undefined}
         >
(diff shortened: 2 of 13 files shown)


app/models/Amounts.scala
```diff
@@ -0,0 +1,15 @@
+package models
+
+case class Amount(value: String, isDefault: Option[Boolean])
+
+case class Amounts(ONE_OFF: List[Amount], MONTHLY: List[Amount], ANNUAL: List[Amount])
```
joelochlann: so UI-wise, this is independent of the contribution types as set in the JSON file? if you've only got "one-off" enabled for the US, say, you'll still be able to set amounts for all three types

tomrf1: Correct, contributionTypes and amounts are still configured separately. This is a bit confusing and perhaps the 2 config files can be unified later

public/src/components/amounts/amount.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+interface Props extends WithStyles<typeof styles> {
+  amount: Amount,
+  deleteAmount: (amountToDelete: string) => void,
+  makeDefault: (defaultAmount: string) => void
+}
+
+interface AmountState {
+  amount: Amount,
+  anchor: HTMLElement | null
```
joelochlann: a comment on what the `anchor` is, and why it needs to be tracked in the state, would be helpful

public/src/components/amounts/amountInput.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+const styles = ({ palette, spacing, mixins }: Theme) => createStyles({
+  amountInput: {
+    display: 'inline',
+    marginLeft: 'auto',
+    paddingLeft: '8px',
+    borderLeft: `1px dotted ${palette.grey['300']}`
+  },
+  amountTextField: {
+    width: '40px'
```
joelochlann: having played around with it, this is a bit narrow and cuts off the numbers... maybe 80px?

public/src/components/amounts/amounts.tsx
```diff
@@ -0,0 +1,238 @@
...
+  ContributionTypeName: {
+    fontWeight: 'bold',
+    marginTop: 'auto',
+    minWidth: '88px'
+  }
+});
+
+interface Props extends WithStyles<typeof styles> {}
+
+class AmountsComponent extends React.Component<Props, AmountsRegions> {
```
joelochlann: I think we could rename these `...Component` ones across the board for clarity, since component is such a general term (everything is a component). Reading this through initially I was getting confused between what were the granular elements and higher-level parents - maybe `AmountsForm`, `ContributionTypesForm`, etc would be clearer?

joelochlann: or maybe just the form ones, to `AmountsFormComponent`, `ContributionTypesFormComponent`?

joelochlann:
> And also if there's a nicer UI for contribution types - the radio buttons for set default isn't the most intuitive to me (maybe just a select dropdown would be better? so it displays the name of those one that's selected as default)

Actually that point is greatly ameliorated by this PR so maybe not so relevant:
https://github.com/guardian/support-admin-console/pull/19#pullrequestreview-232107164

tomrf1:
@joelochlann the UI could definitely be improved, especially if it's going to handle a/b tests as well

PR #23, opened 2019-07-12
Epic test configuration

tomrf1:
So far this branch defines the model and handles publishing to S3 (so it replaces the visuals tool, but is much more robust).
We now need to build a UI in react for viewing/updating the json (i.e. replace the spreadsheet).


paulbrown1982:
Could the tool download the exact same JSON file that's currently used by the site and at least display (read-only) the config in a better way than the sheets do, even if the sheets are the editing method at the moment?

tomrf1:
> Could the tool download the exact same JSON file that's currently used by the site and at least display (read-only) the config in a better way than the sheets do, even if the sheets are the editing method at the moment?

@paulbrown1982 the model is very different to what is served up by the spreadsheet (frontend currently has to do a lot of work to transform it into its own model). So I'm not sure it's worth the trouble. The spreadsheet turned out to be a poor format for configuring tests - just convenient at the start.

PR #26, opened 2019-08-05
Form components for Epic Tests

tjsilver:
- All test-level and variant-level fields created.
- All fields update test to S3.

### DIFF
--- a/public/src/components/epicTests/epicTestEditor.tsx
+++ b/public/src/components/epicTests/epicTestEditor.tsx
@@ -113,9 +113,8 @@ class EpicTestEditor extends React.Component<Props, any> {
               }
             }}
           >
-            <MenuItem value={UserCohort.OnlyExistingSupporters}>{UserCohort.OnlyExistingSupporters}</MenuItem>
-            <MenuItem value={UserCohort.OnlyNonSupporters}>{UserCohort.OnlyNonSupporters}</MenuItem>
-            <MenuItem value={UserCohort.Everyone}>{UserCohort.Everyone}</MenuItem>
+            
+            {Object.values(UserCohort).map(cohort => <MenuItem key={cohort} value={cohort}>{cohort}</MenuItem>)}
 
           </Select>
           </FormControl>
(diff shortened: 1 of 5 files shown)


public/src/components/epicTests/epicTestEditor.tsx
```diff
@@ -29,43 +33,93 @@ const styles = ({ palette, spacing, mixins }: Theme) => createStyles({
...
-    }
+  onVariantSelected = (variantName: string): void => {
+    this.setState({
+      selectedVariantName: variantName
+    })
   };
 
-  onUserCohortChange = (cohort: UserCohort) => {
+
+  updateTest = (fieldName: TestFieldNames, updatedData: string | (string | EpicVariant)[] | boolean) => {
```
tomrf1: I know I started this with `onListChange`, but looking at this again I think we should avoid this pattern.
The function signature gives the illusion of type safety, but actually the compiler cannot guarantee that the type of `updatedData` matches the `fieldName`. E.g. you could call `updateTest('isLiveBlog', 'should be a bool')` and it will not complain.

How about this instead? Not as clean, but it type checks:
```
  updateTest = (update: (test: EpicTest) => EpicTest) => {
    if (this.props.test) {
      this.props.onChange(update(this.props.test))
    }
  };

  onUserCohortChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
    let selectedCohort = event.target.value as UserCohort;
    this.updateTest(test => ({ ...test, "userCohort": selectedCohort}));
  };
```

tjsilver: Great - I think I've updated both the updateTest and updateVariant accordingly now.

public/src/components/epicTests/epicTestVariantEditor.tsx
```diff
@@ -0,0 +1,174 @@
...
+  highlightedText = "highlightedText",
+  footer = "footer",
+  showTicker = "showTicker",
+  backgroundImageUrl = "backgroundImageUrl",
+  ctaText = "ctaText",
+  supportBaseURL = "supportBaseURL"
+}
+class EpicTestVariantEditor extends React.Component<Props, any> {
+
+  // updateVariant = (fieldName: VariantFieldNames, updatedData: string | string[] | boolean) => {
```
tomrf1: can be removed

tjsilver: Done!

PR #33, opened 2019-09-05
Read-only, epic and lock functionality - client-side

tjsilver:
Functionality to lock and unlock epic tests for editing.

### DIFF
--- a/public/src/components/epicTests/epicTestVariantsList.tsx
+++ b/public/src/components/epicTests/epicTestVariantsList.tsx
@@ -18,7 +18,8 @@ const styles = ({ typography }: Theme) => createStyles({
 interface EpicTestVariantsListProps extends WithStyles<typeof styles> {
   variants: EpicVariant[],
   onVariantsListChange: (variantList: EpicVariant[]) => void,
-  testName: string
+  testName: string,
+  editMode: boolean
 }
 
 type EpicTestVariantsListState = {
@@ -111,6 +112,7 @@ class EpicTestVariantsList extends React.Component<EpicTestVariantsListProps, Ep
                 <EpicTestVariantEditor
                   variant={variant}
                   onVariantChange={this.onVariantChange}
+                  editMode={this.props.editMode}
                 />
               </ExpansionPanelDetails>
             </ExpansionPanel>
--- a/public/src/components/epicTests/epicTestVariantsList.tsx
+++ b/public/src/components/epicTests/epicTestVariantsList.tsx
@@ -38,11 +38,11 @@ class EpicTestVariantsList extends React.Component<EpicTestVariantsListProps, Ep
   onVariantChange = (updatedVariant: EpicVariant): void => {
     const updatedVariantList: EpicVariant[] = this.props.variants.map(variant => variant.name === updatedVariant.name ? updatedVariant : variant);
     this.props.onVariantsListChange(updatedVariantList);
-  }
+  };
 
   onVariantNameCreation = (name: string) => {
     this.createVariant(name);
-  }
+  };
 
   onExpansionPanelChange = (key: string) => (event: React.ChangeEvent<{}>) => {
     this.state.expandedVariantKey === key ? this.setState({
@@ -52,11 +52,11 @@ class EpicTestVariantsList extends React.Component<EpicTestVariantsListProps, Ep
     this.setState({
       expandedVariantKey: key
     })
-  }
+  };
 
   createVariantKey = (variantName: string): string => {
     return `${this.props.testName}${variantName}`
-  }
+  };
 
   createVariant = (newVariantName: string) => {
     const newVariant: EpicVariant = {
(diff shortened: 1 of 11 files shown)


public/src/components/epicTests/epicTestVariantsList.tsx
```diff
@@ -111,23 +116,34 @@ class EpicTestVariantsList extends React.Component<EpicTestVariantsListProps, Ep
...
             </ExpansionPanel>
           );
         })}
       </>
     );
-  }
+  };
 
-  render(): React.ReactNode {
+  renderVariants = (): React.ReactNode => (
```
tomrf1: this doesn't need to be a separate function

public/src/components/epicTests/epicTestsForm.tsx
```diff
@@ -138,26 +164,91 @@ class EpicTestsForm extends React.Component<EpicTestFormProps, EpicTestsFormStat
       })
   }
 
+  requestEpicTestsLock = () => {
+    requestLock(FrontendSettingsType.epicTests).then(
+      response => {
+        response.ok ? this.setState({ editMode: true }) : alert("Error - can't request lock!");
```
tomrf1: thinking about this again, it would be better to always call `this.fetchStateFromServer()` when these requests come back with `response.ok` and set `editMode` elsewhere.
I'll create a PR with a suggested change

public/src/components/epicTests/epicTestVariantsList.tsx
tjsilver: Ah ok will take a look

tjsilver: I've updated the file now.

public/src/components/epicTests/epicTestsForm.tsx
```diff
@@ -138,26 +165,87 @@ class EpicTestsForm extends React.Component<EpicTestFormProps, EpicTestsFormStat
...
+        onConfirm={() => this.cancel()}
+        icon={<RefreshIcon />}
+        />
+        </div>
+        )
+    }
+  }
+
+  makeFriendlyDate = (timestamp: string): string => {
+    const year = timestamp.slice(0, 4);
```
tomrf1:
```
const dt = new Date(timestamp);
return `${dt.getHours()}:${dt.getMinutes()} on ${dt.getDate()}/${dt.getMonth()}/${dt.getFullYear()}`
```

tjsilver: Of course! Will update.

tomrf1: you can make it a bit smarter with [moment](https://momentjs.com/), but maybe not worth adding a new dependency for

public/src/components/epicTests/epicTestsForm.tsx
```diff
@@ -138,26 +165,87 @@ class EpicTestsForm extends React.Component<EpicTestFormProps, EpicTestsFormStat
...
+    const month = timestamp.slice(5,7);
+    const day = timestamp.slice(8,10);
+    const time = timestamp.slice(11,19);
+    return `${time} on ${day}/${month}/${year}`;
+  }
+
+  makeFriendlyName = (email: string): string => {
```
tomrf1: should this include the second name?

tjsilver: I wasn't sure - was trying to keep it friendly as there is such a small number of users - what do you think?

tomrf1: I think there's a good chance of having 2 people with the same first name!

tjsilver: True - ok will add last name too.

public/src/components/epicTests/epicTestsForm.tsx
tjsilver: agreed.

public/src/components/epicTests/epicTestsForm.tsx
```diff
@@ -138,26 +165,84 @@ class EpicTestsForm extends React.Component<EpicTestFormProps, EpicTestsFormStat
...
+  }
+
+  makeFriendlyDate = (timestamp: string): string => {
+    const datetime = new Date(timestamp);
+    return `${datetime.getHours()}:${datetime.getMinutes()} on ${datetime.getDate()}/${datetime.getMonth() + 1}/${datetime.getFullYear()}`;
+  }
+
+  makeFriendlyName = (email: string): string => {
+    const firstName = email.split('.')[0];
+    const lastName = email.split('.')[1].split('@')[0];
```
tomrf1: an alternative to this is to use `match`, e.g.
`'first.last@example.com'.match(/^([a-z]*)\.([a-z]*)@.*/)`
Which returns `[ "first.last@example.com", "first", "last" ]`

Although regex can be quite unreadable! I don't have a strong preference, what do you think?

paulbrown1982: Don't forget we have x.y.freelancer and x.y.contractor too!

tjsilver: Oh yes!

tomrf1: good point! Perhaps it's more useful just to display the email address then?

tomrf1: otherwise this should work:
`/^([a-z]*)\.([a-z]*)(\.[a-z]*)?@.*/`

tjsilver: Ah yes - was just testing this which also seems to work /^([a-z]*)\.([a-z]*).*@.*/

tjsilver: Have updated with regex in latest commit.

public/src/components/epicTests/epicTestsForm.tsx
```diff
@@ -138,26 +165,83 @@ class EpicTestsForm extends React.Component<EpicTestFormProps, EpicTestsFormStat
...
+  }
+
+  makeFriendlyDate = (timestamp: string): string => {
+    const datetime = new Date(timestamp);
+    return `${datetime.getHours()}:${datetime.getMinutes()} on ${datetime.getDate()}/${datetime.getMonth() + 1}/${datetime.getFullYear()}`;
+  }
+
+  makeFriendlyName = (email: string): string | undefined => {
+    const nameArr: RegExpMatchArray | null = email.match(/^([a-z]*)\.([a-z]*).*@.*/);
+    return nameArr ? `${nameArr[1][0].toUpperCase()}${nameArr[1].slice(1,)} ${nameArr[2][0].toUpperCase()}${nameArr[2].slice(1,)}` : undefined;
```
tomrf1: how about instead of returning `undefined` if the match fails, just return `email` as is?

tjsilver: The UI would then display the email twice in that case - but I guess that won't really happen in practice.

tomrf1: oh right, I didn't realise it's already displaying the email!

tomrf1: 👍 looks good!

tjsilver: Only on the main page (with link), but displays the name alone in the popup warning message.

PR #34, opened 2019-09-09
Always fetch latest state from server

tomrf1:
whatever the operation, always fetch the latest data and set `editMode` based on that

TODO - test it actually works

### DIFF
--- a/app/controllers/LockableSettingsController.scala
+++ b/app/controllers/LockableSettingsController.scala
@@ -24,7 +24,7 @@ object LockStatus {
 
 object LockableSettingsController {
   // The model returned by this controller for GET requests
-  case class LockableSettingsResponse[T](value: T, version: String, status: LockStatus)
+  case class LockableSettingsResponse[T](value: T, version: String, status: LockStatus, userEmail: String)
 }
 
 class LockableSettingsController[T : Decoder : Encoder](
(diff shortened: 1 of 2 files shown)


tomrf1:
This also means that if you open in a new window, it knows you're in edit mode straight away

app/controllers/LockableSettingsController.scala
```diff
@@ -55,11 +55,11 @@ class LockableSettingsController[T : Decoder : Encoder](
     * Returns current version of the settings in s3 as json, with the lock status.
     * The s3 data is validated against the model.
     */
-  def get= authAction.async {
+  def get= authAction.async { request =>
     withLockStatus { case VersionedS3Data(lockStatus, _) =>
       S3Json.getFromJson[T](dataBucket, dataKey)(s3Client).map {
         case Right(VersionedS3Data(value, version)) =>
-          Ok(S3Json.noNulls(LockableSettingsResponse(value, version, lockStatus).asJson))
+          Ok(S3Json.noNulls(LockableSettingsResponse(value, version, lockStatus, request.user.email).asJson))
```
tjsilver: So now we can ensure that the user doesn't get the 'locked by user' message if they have the lock and refresh the page.

PR #35, opened 2019-09-10
Hide epic tests tool in menu

tomrf1:
We're still working on the epic tests tool and so do not want it to be in the menu.
But hardcoding a feature switch is quite inconvenient

Screenshot 2019-09-10 at 16 04 04

### DIFF
--- a/public/src/main.tsx
+++ b/public/src/main.tsx
@@ -17,8 +17,6 @@ import ListItemText from '@material-ui/core/ListItemText';
 import Typography from '@material-ui/core/Typography';
 import ListItem from '@material-ui/core/ListItem';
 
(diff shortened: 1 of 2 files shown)


public/src/main.tsx
```diff
@@ -47,20 +45,13 @@ const styles = ({ palette, spacing, mixins }: Theme) => createStyles({
   }
 });
 
-const getLinkPathsAndNames = (): string[][] => {
-  const baseList = [
+const getLinkPathsAndNames = (): string[][] =>
+  [
     ['/', 'Home'],
```
tjsilver: Do we need home on the menu? The page is empty...

tomrf1: good point!

PR #41, opened 2019-09-12
Set surrogate-control on epic-tests.json

tomrf1:
Make fastly cache the file.
see https://docs.fastly.com/en/guides/how-caching-and-cdns-work#surrogate-headers
TODO - purge the fastly cache after updates


tomrf1:
@jlieb10 The intention is to have a short ttl in users' browsers, but also have fastly cache the file for a long time to avoid hitting s3 frequently.
Fastly decides how long to cache the resource based on the `surrogate-control` header. But fastly only passes through the `cache-control` header to the client.

This means that when we update the file in s3, we need to manually purge the fastly cache (PR coming soon)
Happy to talk through this IRL!

PR #48, opened 2019-09-17
Field validation

tomrf1:
### With an invalid field input (paragraphs must not be empty):
- Field underlined red
- Variant has red outline
- The test in the tests list has red outline
- Publish button disabled
Screenshot 2019-09-17 at 15 46 46

### All fields are valid:
- Test has green outline, to indicate that it has been modified but is valid
Screenshot 2019-09-17 at 15 46 55

Introduces `validation.tsx` with common logic for tracking the validation status of each child component, and reporting a single true/false to the parent.
E.g. VariantsList component reports `false` if any of its VariantEditor components are invalid.


public/src/components/epicTests/epicTestVariantEditor.tsx
```diff
@@ -117,20 +119,26 @@ class EpicTestVariantEditor extends React.Component<Props> {
           <Typography>Fill out each field before publishing your test</Typography>
 
           <EditableTextField
-            required
```
tjsilver: Should we move this field now that it's not required?

tomrf1: Since there's only one mandatory field (paragraphs) and it has `Must not be empty` underneath it, I think we can remove the 'Required' + 'Optional' headings. What do you think?

tjsilver: Yes - we can explain to UX team why that makes more sense now.

tomrf1: 👍 I've updated the screenshots

tjsilver: Great!!

public/src/components/helpers/validation.tsx
```diff
@@ -0,0 +1,56 @@
...
+export type ValidationStatus = {
+  [fieldName: string]: boolean
+}
+
+export interface ValidationComponentState {
+  validationStatus: ValidationStatus
+}
+
+// Call this when a single field in the component updates
+export const onFieldValidationChange = <
```
tomrf1: this approach is up for debate, perhaps we can revisit later

PR #50, opened 2019-09-18
Delete epic test

tjsilver:
Addition of a 'delete test' button. Deleted tests will appear in the test list on the left and be disabled for editing. They will be fully deleted upon publish. Deletion can only be reversed by pressing 'cancel' button at the all-test level.

image


jlieb10:
Well done Thalia!

PR #51, opened 2019-09-18
Truncate long Epic test names in the list

tomrf1:
Screenshot 2019-09-18 at 16 47 35


tomrf1:
Just realised this breaks when you're in edit mode
Screenshot 2019-09-18 at 16 58 59

PR #55, opened 2019-09-19
Report isValid == true when a variant is deleted

tomrf1:
otherwise the test is still marked as invalid, even though the variant is gone


public/src/components/epicTests/epicTestVariantEditor.tsx
```diff
@@ -53,7 +53,7 @@ interface Props extends WithStyles<typeof styles> {
   variant?: EpicVariant,
   onVariantChange: (updatedVariant: EpicVariant) => void,
   editMode: boolean,
-  onDelete: (variantName: string) => void
+  onDelete: () => void
```
tomrf1: The parent knows the variant's name when it creates this component, so no need to pass the name up

tjsilver: Ah!

PR #64, opened 2019-09-20
Case-insensitive name validation

tjsilver:
Names of tests and variants will be unique regardless of case.

### DIFF
--- a/public/src/components/epicTests/newNameCreator.tsx
+++ b/public/src/components/epicTests/newNameCreator.tsx
@@ -51,8 +51,7 @@ class NewNameCreator extends React.Component<NewNameCreatorProps, NewNameCreator
 
   isDuplicateName = (newName: string) => {
     const newLowerCase: string = newName.toLowerCase();
-    const existingLowerCase: string[] = this.props.existingNames.map(name => name.toLowerCase());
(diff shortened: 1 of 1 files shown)


public/src/components/epicTests/newNameCreator.tsx
```diff
@@ -50,7 +50,9 @@ class NewNameCreator extends React.Component<NewNameCreatorProps, NewNameCreator
   }
 
   isDuplicateName = (newName: string) => {
-    const isDuplicate = this.props.existingNames.includes(newName);
+    const newLowerCase: string = newName.toLowerCase();
+    const existingLowerCase: string[] = this.props.existingNames.map(name => name.toLowerCase());
+    const isDuplicate = existingLowerCase.includes(newLowerCase);
```
tomrf1: This is fine, but it can be slightly more efficient (only requires up to 1 pass through the array):
```
const newLowerCase: string = newName.toLowerCase();
const isDuplicate = this.props.existingNames.some(name => name.toLowerCase() === newLowerCase);
```

tjsilver: 👍

PR #77, opened 2019-09-27
Template validation in the Epic Tests tool

tomrf1:
We have some special text templates which get replaced by dotcom, e.g. `%%COUNTRY_NAME%%`.
There is validation in dotcom as well, but it's helpful to validate here in case the user makes a typo.

Also, I've changed how the validation in the `EditableTextField` component works.
Instead of taking
`isValid: (value: string) => boolean`
it now takes
`getError: (value: string) => string | null`

This is so that it can display an error message under the field.

No errors:
Screenshot 2019-09-27 at 16 22 58

Errors:
Screenshot 2019-09-27 at 16 23 29

Also, ideally the UI would make it a bit easier to use these templates, e.g. something you can drag and drop?


public/src/components/epicTests/epicTestVariantEditor.tsx
```diff
@@ -10,6 +10,21 @@ import DeleteSweepIcon from '@material-ui/icons/DeleteSweep';
 import {onFieldValidationChange, ValidationStatus} from '../helpers/validation';
 
 
+const validTemplates = ["%%CURRENCY_SYMBOL%%", "%%COUNTRY_NAME%%"];
+
+const getInvalidTemplateError = (text: string): string | null => {
```
tomrf1: It's really time we had client-side testing in this project! This function could do with some tests.
I'll do this next week

tomrf1: see latest commit!

build-tc
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
tomrf1: this script is now run by teamcity, which previously ran:
```
npm install
npm run build
```

PR #78, opened 2019-09-27
Update npm packages

tjsilver:
This updates React and Material UI (among other things) to the latest versions.

### DIFF
--- a/public/src/components/epicTests/epicTestEditor.tsx
+++ b/public/src/components/epicTests/epicTestEditor.tsx
@@ -67,13 +67,13 @@ const styles = ({ spacing, typography}: Theme) => createStyles({
     marginBottom: "10px"
   },
   visibilityIcons: {
-    marginTop: spacing(1.5)
+    marginTop: spacing(1)
   },
   switchWithIcon: {
     display: "flex"
   },
   visibilityHelperText: {
-    marginTop: spacing(1.8),
+    marginTop: spacing(1),
     marginLeft: spacing(1)
   },
   deleteButton: {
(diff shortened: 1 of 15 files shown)


public/src/components/epicTests/epicTestEditor.tsx
```diff
@@ -287,7 +287,7 @@ class EpicTestEditor extends React.Component<EpicTestEditorProps, EpicTestEditor
                 className={classes.select}
                 multiple
                 value={test.locations}
-                onChange={this.onLocationsChange}
+                onChange={() => this.onLocationsChange}
```
tomrf1: this change means it will not actually call `this.onLocationsChange`

public/src/components/epicTests/newNameCreator.tsx
```diff
@@ -80,37 +80,37 @@ class NewNameCreator extends React.Component<NewNameCreatorProps, NewNameCreator
...
-              vertical: 'bottom',
-              horizontal: 'center',
-            }}
-            transformOrigin={{
-              vertical: 'top',
-              horizontal: 'center',
-            }}
+        </Button>
+        <Popover
+      open={this.state.newNamePopoverOpen}
```
tomrf1: some strange indentation here

public/src/components/epicTests/epicTestEditor.tsx
tjsilver: I think I've fixed it now with latest commit.

public/src/components/epicTests/newNameCreator.tsx
tjsilver: Oops! also fixed now.

tomrf1:
👍

PR #80, opened 2019-10-01
Add classNames field to Epic tests model (backend)

tomrf1:
This is for adding custom css classes to the epic.
Corresponding frontend change here https://github.com/guardian/frontend/pull/21873

TODO - add the field to the client

### DIFF
--- a/app/models/EpicTests.scala
+++ b/app/models/EpicTests.scala
@@ -42,7 +42,8 @@ case class EpicVariant(
   footer: Option[String] = None,
   showTicker: Boolean = false,
   backgroundImageUrl: Option[String] = None,
-  cta: Option[Cta]
+  cta: Option[Cta],
+  classNames: List[String] = Nil
 )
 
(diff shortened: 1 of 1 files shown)


app/models/EpicTests.scala
```diff
@@ -42,7 +42,8 @@ case class EpicVariant(
   footer: Option[String] = None,
   showTicker: Boolean = false,
   backgroundImageUrl: Option[String] = None,
-  cta: Option[Cta]
+  cta: Option[Cta],
+  classNames: List[String] = Nil
```
tomrf1: this change is backwards-compatible with our existing data because, although the field is not optional, it defaults to `Nil` if not present

tomrf1:
Closing this PR in favour of @jlieb10 's excellent suggestion, which is to construct class names in frontend based on the test and variant names. It means we don't have to maintain this from the tool at all

PR #82, opened 2019-10-04
Add validation to new name creator to restrict characters

tjsilver:
Validation on the new name creator, used to new epic and variants. Spaces and special characters other than '-' and '_' are not allowed. This is to make it safer when the variant or test names are used in other platforms for analysis or as classnames.

### DIFF
--- a/public/src/components/epicTests/newNameCreator.tsx
+++ b/public/src/components/epicTests/newNameCreator.tsx
@@ -38,40 +38,52 @@ class NewNameCreator extends React.Component<NewNameCreatorProps, NewNameCreator
     errorMessage: ""
   }
 
-  onNewNameButtonClick = (event: React.MouseEvent<HTMLButtonElement>) =>  {
+  onNewNameButtonClick = (event: React.MouseEvent<HTMLButtonElement>): void =>  {
     this.setState({
       newNamePopoverOpen: true,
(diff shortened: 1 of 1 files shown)


public/src/components/epicTests/newNameCreator.tsx
```diff
@@ -38,40 +38,52 @@ class NewNameCreator extends React.Component<NewNameCreatorProps, NewNameCreator
...
+  isDuplicateName = (newName: string): boolean => {
     const newLowerCase: string = newName.toLowerCase();
     const isDuplicate = this.props.existingNames.some(name => name.toLowerCase() === newLowerCase);
     return isDuplicate;
-  }
+  };
 
-  handleName = (newTestName: string) => {
+  containsInvalidChars = (newName: string): boolean => {
+    return (/[^\w-]/.test(newName));
```
tomrf1: we should allow `_` as well

tomrf1: sorry, just realised `\w` covers this! Perfect!

tjsilver: Yep, that's including in `\w`: Matches any alphanumeric character including the underscore. Equivalent to [A-Za-z0-9_].

PR #84, opened 2019-10-09
Set default highlightedText

tomrf1:
The UI was displaying the default value as a placeholder, but actually setting the value to `""`.
This is very misleading as it makes the user think it has been set.

I've manually fixed all the PROD epic tests where this was the case.
This change will ensure it doesn't happen again.
I've also made the variantEditor interpret empty strings in optional fields as unset (`undefined`).


tomrf1:
After some googling [we discovered](https://medium.com/@apdharshi/understanding-the-usage-of-optional-and-nullable-in-typescript-826c1754df3) that in typescript:
- `foo?: string` must be either `string` or `undefined`
- `foo: string?` may be one of `string`, `undefined`, `null`

So in this case, it must be `undefined`

tjsilver:
Great work!

PR #86, opened 2019-10-11
Add archive endpoint to epic tests backend

tomrf1:
Allows the client to send a single epic test for archiving. It is saved to a new file at:
`support-admin-console/<STAGE>/archived-epic-tests/<TEST NAME>.json`

When we build a UI for managing these archived tests we'll need new endpoints for listing and fetching.

### DIFF
--- a/app/controllers/EpicTestsController.scala
+++ b/app/controllers/EpicTestsController.scala
@@ -2,10 +2,11 @@ package controllers
 
 import com.gu.googleauth.AuthAction
 import play.api.mvc.{AnyContent, ControllerComponents}
-import models.EpicTests
+import models.{EpicTest, EpicTests}
 import play.api.libs.ws.WSClient
-import services.FastlyPurger
+import services.{FastlyPurger, S3Json, VersionedS3Data}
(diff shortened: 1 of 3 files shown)


app/controllers/EpicTestsController.scala
```diff
@@ -34,4 +35,25 @@ class EpicTestsController(authAction: AuthAction[AnyContent], components: Contro
     ),
     fastlyPurger = EpicTestsController.fastlyPurger(stage, ws)
   ) {
+
+  /**
+    * Saves a single test to a new file. The file name will be the test's name.
+    * Overwrites any existing archived test of the same name.
+    * Removing the test from the current active tests list is done separately.
+    */
+  def archive = authAction.async(circe.json[VersionedS3Data[EpicTest]]) { request =>
```
tomrf1: @tjsilver I ended up adding this handler to the `EpicTestsController` because it's very specific to epic tests, and so doesn't need to be in a more generic controller

tjsilver: good idea 👍

PR #87, opened 2019-10-11
Only fetch s3 object metadata

tomrf1:
we're still getting `Timeout waiting for connection from pool` errors when the app has been running for a while.
The `put` function fetches the current object before updating it, even though it doesn't read the content stream. The s3 object should then be closed - this is likely the cause.
However, we can just fetch the object metadata instead, since that's all we need.

@JoemG

### DIFF
--- a/app/services/S3.scala
+++ b/app/services/S3.scala
@@ -63,7 +63,9 @@ object S3 extends S3Client with StrictLogging {
 
   def put(objectSettings: S3ObjectSettings, data: RawVersionedS3Data)(implicit ec: ExecutionContext): Future[Either[String,RawVersionedS3Data]] = Future {
     try {
-      val currentVersion = s3Client.getObject(objectSettings.bucket, objectSettings.key).getObjectMetadata.getVersionId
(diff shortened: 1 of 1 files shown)


app/services/S3.scala
```diff
@@ -63,7 +66,7 @@ object S3 extends S3Client with StrictLogging {
 
   def put(objectSettings: S3ObjectSettings, data: RawVersionedS3Data)(implicit ec: ExecutionContext): Future[Either[String,RawVersionedS3Data]] = Future {
     try {
-      val currentVersion = s3Client.getObject(objectSettings.bucket, objectSettings.key).getObjectMetadata.getVersionId
+      val currentVersion = s3Client.getObjectMetadata(objectSettings.bucket, objectSettings.key).getVersionId
```
tomrf1: this is the actual fix (hopefully)

tomrf1:
It was probably breaking after the 50th write to s3, assuming the connections were never being released after each update:
https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/ClientConfiguration.java#L69

PR #94, opened 2019-11-13
Set articlesViewedSettings in the epic tool

tomrf1:
[Corresponding frontend PR](https://github.com/guardian/frontend/pull/21991).

This change means:
1. Allows users to add the `%%ARTICLE_COUNT%%` template in the copy, which frontend replaces with a count of articles viewed over a period of time
2. If this template is present then the test configuration must also include an `articlesViewedSettings` field, which has the period in weeks and optional max/min counts. The max/min counts are for limiting who sees the test based on the count.

For now I have hardcoded the `articlesViewedSettings` to min 5 articles per 4 weeks. We'll probably need a UI change to make this configurable later, but just having 5/month is a big leap forward by enabling non-devs to include article counts in epic tests.

E.g. the POST to server contains `articlesViewedSettings` when paragraphs contains the template:
Screenshot 2019-11-13 at 08 20 25

### DIFF
--- a/app/models/EpicTests.scala
+++ b/app/models/EpicTests.scala
@@ -51,6 +51,12 @@ case class MaxViews(
   minDaysBetweenViews: Int
 )
 
+case class ArticlesViewedSettings(
+  minViews: Option[Int],
+  maxViews: Option[Int],
+  periodInWeeks: Int
+)
+
 case class EpicTest(
   name: String,
   isOn: Boolean,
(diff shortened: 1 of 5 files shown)


app/models/EpicTests.scala
```diff
@@ -51,6 +51,12 @@ case class MaxViews(
   minDaysBetweenViews: Int
 )
 
+case class ArticlesViewedSettings(
+  minViews: Option[Int],
+  maxViews: Option[Int],
```
jlieb10: Curious to hear about the maxViews cases

PR #98, opened 2019-11-26
Use Zio

tomrf1:
This is an experiment to see if [zio](https://zio.dev/) can improve the app's code (I'm not concerned about performance improvements here).

This branch does not make direct use of the lower level concurrency features that zio provides.
The primary benefit (IMO) is that it provides an alternative abstraction for handling async/blocking tasks that can fail.
So whereas previously it used `Future[Either[E,T]]` (and cats `EitherT`), it now uses the `ZIO` type. Only in the Play controllers do we have to produce a `Future` using zio's `unsafeRunToFuture`.

#### Benefits:
- the zio api arguably allows for cleaner/nicer code compared to handling `Future[Either[...]]` everywhere.
- much safer than using Futures.
- synchronous blocking (e.g. s3 requests) easily handled by zio which [uses a separate thread pool designed specifically for blocking effects](https://zio.dev/docs/overview/overview_creating_effects#blocking-synchronous-side-effects).
- nice management of resources that need to be released.

#### Drawbacks:
- another dependency.
- yet another abstraction/api that developers have to learn. It's quite a big barrier to entry for people who are new to scala (not that `EitherT` was very friendly).
- if no other codebases at the Guardian end up using zio then this one is going to be annoying to maintain...but we have to start somewhere!
- it has a dsl with silly operators but we do not have to use them.


app/services/S3.scala
```diff
@@ -9,23 +9,25 @@ import com.typesafe.scalalogging.StrictLogging
...
-  def listKeys(objectSettings: S3ObjectSettings)(implicit ec: ExecutionContext): Future[Either[String, List[String]]]
+  def get: S3Action[RawVersionedS3Data]
+  def update(data: RawVersionedS3Data): S3Action[Unit]
+  def createOrUpdate(data: String): S3Action[Unit]
+  def listKeys: S3Action[List[String]]
 }
 
 object S3Client {
   type RawVersionedS3Data = VersionedS3Data[String]
+  type S3Action[T] = S3ObjectSettings => ZIO[Blocking, S3ClientError,T]
```
tomrf1: Uses the `Blocking` environment for the s3 client

app/services/S3.scala
```diff
@@ -44,116 +60,111 @@ object S3 extends S3Client with StrictLogging {
...
-
-      val stream = s3Object.getObjectContent
-      val raw: String = try {
-        scala.io.Source.fromInputStream(stream).mkString
-      } finally {
-        stream.close()
-        s3Object.close()
+  def get: S3Action[RawVersionedS3Data] = { objectSettings =>
+    val s3ObjectAndStream = for {
+      s3Object <- ZManaged.fromAutoCloseable(effectBlocking(s3Client.getObject(objectSettings.bucket, objectSettings.key)))
```
tomrf1: `ZManaged.fromAutoCloseable` ensures that `.close()` is called on the managed resource

app/services/S3.scala
```diff
@@ -44,116 +60,111 @@ object S3 extends S3Client with StrictLogging {
...
-      if (currentVersion == data.version) {
-        put(objectSettings, data.value).map(_ => data)
-      } else {
-        logger.warn(s"Cannot update S3 object $objectSettings because provided version (${data.version}) does not match latest version ($currentVersion)")
-        Left(s"Can't save your settings because someone else has updated them since they were last fetched")
+    Task(new ByteArrayInputStream(bytes))
+      .bracket(stream => UIO(stream.close()))
+      .apply { stream =>
+
+        UIO.effectTotal {
```
tomrf1: this block will not fail, but has mutable state while it's setting up the request object, and so it's good to handle it as a synchronous effect

app/services/S3.scala
```diff
@@ -44,116 +60,111 @@ object S3 extends S3Client with StrictLogging {
...
-      result.flatMap { raw =>
-        decode[T](raw.value) match {
-          case Right(decoded) => Right(raw.copy(value = decoded))
-          case Left(error) =>
-            logger.error(s"Error decoding json from S3 ($objectSettings): ${error.getMessage}", error)
-            Left(s"Error decoding json from S3 ($objectSettings): ${error.getMessage}")
+  def getFromJson[T : Decoder](s3: S3Client): S3ObjectSettings => ZIO[Blocking,Throwable,VersionedS3Data[T]] = { objectSettings =>
+    s3.get(objectSettings).flatMap { raw =>
+      IO(decode[T](raw.value))
+        .absolve
```
tomrf1: absolve just turns an Either into a ZIO (circe's `decode` returns Either)

tomrf1: and actually, `fromEither` is clearer

app/controllers/EpicTestsController.scala
```diff
@@ -76,14 +89,17 @@ class EpicTestsController(authAction: AuthAction[AnyContent], components: Contro
...
-        logger.error(s"Failed to fetch list of archived test names: $error")
-        InternalServerError(error)
+    run {
+      s3Client
+        .listKeys(objectSettings)
+        .map { keys =>
+          val testNames: List[String] = keys.flatMap(extractTestName)
+          Ok(S3Json.noNulls(testNames.asJson))
+        }
+        .mapError { error =>
```
regiskuckaertz: Use [`tapError`](https://javadoc.io/static/dev.zio/zio_2.12/1.0.0-RC17/zio/ZIO.html#tapError[R1<:R,E1>:E](f:E=>zio.ZIO[R1,E1,Any])(implicitev:zio.CanFail[E]):zio.ZIO[R1,E1,A]) instead.

regiskuckaertz: (logging is an effect which can throw)

app/controllers/SettingsController.scala
```diff
@@ -22,14 +24,27 @@ abstract class SettingsController[T : Decoder : Encoder](authAction: AuthAction[
   )
   private val s3Client = services.S3
 
+  protected val runtime = new DefaultRuntime {}
```
regiskuckaertz: If possible, share a single runtime across controllers.

app/services/S3.scala
```diff
@@ -44,116 +60,111 @@ object S3 extends S3Client with StrictLogging {
...
+  def get: S3Action[RawVersionedS3Data] = { objectSettings =>
+    val s3ObjectAndStream = for {
+      s3Object <- ZManaged.fromAutoCloseable(effectBlocking(s3Client.getObject(objectSettings.bucket, objectSettings.key)))
+      stream <- ZManaged.fromAutoCloseable(effectBlocking(s3Object.getObjectContent))
+    } yield (s3Object, stream)
+
+    s3ObjectAndStream.use { case (s3Object, stream) =>
+      Task {
+        VersionedS3Data(
+          value = scala.io.Source.fromInputStream(stream).mkString,
```
regiskuckaertz: This looks like the only effect in that hunk of code.

In the future you might want to use `ZStreamChunk.fromInputStream` to avoid loading everything into a string on the heap 😎

app/controllers/EpicTestsController.scala
JonNorman: What does `tapError` do if the logging throws? Provide the original error, or the one produced from the logging?

regiskuckaertz: Errors are never lost in ZIO. They are accumulated in sequence or in parallel, depending. An action produces an `Exit[E, A]` which can either be a success, or a failure. In the latter case, the failure contains a [`Cause[E]`](https://javadoc.io/doc/dev.zio/zio_2.12/latest/zio/Cause.html) where you have access to all the errors that happened.

JonNorman: Thanks @regiskuckaertz, I really want to have play with zio, it looks like a breath of fresh air.

tomrf1: In this case `tapError` works fine. In cases where I want to narrow the type of the error it fails because of the type bound on the error type in `tapError[R1 <: R, E1 >: E](f: E => ZIO[R1, E1, _]): ZIO[R1, E1, A]`
E.g.
```
[error]  found   : zio.ZIO[zio.blocking.Blocking,Throwable,services.VersionedS3Data[String]]
[error]  required: zio.ZIO[zio.blocking.Blocking,services.S3Client.S3ClientError,services.S3Client.RawVersionedS3Data]
```
I think I just have to stick with `mapError` in these cases?

regiskuckaertz: If your code looks sth like this:
```scala
.tapError(x => Task(logger.error(...)))
```
Then it makes sense, as `Throwable` is the least upper bound between `S3ClientError` and `Throwable`. Since logging is not essential, I suggest you just forget the error, like this:
```scala
.tapError(x => Task(logger.error(...)).ignore)
```

app/wiring/AppComponents.scala
```diff
@@ -47,14 +48,16 @@ class AppComponents(context: Context, stage: String) extends BuiltInComponentsFr
...
     new Application(authAction, controllerComponents),
     new Login(authConfig, wsClient, requiredGoogleGroups, googleGroupChecker, controllerComponents),
-    new SwitchesController(authAction, controllerComponents, stage),
-    new ContributionTypesController(authAction, controllerComponents, stage),
-    new AmountsController(authAction, controllerComponents, stage),
-    new EpicTestsController(authAction, controllerComponents, wsClient, stage),
+    new SwitchesController(authAction, controllerComponents, stage, runtime),
+    new ContributionTypesController(authAction, controllerComponents, stage, runtime),
+    new AmountsController(authAction, controllerComponents, stage, runtime),
+    new EpicTestsController(authAction, controllerComponents, wsClient, stage, runtime),
```
kelvin-chappell: Perhaps could avoid feeding this new dependency through by changing the implementation of the controller action, something like [this](https://github.com/loicdescotte/play-zio/blob/master/app/controllers/UserController.scala#L19) that uses [this implementation](https://github.com/loicdescotte/play-zio/blob/master/app/libs/http.scala#L19).

tomrf1: Nice! I'll try this in a follow up PR

PR #99, opened 2019-11-28
Articles viewed settings

jlieb10:
This change allows for test editors to configure custom settings for who sees an article based on how many or how few articles they've looked at in the past X weeks

### DIFF
--- a/public/src/components/epicTests/articlesViewedEditor.tsx
+++ b/public/src/components/epicTests/articlesViewedEditor.tsx
@@ -72,7 +72,7 @@ class ArticlesViewedEditor extends React.Component<Props, State> {
           this.props.onChange(
             {
               ...articlesViewedSettings || defaultArticlesViewedSettings,
-              [fieldName]: value !== null ? Number(value) : null;
+              [fieldName]: value !== null ? Number(value) : null,
             }
           )
         }
--- a/public/src/components/epicTests/articlesViewedEditor.tsx
+++ b/public/src/components/epicTests/articlesViewedEditor.tsx
@@ -68,11 +68,11 @@ class ArticlesViewedEditor extends React.Component<Props, State> {
     return (<EditableTextField
       text={ setting ? setting.toString() : ''}
       onSubmit={ (value: string) => {
-        if (isNumber(value)) {
+        if (value === null || isNumber(value)) {
           this.props.onChange(
             {
               ...articlesViewedSettings || defaultArticlesViewedSettings,
-              [fieldName]: Number(value)
+              [fieldName]: value !== null ? Number(value) : null;
             }
           )
         }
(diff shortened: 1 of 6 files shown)


public/src/components/epicTests/articlesViewedEditor.tsx
```diff
@@ -0,0 +1,134 @@
...
+    />)
+  };
+
+  onArticlesViewedEditorSwitchChange = (enabled: boolean) => {
+    if (enabled) {
+      this.props.onChange(defaultArticlesViewedSettings);
+    } else {
+      this.props.onChange(undefined);
+    }
+  }
```
tjsilver: Do we need the else here? Could it be `enabled && this.props.onChange(defaultArticlesViewedSettings)`?

public/src/components/epicTests/articlesViewedEditor.tsx
```diff
@@ -0,0 +1,134 @@
...
+interface State {
+  validationStatus: ValidationStatus
+}
+
+class ArticlesViewedEditor extends React.Component<Props, State> {
+  state: State = {
+    validationStatus: {}
+  };
+
+  buildField = (fieldName: ArticleViewsFieldName, label: string, articlesViewedSettings: ArticlesViewedSettings) => {
```
tjsilver: Do we need a way to set the `required` field as some fields are required?

public/src/components/epicTests/articlesViewedEditor.tsx
jlieb10: we need to set them specifically as undefined in the case the switch is disabled - once they are undefined, they will be hidden

public/src/components/epicTests/articlesViewedEditor.tsx
jlieb10: Added! Thanks ✅

public/src/components/epicTests/articlesViewedEditor.tsx
```diff
@@ -0,0 +1,135 @@
...
+    const setting: number | null = articlesViewedSettings[fieldName];
+
+    return (<EditableTextField
+      text={ setting ? setting.toString() : ''}
+      onSubmit={ (value: string) => {
+        if (isNumber(value)) {
+          this.props.onChange(
+            {
+              ...articlesViewedSettings || defaultArticlesViewedSettings,
+              [fieldName]: Number(value)
```
tomrf1: @jlieb10 this line is why it's setting the value to 0 when the user deletes the value:
If the field is empty then `value` will be null.
And `Number(null) === 0`

So this should work: `[fieldName]: value !== null ? Number(value) : null`

And to make things clearer, the `if` condition should explicitly allow nulls:
`if (value === null || isNumber(value))`

jlieb10: Amazing, thank you

PR #101, opened 2019-12-13
Epic test 'copy' button

tomrf1:
Adds a new 'copy test' button to the bottom of the epic test editor component.
This behaves the same as the 'new test' button, except it copies everything but the name.
The name input field is pre-populated with the name of the existing test, as it's likely to be similar.
Screenshot 2019-12-13 at 04 01 19

This is useful because we are often running very similar tests, e.g. with and without article view count

### DIFF
--- a/public/src/components/epicTests/epicTestEditor.tsx
+++ b/public/src/components/epicTests/epicTestEditor.tsx
@@ -88,7 +88,6 @@ const styles = ({ spacing, typography}: Theme) => createStyles({
   button: {
     marginTop: spacing(2),
     marginLeft: spacing(2),
-    float: 'right'
   },
   isDeleted: {
     color: '#ab0613'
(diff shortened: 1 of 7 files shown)


ionamckendrick:
This is incredible 👏

public/src/components/epicTests/epicTestEditor.tsx
```diff
@@ -86,6 +87,7 @@ const styles = ({ spacing, typography}: Theme) => createStyles({
   },
   button: {
     marginTop: spacing(2),
+    marginLeft: spacing(2),
     float: 'right'
```
tjsilver: Just wondering if float is redundant since the buttons are in a flexbox?

tomrf1: good point, you're right!

PR #104, opened 2019-12-24
Added 2 new switches: stripeApplePay and stripePaymentRequestButton

paulbrown1982:
Adding 2 new switches: stripeApplePay and stripePaymentRequestButton to both Single and Recurring switchboards. The first means we can roll out Apple Pay to 100 of eligible audience, as the test showed a 300% increase! Also added a second means we can turn off the ugly purple button as it showed little boost in conversion.

I also made a change to alphabetically sort the switch ordering, and renamed the case class PaymentMethodsSwitch to PaymentMethodSwitches as I feel it's a better name - it also fixed a bizarre compile issue.

Finally, I fixed some code analysis issues that IntelliJ complained about.

Screen Shot 2019-12-24 at 10 26 13

### DIFF
--- a/app/models/SupportFrontendSwitches.scala
+++ b/app/models/SupportFrontendSwitches.scala
@@ -25,8 +25,10 @@ case object SwitchState {
 
 case class ExperimentSwitch(name: String, description: String, state: SwitchState)
 
-case class PaymentMethodsSwitch(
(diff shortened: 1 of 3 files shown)


paulbrown1982:
Tested in CODE.

PR #108, opened 2020-01-30
Do not set default 'articles viewed' settings

tomrf1:
Edit: after discussion, we decided just to set the default to 52 weeks. The stuff below is now incorrect...

because it's easy to forget to adjust them to match what's in the copy.
The 'default' is now all nulls, and so the form now immediately fails validation until you set it yourself.

So, as soon as you add the `%%ARTICLE_COUNT%%` template to the copy it adds this:
Screen Shot 2020-01-30 at 11 52 33


ionamckendrick:
Great! Though please could we make the minimum articles forced too? And auto-populate it with 5? I can't see us doing that any differently any time soon. Not everyone using the tool will know it's 5 x so we should have that helpfully filled I think...

ionamckendrick:
Whilst I'm thinking about it, do we even need a "max articles" field? I think it's potentially confusing to have that there when I can't think of a use case for it...

PR #109, opened 2020-03-16
Improve selection patterns

tjsilver:
Most of the users are confused about the action they're going to perform when using the toggles and the countries dropdown. 

## Changes: 

* Remove 'High priority' toggle
* Change 'Test is live' toggle style
* Change 'Liveblog epic' switch to 'Epic type' radio buttons
* Change 'Locations' selector to nested 'Target regions' checkboxes

### Before - Toggles
image

### After - Toggles
image

### Before - Locations
image

image

### All selected
image

### Some selected
image

### None selected
image

### DIFF
--- a/app/models/EpicTests.scala
+++ b/app/models/EpicTests.scala
@@ -72,7 +72,7 @@ case class EpicTest(
   isLiveBlog: Boolean = false,
   hasCountryName: Boolean = false,
   variants: List[EpicVariant],
-  highPriority: Boolean = false,
+  highPriority: Boolean = false, // has been removed from form, but might be used in future
   useLocalViewLog: Boolean = false,
   articlesViewedSettings: Option[ArticlesViewedSettings] = None
 )
--- a/public/src/components/epicTests/epicTestEditor.tsx
+++ b/public/src/components/epicTests/epicTestEditor.tsx
@@ -1,4 +1,4 @@
-import React, { ReactNode, ChangeEvent } from 'react';
+import React, { ReactNode } from 'react';
 import {EpicTest, EpicVariant, UserCohort, MaxEpicViews, ArticlesViewedSettings} from "./epicTestsForm";
 import {
   Checkbox,
@@ -22,14 +22,12 @@ import EditableTextField from "../helpers/editableTextField"
 import { Region } from '../../utils/models';
 import EpicTestVariantsList from './epicTestVariantsList';
 import MaxEpicViewsEditor from './maxEpicViewsEditor';
-import { renderVisibilityIcons, renderVisibilityHelpText } from './utilities';
 import {onFieldValidationChange, ValidationStatus} from '../helpers/validation';
 import ButtonWithConfirmationPopup from '../helpers/buttonWithConfirmationPopup';
 import DeleteSweepIcon from '@material-ui/icons/DeleteSweep';
 import ArchiveIcon from '@material-ui/icons/Archive';
 import {articleCountTemplate, countryNameTemplate} from "./epicTestVariantEditor";
 import ArticlesViewedEditor, {defaultArticlesViewedSettings} from "./articlesViewedEditor";
-import articlesViewedEditor from "./articlesViewedEditor";
 import NewNameCreator from "./newNameCreator";
 
 const styles = ({ spacing, typography}: Theme) => createStyles({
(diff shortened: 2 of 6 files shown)


app/models/EpicTests.scala
```diff
@@ -72,7 +72,7 @@ case class EpicTest(
   isLiveBlog: Boolean = false,
   hasCountryName: Boolean = false,
   variants: List[EpicVariant],
-  highPriority: Boolean = false,
+  highPriority: Boolean = false, // has been removed from form, but might be used in future
```
tomrf1: 👍

public/src/components/epicTests/epicTestEditor.tsx
```diff
@@ -187,9 +193,14 @@ class EpicTestEditor extends React.Component<EpicTestEditorProps, EpicTestEditor
...
-  onLocationsChange = (event: React.ChangeEvent<{ name?: string | undefined; value: unknown; }>, child: ReactNode) => {
-    const selectedLocations = event.target.value as Region[];
-    this.updateTest(test => ({...test, "locations": selectedLocations}));
+  onEpicTypeChange = (epicType: EpicType): void => {
+    const isLiveBlog = epicType === 'LiveBlog';
+    this.updateTest(test => ({...test, "isLiveBlog": isLiveBlog}))
+  }
+
+  onTargetRegionsChange = (selectedRegions: Region[]): void => {
+    console.log('onTargetRegionsChange', selectedRegions)
```
tomrf1: remove?

public/src/components/epicTests/epicTestEditor.tsx
```diff
@@ -301,63 +285,44 @@ class EpicTestEditor extends React.Component<EpicTestEditorProps, EpicTestEditor
...
-              >
-                {Object.values(Region).map(region => (
-                  <MenuItem key={region} value={region} >
-                    <Checkbox checked={test.locations.indexOf(region) > -1} />
-                    <ListItemText primary={region} />
-                  </MenuItem>
-                ))}
-              </Select>
-          </FormControl>
+          <TargetRegionsSelector
```
tomrf1: I really like all these new custom components making the parent less cluttered! 👍

public/src/components/epicTests/targetRegionsSelector.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+interface TargetRegionsSelectorProps extends WithStyles<typeof styles> {
+  regions: Region[],
+  onRegionsUpdate: (selectedRegions: Region[]) => void,
+  isEditable: boolean,
+}
+
+interface TargetRegionsSelectorState {
+  selectedRegions: Region[],
+  allRegions: boolean,
```
tomrf1: I think we can implement this with a single field in the state, so just:
```
interface TargetRegionsSelectorState {
  selectedRegions: Region[]
}
```
This is because the other 2 fields are derived from `selectedRegions`, and it's generally preferable to avoid maintaining state based on some other state.

public/src/components/epicTests/targetRegionsSelector.tsx
```diff
@@ -0,0 +1,169 @@
...
+    const { classes } = this.props;
+
+    return (
+      <>
+        <Typography className={classes.selectLabel}>Target regions</Typography>
+          <FormGroup>
+            <FormControlLabel
+              control={
+                <Checkbox
+                  checked={this.state.allRegions}
```
tomrf1: can be:
`checked={this.state.selectedRegions.length === this.allRegionsLength}`

public/src/components/epicTests/targetRegionsSelector.tsx
```diff
@@ -0,0 +1,169 @@
...
+                  onChange={this.onAllRegionsChange}
+                  value={'allRegions'}
+                  indeterminate={this.state.indeterminate}
```
tomrf1: `indeterminate={this.state.selectedRegions.length > 0 && this.state.selectedRegions.length < this.allRegionsLength}`

tomrf1: though this can still be in a separate function. But we don't need to store `indeterminate` in the state

public/src/components/epicTests/targetRegionsSelector.tsx
```diff
@@ -41,89 +41,46 @@ interface TargetRegionsSelectorProps extends WithStyles<typeof styles> {
...
-      selectedRegions: all ? this.allRegions : [],
-    })
-  }
-
-  setIndeterminate = (checked: boolean) => {
-    this.setState({
-      indeterminate: checked
-    })
-  }
+  allRegionsStatus = (checkedRegions: Region[]): boolean => checkedRegions.length === this.allRegions.length;
```
tomrf1: is this being used?

PR #110, opened 2020-03-19
Add a switch to enable use of the dotcom contacts page

rupertbates:
During the Corona virus period we want to update all the contact us information on support-frontend to point to the contact us page on dotcom so that it can be easily updated by central production.

This PR adds a switch which will toggle that behaviour. It also removes the old `optimize` switch as this is no longer used by support-frontend.


rupertbates:
> How well does frontend handle removed or added switches? Just wondering in what situation it would have trouble decoding the file in S3.

It will balk if it's missing but be fine with a new one so I'll add the new one in then deploy then remove old one

PR #111, opened 2020-03-20
Test naming

tjsilver:
## Problem we're trying to solve
- Browsing tests at the menu level isn’t easy / intuitive
- Naming convention isn’t baked into the tool
- No communication that you cannot edit the test name at a later date
- UI could be improved (it’s a small box overlay with a lot going on behind it)

## Actions taken
- Create a focused UI (grey out the background, display the pop up at the center of the page)
- Create 2 fields : one for the name, another for the nickname
- Make sure the name respects the naming convention (and display helper text)
- Use a new CTA with plain english "CREATE YOUR TEST"
- Create a similar UI for new variant without nickname field.

### Before - new test:
image

### After - new test: 
image

### Before - new variant:
image

### After - new variant:
image


public/src/components/epicTests/epicTestEditor.tsx
```diff
@@ -121,6 +121,7 @@ interface EpicTestEditorProps extends WithStyles<typeof styles> {
   isArchived: boolean,
   isNew: boolean,
   testNames: string[],
+  testNicknames: string[],
```
jlieb10: just wondering why are this and `testNames` plural and an array? everything else in this type seems to be describing an individual test rather than a set of tests, but I'm probably just missing some context to understand this.

tjsilver: In order to validate the new name, we have to compare it with the existing names, to make sure that there are no duplicates - so we need to pull in the existing test names and nicknames to compare the new name with.

jlieb10: 👍

app/models/EpicTests.scala
```diff
@@ -60,6 +60,7 @@ case class ArticlesViewedSettings(
 
 case class EpicTest(
   name: String,
+  nickname: String = "",
```
tomrf1: when adding a new field it's preferable to make it optional:
`nickname: Option[String]`

PR #114, opened 2020-04-27
Add model and controller for banner test

JoemG:


app/controllers/BannerTestsController.scala
```diff
@@ -0,0 +1,128 @@
...
+import services.S3Client.S3ObjectSettings
+import io.circe.generic.auto._
+import io.circe.syntax._
+import play.api.libs.circe.Circe
+import zio.blocking.Blocking
+import zio.{DefaultRuntime, IO, ZIO}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object BannerTestsController {
```
tomrf1, line 18: pretty much all of the archive logic here is the same as in the EpicTestsController.
Can we make it a trait, along with the `run` method?
That way `EpicTestsController` and `BannerTestsController` would have nothing in their bodies

app/models/BannerTests.scala
```diff
@@ -0,0 +1,64 @@
...
+  signInUrl: String,
+  secondaryLinkUrl: Option[String],
+  secondaryLinkLabel: Option[String],
+  subsLinkUrl: Option[String]
+)
+
+case class BannerTest(
+  name: String,
+  nickname: Option[String],
+  campaignCode: String,
```
tomrf1, line 44: remove `campaignCode`. For epics it's generated by frontend based on test/variant names, it should be the same for banners

app/models/BannerTests.scala
```diff
@@ -0,0 +1,64 @@
...
+  pageviewId: String,
```
tomrf1, line 45: this is a runtime thing, can be removed

app/models/BannerTests.scala
```diff
@@ -0,0 +1,64 @@
...
+  isOn: Boolean,
+  isHardcodedFallback: Boolean,
```
tomrf1, line 47: not entirely sure why we have `isHardcodedFallback`, but it can be removed from the tool

app/models/BannerTests.scala
```diff
@@ -0,0 +1,64 @@
...
+  minArticlesBeforeShowingBanner: Int,
+  userCohort: UserCohort,
+  bannerModifierClass: Option[String],
```
tomrf1, line 50: this was used for applying css classes for custom design. It can be removed, we'll worry about design another time

app/models/BannerTests.scala
```diff
@@ -0,0 +1,64 @@
...
+  products: Option[List[OphanProduct]],
+  locations: List[Region] = Nil,
+  variants: List[BannerVariant],
+  template: String,
```
tomrf1, line 54: not sure what this is, but it can go

app/models/BannerTests.scala
```diff
@@ -0,0 +1,64 @@
...
+  articlesViewedSettings: Option[ArticlesViewedSettings] = None
```
tomrf1, line 55: Is the `ArticlesViewedSettings` model in the `EpicTest` file? Let's make it a shared file

app/controllers/BannerTestsController.scala
JoemG: @tjsilver shall we have a look at this when you're back ?

app/models/BannerTests.scala
JoemG: This is a placeholder based on the [requirement](https://docs.google.com/document/d/1jgc8nK7fognfXan9OB_g5EOQywOPF4roNKP_3T4Na8Y):
```"Select a banner template 
Display a list of max 3 templates of existing banners 
Select one 
```

Happy to take it out but need to understand the requirement in more detail to suggest a solution

JoemG: update: discussed with @tomrf1  - we are not going to do this for MVP so will remove the field for now.

app/models/BannerTests.scala
```diff
@@ -0,0 +1,64 @@
...
+  case object CONTRIBUTION extends Product
+  case object RECURRING_CONTRIBUTION extends Product
+  case object MEMBERSHIP_SUPPORTER extends Product
+  case object MEMBERSHIP_PATRON extends Product
+  case object MEMBERSHIP_PARTNER extends Product
+  case object DIGITAL_SUBSCRIPTION extends Product
+  case object PRINT_SUBSCRIPTION extends Product
+}
+
+case class BannerVariant(
```
tomrf1, line 24: The model in frontend is a bit of a mess - lots of legacy fields.
Let's focus on the tool requirements for now:
```
BannerVariant(
  name: String,
  headline: Option[String],
  body: String,
  highlightedText: String,
  cta: Option[Cta],
  secondaryCta: Option[Cta],
  hasTicker: Boolean
)
```

tomrf1: we'll do any necessary transformation on frontend, and in the long term change the model there

JoemG: Keep it simple. I like it.

app/controllers/BannerTestsController.scala
tomrf1: This turned out to be quite hard because:
1. it cannot be a class, because we cannot mix in more than 1 class,
1. traits cannot have generic parameters with context bounds

Let's look at it another time...

PR #117, opened 2020-05-01
Adding recaptcha control switches

JoemG:
Adding recaptcha control switches

- We want to be able to turn recaptcha off in support-frontend and payment-api.

image


app/controllers/SwitchesController.scala
```diff
@@ -3,7 +3,8 @@ package controllers
 import com.gu.googleauth.AuthAction
 import models.SupportFrontendSwitches
 import play.api.mvc.{AnyContent, ControllerComponents}
-import io.circe.generic.auto._
```
tomrf1, line 6: 👍 not providing our own decoder/encoder was having some weird effects when the model was changed

PR #118, opened 2020-05-05
Open ELB

tomrf1:
currently it requires VPN to access, which causes a lot of pain, particularly for US colleagues.
Composer does not have this restriction, so I think it's reasonable to trust google auth here as well.


tomrf1:
@paulbrown1982 👍 just checked - only the healthcheck is not behind auth

PR #121, opened 2020-05-14
fix name in home page copy

michaelbjacobson:
Uses correct name (Reader Revenue Control Panel) in the home page. See screenshot below:
<br>
Screenshot 2020-05-14 at 17 00 38


michaelbjacobson:
Merging now and deleting remote branch.

PR #132, opened 2020-08-07
Banner Manager UI - sidebar + stick bottom bar

TomPretty:
## What does this change?

Add the styling/functionality of the sidebar and sticky bottom bar. 

## Images
Screenshot 2020-08-07 at 09 02 24


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "editor.formatOnSave": true,
+}
```
TomPretty, line 3: Should this be in the gitignore? I wasn't sure, so I've just left it here for now

tomrf1: yes - please add to gitignore, I think we should keep IDE-specific files out of here

TomPretty: nice - I've removed it now and update the gitignore

public/src/components/channelManagement/testPriorityLabelListLabel.tsx
```diff
@@ -0,0 +1,55 @@
...
+      padding: "4px 0",
+    },
+    text: {
+      fontSize: "12px",
+      color: palette.grey[700],
+    },
+    dashedLinesContainer: {
+      display: "flex",
+      marginLeft: "4px",
+      "& > * + *": {
```
paulbrown1982, line 24: 😝

PR #139, opened 2020-08-26
Disable test selection when editing and rename 'Cancel' to 'Discard' for clarity

tomrf1:
1. If in edit mode, display an alert if the user tries to select another test (we can revisit this design later):
Screen Shot 2020-08-26 at 15 02 20

2. Say 'discard' instead of 'cancel', to make it clear it's going to lose changes:
Screen Shot 2020-08-26 at 14 58 02


TomPretty (approved):
Looks good 🚀

PR #140, opened 2020-08-27
Tp banner manager editor ui

TomPretty:
## What does this change?
Implement the majority of the styling for the banner manager UI as per [this card](https://trello.com/c/aLBjiWBk/2093-styling-the-ui-for-banner-manager). It also makes some tiny fixes to the epic manager UI, but for now that remains the same as before. 

## Images
banner-manager-ui


public/src/components/channelManagement/hooks/useValidation.ts
```diff
@@ -0,0 +1,30 @@
...
+const useValidation = (onValidationChanged: (isValid: boolean) => void) => {
+  const [validationStatus, setValidationStatus] = useState<ValidationStatus>(
+    {}
+  );
+
+  const [wasValid, setWasValid] = useState(true);
+
+  useEffect(() => {
+    const isValid = allValid(validationStatus);
+    if (isValid !== wasValid) {
```
tomrf1, line 17: is this extra `wasValid` state for performance reasons? Does it change the behaviour of the app?

public/src/components/channelManagement/hooks/useValidation.ts
```diff
@@ -0,0 +1,30 @@
...
+      setWasValid(isValid);
+      onValidationChanged(allValid(validationStatus));
+    }
+  }, [validationStatus]);
+
+  const setValidationStatusForField = (field: string, isValid: boolean) => {
+    setValidationStatus({ ...validationStatus, [field]: isValid });
```
tomrf1, line 24: Is it safer to do this?
```
    setValidationStatus(current => ({
      ...current,
      [fieldName]: isValid,
    }));
```

public/src/components/channelManagement/createTestDialog.tsx
```diff
@@ -0,0 +1,168 @@
...
+    if (mode === "COPY" && copiedTestName) {
+      setName(`Copy of ${copiedTestName}`);
+    }
+  }, []);
+
+  const getDuplicateNicknameError = createGetDuplicateError(existingNicknames);
+  const getNicknameError = (value: string) =>
+    getEmptyError(value) || getDuplicateNicknameError(value);
+
+  const [
```
tomrf1, line 87: All the state returned by `useValidatableField` appears to be used only by a `TextField`.
Why not have a generic text field component with all the state inside? I think this is pretty much what `EditableTextField` does, though as a class component.

public/src/components/channelManagement/hooks/useValidation.ts
```diff
@@ -0,0 +1,30 @@
+import { useState, useEffect } from "react";
+import { ValidationStatus } from "../helpers/validation";
+
+const allValid = (validationStatus: ValidationStatus) => {
+  return Object.values(validationStatus).every((status) => status);
+};
+
+const useValidation = (onValidationChanged: (isValid: boolean) => void) => {
```
tomrf1, line 8: Let's specify the return type for this function, so it's clear how the rest of the program uses this API:
```
type FieldValidationChange = (fieldName: string, isValid: boolean) => void;
const useValidation = (onValidationChanged: (isValid: boolean) => void): FieldValidationChange => { }...
```

public/src/components/channelManagement/hooks/useValidation.ts
tomrf1: discussed IRL, it was triggering an infinite loop without this

public/src/components/channelManagement/hooks/useValidation.ts
TomPretty: 👍

public/src/components/channelManagement/hooks/useValidation.ts
TomPretty: 👍

public/src/components/channelManagement/createTestDialog.tsx
TomPretty: Will investigate this now!

TomPretty: After IRL chat: will revisit this and put in the time to get a nice, general purpose solution

tomrf1 (approved):
brilliant!

PR #141, opened 2020-08-27
Do not save if form has errors

tomrf1:
I think we used to disable the Save button if there were any errors, but since the redesign that has been lost.
The quick fix for now is to alert if the user clicks save but there are errors.
In future we may want to implement something that fits with the design, possibly in `StickyBottomBar`.


TomPretty (approved):
Looks good! 👍

PR #145, opened 2020-08-28
Add eslint + prettier

TomPretty:
## What does this change?
Adds eslint and prettier (run as an eslint rule) to the project. This means running 

```
npm run lint --fix
```

Will show all warnings/erorrs and reformat all files. It should be pretty straight forward to set up an IDE to run this on save. Additionally, linting must pass for the team city build to succeed. 

There a currently 0 errors and 0 warnings! Let's try and keep it that way!
Screenshot 2020-08-28 at 15 37 10


tomrf1 (approved):
🚀

PR #146, opened 2020-09-03
Add basic preview feature

TomPretty:
## What does this change?
Add the ability to preview variants. There are a few things to note here:

- doesn't work in DEV
- the button is disabled in edit mode, as you can't view a test until it's been saved

The current implementation is a bit of a hack. It looks at the current url to figure out if it's running in dev/code/prod. We can refactor to have a backend endpoint e.g. `/banner-preview-url` and do the url construction on the server instead (which already has access to the stage). 
## Images

Screenshot 2020-09-03 at 12 01 23


michaelbjacobson (approved):
All looks good! :shipit:

tomrf1:
🎉

PR #147, opened 2020-09-03
Tp banner manager fixes

TomPretty:
## What does this change?
A handfull of useability fixes and improvements:
- stop saving tests with erros
- select tests on creation
- select variants on creation
- close delete dialog
- names are now case insensitive
- allow swapping to a new test if nothing has been modified

### DIFF
--- a/public/src/components/channelManagement/bannerTests/bannerTestsForm.tsx
+++ b/public/src/components/channelManagement/bannerTests/bannerTestsForm.tsx
@@ -57,8 +57,13 @@ const BannerTestsForm: React.FC<Props> = ({
   editMode,
 }: Props) => {
   const createTest = (name: string, nickname: string): void => {
-    const newTests = [...tests, createDefaultBannerTest(name, nickname)];
-    onTestsChange(newTests, name);
+    if (Object.keys(modifiedTests).length > 0) {
+      alert('Please either save or discard before creating a test.');
+    } else {
+      const newTests = [...tests, createDefaultBannerTest(name, nickname)];
+      onTestsChange(newTests, name);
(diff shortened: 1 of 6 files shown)


public/src/components/channelManagement/bannerTests/bannerTestsForm.tsx
```diff
@@ -57,8 +57,13 @@ const BannerTestsForm: React.FC<Props> = ({
   editMode,
 }: Props) => {
   const createTest = (name: string, nickname: string): void => {
-    const newTests = [...tests, createDefaultBannerTest(name, nickname)];
-    onTestsChange(newTests, name);
+    if (Object.keys(modifiedTests).length > 0) {
```
paulbrown1982, line 60: I think we should refactor `modifiedTests` to be an array in a future PR.

paulbrown1982 (approved):
👍  LGTM, and fine to go out to make the UX more reliable for the first user. However I think we should add a card to refactor the `modifiedTests` object to be an array or structured object.

public/src/components/channelManagement/bannerTests/bannerTestsForm.tsx
TomPretty: Yeah we can actually refactor to get rid of it entirely. It's a hangover from when mutliple tests could be worked on in one go. The new UX is that only one test can be worked on at any one time. Now we can just use a single `modifiedTest` object to keep track of the changes to a single test.

tomrf1: 👍 agreed, let's get rid of `modifiedTest` now that things are simpler

PR #149, opened 2020-09-07
Swap from ExpansionPanel to Accordion

TomPretty:
## What does this change?
Material UI renamed ExpansionPanel to Accordion. This PR migrates all ExpansionPanel components over to Accordions


michaelbjacobson (approved):
💯 💯 💯

PR #150, opened 2020-09-08
Add react-hook-form validation

TomPretty:
## What does this change?
Use `react-hook-form` for validation. This allows us to remove boilerplate and eventually remove some components like `EditableTextField`. This PR also lays the groundwork for changing how validation works in the app and removing the concept of `modifiedTests` (will be addressed in a follow up PR).

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "react": "^16.9.0",
     "react-beautiful-dnd": "^13.0.0",
     "react-dom": "^16.9.0",
+    "react-hook-form": "^6.7.2",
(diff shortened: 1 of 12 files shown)


tomrf1 (approved):
This looks positive.
Can all of our custom validation logic be removed?

TomPretty:
> Can all of our custom validation logic be removed?

Yeah should be able to. We should eventually be able to remove `EditableTextField` and `useValidatableField`, potentially `useValidation` - should be good!

PR #151, opened 2020-09-09
Add preview to epic and refactor to get stage from server

TomPretty:
## What does this change?
Add preview link to epic variants. Additionally refactor the logic to get the current stage from the server instead of the url.

## Images
Screenshot 2020-09-09 at 16 17 57


paulbrown1982 (approved):
👍 🎉  Looks good to me! I wonder if we should break the rule around enums being capital? Then you won't have to write them in uppercase and have code to lower-case them!

tomrf1 (approved):
great!

PR #152, opened 2020-09-10
Remove logic that caused bug in epic manager

TomPretty:
## What does this change?
Removes some logic that was causing a bug in the epic manager. The current version in prod considers a test modified as soon as it has been selected (even though you're in read-only mode). This leads to a really bad UX!  The logic was added initially to catch an edge case that allowed an invalid test to be saved. I think it's best to remove it for now, and re-add/re-write it after we overhaul the epic manager

## Images
Screenshot 2020-09-10 at 08 25 13

### DIFF
--- a/public/src/components/channelManagement/testEditor.tsx
+++ b/public/src/components/channelManagement/testEditor.tsx
@@ -211,18 +211,6 @@ const TestEditor = <T extends Test>(
             },
           },
         });
-      } else {
-        this.setState({
-          modifiedTests: {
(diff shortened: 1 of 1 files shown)


tomrf1 (approved):
👍 assuming it still works correctly when in edit mode?

TomPretty:
Yeah that's right - it's now reverting back to how it was before I made any changes. We can improve all this logic after we redo the epic manager.

PR #154, opened 2020-09-10
Use regions prop instead of state

TomPretty:
## What does this change?
Quick fix to sovle the bug in which selecting a new test wouldn't update the list of target regions. This was happening because the component is stateful and was using it's own state to decide which boxes are ticked. When the test changed, it didn't trigger an update of state. To fix this, I've removed any state and now the component just uses the props passed down to it to render the checkboxes.

The proper fix to this is to remove this component and reuse the one implemented for the banner manager. This will be done when we overhaul the epic manager.


paulbrown1982 (approved):
👍  Good catch

tomrf1 (approved):
thanks, not sure why this was stateful

PR #155, opened 2020-09-11
Add BannerTemplate field

tomrf1:
The banner tool will soon be used for Subscriptions banners as well as contributions.
This change adds a radio group for each variant for setting a new `template` field.
Also, the highlightedText field is hidden for subscriptions banners.
Screen Shot 2020-09-14 at 10 48 41

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "build-dev": "webpack --config webpack.dev.js",
     "watch": "webpack --config webpack.dev.js --watch",
     "lint": "eslint . --ignore-path .gitignore --ext .js,.jsx,.ts,.tsx",
+    "fix": "npm run lint -- --fix",
     "test": "echo 'Running JS tests' && jest"
   },
   "jest": {
--- a/public/src/components/channelManagement/bannerTests/bannerTestVariantEditor.tsx
+++ b/public/src/components/channelManagement/bannerTests/bannerTestVariantEditor.tsx
@@ -9,9 +9,12 @@ import {
   withStyles,
 } from '@material-ui/core';
 import VariantEditorButtonsEditor from '../variantEditorButtonsEditor';
-import { BannerVariant } from './bannerTestsForm';
+import {BannerTemplate, BannerVariant} from './bannerTestsForm';
 import { invalidTemplateValidator, EMPTY_ERROR_HELPER_TEXT } from '../helpers/validation';
 import { Cta } from '../helpers/shared';
+import RadioGroup from "@material-ui/core/RadioGroup";
+import Radio from "@material-ui/core/Radio";
+import FormControlLabel from "@material-ui/core/FormControlLabel";
 
 // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
 const styles = ({ palette, spacing }: Theme) =>
@@ -35,10 +38,19 @@ const styles = ({ palette, spacing }: Theme) =>
         marginTop: spacing(3),
       },
     },
-    buttonsSectionHeader: {
+    sectionHeader: {
       fontSize: 16,
       color: palette.grey[900],
+      fontWeight: 500,
     },
+    sectionContainer: {
+      paddingTop: spacing(1),
+      paddingBottom: spacing(2),
+      borderBottom: `1px solid ${palette.grey[500]}`,
+    },
+    templates: {
+      marginTop: '15px',
+    }
   });
 
 const HEADER_DEFAULT_HELPER_TEXT = 'Assitive text';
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -36,6 +36,7 @@
     "build-dev": "webpack --config webpack.dev.js",
     "watch": "webpack --config webpack.dev.js --watch",
     "lint": "eslint . --ignore-path .gitignore --ext .js,.jsx,.ts,.tsx",
+    "fix": "npm run lint -- --fix",
```
TomPretty, line 39: I think we don't want to run fix by default, as we want the build to fail when committing code that doesn't pass the linting

public/src/components/channelManagement/bannerTests/bannerTestVariantEditor.tsx
```diff
@@ -90,8 +102,41 @@ const BannerTestVariantEditor: React.FC<BannerTestVariantEditorProps> = ({
...
+          {Object.values(BannerTemplate).map(bannerTemplate => (
+            <FormControlLabel
+              key={bannerTemplate}
+              value={bannerTemplate}
+              control={<Radio disabled={!editMode} />}
+              label={bannerTemplate}
+            />
+          ))}
+        </RadioGroup>
+      </div>
```
TomPretty, line 138: Maybe we could factor this into its own component?

public/src/components/channelManagement/bannerTests/bannerTestVariantsEditor.tsx
```diff
@@ -1,6 +1,6 @@
-import React, { useState, useEffect } from 'react';
-import { Theme, createStyles, WithStyles, withStyles } from '@material-ui/core';
-import { BannerVariant } from './bannerTestsForm';
+import React, { useEffect, useState } from 'react';
+import { createStyles, Theme, WithStyles, withStyles } from '@material-ui/core';
+import { BannerTemplate, BannerVariant } from './bannerTestsForm';
```
TomPretty, line 3: Do you have something to auto sort alphabetically? If so hit me up! 🔥

TomPretty:
The design looks like the selection is on a per test basis, not per variant basis?

package.json
tomrf1: yes I wasn't planning on adding this to the build, it's just so I don't have to remember the weird syntax

public/src/components/channelManagement/bannerTests/bannerTestVariantsEditor.tsx
tomrf1: must be intellij!

tomrf1:
> The design looks like the selection is on a per test basis, not per variant basis?

yes the design is incorrect, it should be per-variant

package.json
TomPretty: 👍

public/src/components/channelManagement/bannerTests/bannerTestVariantsEditor.tsx
TomPretty: I'm jealous!

public/src/components/channelManagement/bannerTests/bannerTestVariantEditor.tsx
tomrf1: Good question, I'm undecided.
It would be neater here, but it's also a lot of boilerplate for a small component, e.g. it needs its own `classes`, we have to pass in the callback etc.
Also, if something like this is moved out, should the container/heading come with it, or remain in the parent?

TomPretty: Yeah I hate that Material UI boilerplate 😕. I think if we wanted to 'prettify' the labels (i.e. ContributionsBanner -> Contributions, DigitalSubscriptionsBanner -> Subscriptions banner) it might be just enough logic to justify it. Good question! I've been leaving the header and container in the parent i.e

```jsx
      <div className={classes.buttonsSectionContainer}>
        <Typography className={classes.buttonsSectionHeader} variant="h4">
          Buttons
        </Typography>

        <VariantEditorButtonsEditor
          primaryCta={variant.cta}
          secondaryCta={variant.secondaryCta}
          updatePrimaryCta={updatePrimaryCta}
          updateSecondaryCta={updateSecondaryCta}
          isDisabled={!editMode}
          onValidationChange={onValidationChange}
        />
      </div>
```

This is from the same file. So Maybe it'd be something like

```jsx
      <div className={classes.sectionContainer}>
        <Typography className={classes.sectionHeader} variant="h4">
          Banner template
        </Typography>

        <BannerVariantEditorTemplateEditor
          template={variant.template}
          updateTemplate={updateTemplate}
        />
      </div>
```

I'm imagining we'd rename `buttonSectionContainer` and `buttonSectionHeader` to `sectionContainer` and `sectionHeader` as I think they can probably share styles (I think Hyuna would have tweaked the design slightly if she'd done it per variant)

tomrf1: I had forgotten about prettyfying the labels! I think that's enough justification for a separate component

public/src/models/banner.ts
```diff
@@ -0,0 +1,33 @@
+import {
+  ArticlesViewedSettings,
+  Cta,
+  Test,
+  UserCohort,
+} from '../components/channelManagement/helpers/shared';
+import { Region } from '../utils/models';
+
+export enum BannerTemplate {
```
tomrf1, line 9: I had to move the banner models into a new file because a bug in webpack was causing import errors at runtime in specific places.
We should move all models into this directory, but I don't want this PR to get too big!

public/src/components/channelManagement/bannerTests/bannerTestVariantEditor.tsx
tomrf1: I've moved it into a separate file (but had to refactor the models because of a weird webpack bug)

i-hardy (approved):
Looks great to me, but happy to defer to @TomPretty if he has any remaining concerns.

TomPretty (approved):
This is so sweet 🔥 All good on my end!

PR #156, opened 2020-09-14
Second banner tests list

tomrf1:
We are adding support for the subscriptions banner in the tool.
For the initial implementation we are temporarily having 2 banner lists.
This is because we currently have the ability to dismiss the contributions banner and the subscriptions banner separately (the client stores 2 separate timestamps when these banners are dismissed).
Having 2 separate banner test lists allows us to continue doing this, for now.

So, in the code `banner-tests2` is identical to `banner-tests`, except it writes to a new file in S3.
`support-dotcom-components` can then poll both S3 files, to get both banner test lists.

Screen Shot 2020-09-11 at 15 39 29

Before deploying to CODE + PROD:

- [x] create `support-admin-console/<STAGE>/locks/banner-tests2.lock` in S3
- [x] create `gu-contributions-public/banner/<STAGE>/banner-tests2.json` in S3

### DIFF
--- a/app/controllers/BannerTestArchiveController2.scala
+++ b/app/controllers/BannerTestArchiveController2.scala
@@ -1,0 +1,24 @@
+package controllers
+
+import com.gu.googleauth.AuthAction
+import models.BannerTest
+import models.BannerTests._
+import play.api.libs.ws.WSClient
+import play.api.mvc.{AnyContent, ControllerComponents}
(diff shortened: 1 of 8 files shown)


TomPretty (approved):
Looks good! 🔥

i-hardy (approved):
Looks good! :shipit:

PR #158, opened 2020-09-22
Tp banner manager bug fixes

TomPretty:
## What does this change?
Fixes a couple of bugs with the banner manager.

### Variant editor

The variant editor wouldn't update if a new test was selected, but the variant had the same name. This has been fixed by scoping the key by the testName (which is unique).

### Article count editor

The article count editor wouldn't update when selecting a new test. This was because the defaultValues is cached by react-hook-form. Now whenever any of those values change, we reset the form to the new values inside a useEffect hook.

Also, always shrink the label to avoid and issue where the label was overlapping the text

Screenshot 2020-09-22 at 09 30 35

### DIFF
--- a/public/src/components/channelManagement/testEditorArticleCountEditor.tsx
+++ b/public/src/components/channelManagement/testEditorArticleCountEditor.tsx
@@ -62,7 +62,14 @@ const TestEditorArticleCountEditor: React.FC<TestEditorArticleCountEditorProps> 
     periodInWeeks: articlesViewedSettings?.periodInWeeks.toString() || '',
   };
 
-  const { register, errors, handleSubmit } = useForm<FormData>({ mode: 'onChange', defaultValues });
+  const { register, errors, handleSubmit, reset } = useForm<FormData>({
+    mode: 'onChange',
+    defaultValues,
+  });
+
+  useEffect(() => {
+    reset(defaultValues);
+  }, [defaultValues.minViews, defaultValues.maxViews, defaultValues.periodInWeeks]);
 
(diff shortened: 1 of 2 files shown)


public/src/components/channelManagement/testEditorArticleCountEditor.tsx
```diff
@@ -62,7 +62,14 @@ const TestEditorArticleCountEditor: React.FC<TestEditorArticleCountEditorProps>
...
 
-  const { register, errors, handleSubmit } = useForm<FormData>({ mode: 'onChange', defaultValues });
+  const { register, errors, handleSubmit, reset } = useForm<FormData>({
+    mode: 'onChange',
+    defaultValues,
+  });
+
+  useEffect(() => {
+    reset(defaultValues);
+  }, [defaultValues.minViews, defaultValues.maxViews, defaultValues.periodInWeeks]);
```
tomrf1, line 72: is `defaultValues` the current state of the form fields?

TomPretty: it's the state when the form is first created - after a user starts interacting with the form react-hook-form handles the state internally

tomrf1: ok, so this `useEffect` block will only execute once, at the start?

TomPretty: yeah that's right! it should execute each time the user selects a new test (unless the test had identical settings)

tomrf1: 👍

tomrf1 (approved):
thanks!

PR #160, opened 2020-09-29
Epic/Banner manager bug fixes

TomPretty:
## What does this change?
Fix a number of bugs with the banner manager, as per [this card](https://trello.com/c/muzZ8j5M/2281-epic-banner-manager-parent-card-of-bugs-and-fixes). The fixes are:

- automatically selecting newly created tests
- copying a test will always create a draft test
- remove placeholder values when copying a test
- set default value for page views (depending on banner channel)
- force upper case test name/nickname and variant name
- stop requiring clicking save twice**

** the solution to this one was a bit of a hack. This happened when you had a text field focused and then clicked save. The issue is that the text fields have an `onBlur` handler that updates the state. This then leads to a rerender, meaning the button never gets a click event. But, after the rerender the button does receive a mouseup event. We now use this to trigger saving. Ideally, we would come up with a better solution to cut down on pointless rerenders. For more info, check out [this article](https://www.eventbrite.com/engineering/a-story-of-a-react-re-rendering-bug/).


tomrf1 (approved):
great stuff!

PR #163, opened 2020-10-06
Add liveblog epic channel

TomPretty:
## What does this change?
Create a new liveblog epic channel. As liveblog epics don't compete with regular epics and have slightly different config it makes sense to split them out. Currently the liveblog channel is hidden in the drawer. This is so we can deploy and test by going directly to the endpoint before anybody starts using it

## Follow up work

- migrate over liveblog epics (manually updating the json files in s3)
- re-enable the the channel in the drawer
- remove epic type editor

### DIFF
--- a/app/controllers/LiveblogEpicTestsController.scala
+++ b/app/controllers/LiveblogEpicTestsController.scala
@@ -33,6 +33,6 @@ class LiveblogEpicTestsController(
       cacheControl = Some("max-age=30"),
       surrogateControl = Some("max-age=86400")  // Cache for a day, and use cache purging after updates
     ),
-    fastlyPurger = FastlyPurger.fastlyPurger(stage, s"${EpicTestsController.name}.json", ws),
+    fastlyPurger = None,
     runtime = runtime
   ) with Circe
--- a/app/controllers/LiveblogEpicTestsController.scala
+++ b/app/controllers/LiveblogEpicTestsController.scala
@@ -33,6 +33,6 @@ class LiveblogEpicTestsController(
       cacheControl = Some("max-age=30"),
       surrogateControl = Some("max-age=86400")  // Cache for a day, and use cache purging after updates
     ),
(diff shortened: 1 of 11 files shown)


public/src/components/channelManagement/epicTests/epicTestEditor.tsx
```diff
@@ -274,12 +269,6 @@ class EpicTestEditor extends React.Component<EpicTestEditorProps, TestEditorStat
 
         <hr />
 
-        <EpicTypeComponent
```
tomrf1, line 277: I think we still need this until the migration is complete?
Can we leave it in for now, and remove at the same time as we add the channel to the menu?

TomPretty: Ahhh yeah of course!

app/controllers/LiveblogEpicTestsController.scala
```diff
@@ -0,0 +1,38 @@
...
+    stage,
+    name = LiveblogEpicTestsController.name,
+    dataObjectSettings = S3ObjectSettings(
+      bucket = "gu-contributions-public",
+      key = s"epic/$stage/${LiveblogEpicTestsController.name}.json",
+      publicRead = true,  // This data will be requested by dotcom
+      cacheControl = Some("max-age=30"),
+      surrogateControl = Some("max-age=86400")  // Cache for a day, and use cache purging after updates
+    ),
+    fastlyPurger = FastlyPurger.fastlyPurger(stage, s"${EpicTestsController.name}.json", ws),
```
tomrf1, line 36: This can be `fastlyPurger = None,`
We needed it for epic-tests.json because it was being requested directly by clients via fastly, but there's no need for that with this file

TomPretty: 👍

tomrf1 (approved):
💯

PR #167, opened 2020-10-16
Fix scroll on amounts page

tomrf1:
The main container now has `overflow: 'hidden'`, which breaks the amounts page.


TomPretty (approved):
LGTM 👍

PR #168, opened 2020-10-22
Tp style epic manager

TomPretty:
## What does this change?
Style the epic manager to bring it inline with the banner manager, as per [this card](https://trello.com/c/qZFVtYpL/2297-finish-the-work-to-make-epic-manager-match-with-the-banner-manager-ui). The majority of the work here was converting class components into function components, updating to use the new validation, and trying to share as much as possible with the variants list editor. There's still a lot of duplication between `EpicTestEditor` and `BannerTestEditor` that we might be able to reduce somehow?

## Images
screencapture-localhost-9000-epic-tests-2020-10-22-09_06_09


public/src/components/channelManagement/testEditorArticleCountEditor.tsx
```diff
@@ -24,13 +24,15 @@ const styles = ({ spacing }: Theme) =>
...
       '& > * + *': {
         marginTop: spacing(1),
       },
     },
   });
 
-const DEFAULT_ARTICLES_VIEWED_SETTINGS: ArticlesViewedSettings = {
+export const DEFAULT_ARTICLES_VIEWED_SETTINGS: ArticlesViewedSettings = {
   minViews: 3,
   maxViews: 999,
```
tomrf1, line 37: can this be undefined by default? This field should be optional

tomrf1:
Testing this in CODE I noticed that the validation does not work for a new variant. E.g I create a new variant, do not set any fields, but it allows me to save. The existing behaviour is to highlight mandatory fields red, and not allow the user to save.

TomPretty:
I think the fix to this turned out to be quite straight forward:

```js
 const { register, handleSubmit, errors, trigger } = useForm<FormData>({
    mode: 'onChange',
    defaultValues,
  });

  useEffect(() => {
    trigger();
  }, []);
```

`react-hook-form` also gives you this `trigger` method to manually run validation. If we do that when the variant editor is created it will store it wont let you save it until you fix the errors!

I'll redeploy to CODE when it's finished building

public/src/components/channelManagement/testEditorArticleCountEditor.tsx
TomPretty: Both `minViews` and `maxViews` are optional - should we set both to undefined? or just `maxViews`?

tomrf1: Let's give `minViews` a default as it's the one that's always used

TomPretty: sure thing!

tomrf1 (approved):
Great work.
Very happy to have functional components, and I love the new generic components!

PR #169, opened 2020-10-23
Update deploy date display

TomPretty:
## What does this change?
Fixes + updates the banner deploy date. 

There were a couple of bugs:
- we used `getUTCDay` (returns a number from 0-6 which is the current day of week) instead of `getUTCDate` (returns a number from 1-31)
- we didn't +1 to `getUTCMonth` (javascript months run from 0-11)

The update we made was to invert the order so now we have `year/month/day - hour:minute` instead of `hour: minute - year/month/day` to avoid any confusion with UK/US date formats

## Images

before:
Screenshot 2020-10-23 at 16 13 27

after:
Screenshot 2020-10-23 at 16 15 23

(the deploy date was Friday the 23rd October 2020)


tomrf1 (approved):
👍

PR #171, opened 2020-10-27
Make label always at the top

TomPretty:
## What does this change?
Follow up fix for #170. After that fix, if you have a test selected that doesn't have any target tags (so that the field label is in the middle of the field) and then swap to a test that does have some, the label doesn't move, and gets in the way of the content.

Screenshot 2020-10-27 at 08 44 26

As a quick fix, we now just always force the label to be in the top corner.


tomrf1 (approved):
👍

PR #172, opened 2020-10-27
Fix parsing of list fields

TomPretty:
## What does this change?
Before list fields were parsed just using `.split(',')`, which meant an empty string was parsed as `['']`. Now we filter that list to remove an blank entries. 

TODO
- manually fix the affected tests


tomrf1 (approved):
👍

PR #175, opened 2020-11-03
Add Apple News epic manager

TomPretty:
## What does this change?
Add Apple News epic manager to the channel management section. Apple News doesn't allow for A/B testing, but we still leverage our existing code/types hence they're implemented as `EpicTests` with a single `EpicVariant`. 

The main differences are:
- apple news tests are created along with a variant (we could make this the default for all tests?)
- there's no accordian, instead we have the variant editor directly in the form
- there are far fewer fields to play with

## AWS

The `apple-news-epic-tests.json` file has been created in AWS for DEV + CODE + PROD. 

## Images
screencapture-localhost-9000-apple-news-epic-tests-2020-11-03-11_48_42


public/src/components/channelManagement/epicTests/epicTestsForm.tsx
```diff
@@ -67,12 +75,23 @@ export interface EpicTest extends Test {
...
 const getEpicTestForm = (epicType: EpicType): React.FC<Props> => {
   const isLiveBlog = epicType === 'LIVEBLOG';
+  const isAppleNews = epicType == 'APPLE_NEWS';
+
+  const createDefaultEpicVariant = (): EpicVariant => ({
+    name: 'CONTROL',
+    heading: undefined,
+    paragraphs: [],
+    highlightedText:
+      'Support the Guardian from as little as %%CURRENCY_SYMBOL%%1 – and it only takes a minute. Thank you.',
```
tomrf1, line 89: Currently we only have a US channel in AN.
But I think it would be good to support `%%CURRENCY_SYMBOL%%` anyway, for 2 reasons:
1. consistency with other epic tools. Someone may quite reasonably think it needs to be templated.
2. one day we might spin up another regional AN channel

So let's handle the `%%CURRENCY_SYMBOL%%` template in the apple-news backend

TomPretty: Would this involve publishing different versions of the article to different channels?

tomrf1: Yes, and we can worry about that when we have more channels!

tomrf1 (approved):
💯

PR #176, opened 2020-11-03
Fix epic preview

TomPretty:
## What does this change?
Fix the epic preview feature. Before, the `TestEditorVariantSummary` component used the hard coded value `"BANNER"` as the `testType`. I didn't spot this when implementing the new components for the epic manager which means the preview links for epics have `force-banner` in the url, instead of `force-epic`. Now the `testType` is passed all the way down from the `EpicTestEditor`/`BannerTestEditor`.


public/src/components/channelManagement/testEditorVariantSummary.tsx
```diff
@@ -65,7 +68,7 @@ const TestEditorVariantSummary: React.FC<TestEditorVariantSummaryProps> = ({
         <TestEditorVariantSummaryPreviewButton
           name={name}
           testName={testName}
-          testType="BANNER" // will have to change this to make it more reusable
```
tomrf1, line 68: 😀

tomrf1 (approved):
👍

paulbrown1982 (approved):
👍

PR #177, opened 2020-11-05
Editor for epic on AMP

tomrf1:
Similar to the Apple News tool, but instead of targeting by tag we only target by region.
In V2 we can support tags, but it's more complicated.

I also reorganised the /controllers directory, as there are a lot of files in there now

Screen Shot 2020-11-05 at 14 48 42
Screen Shot 2020-11-05 at 15 31 40


app/controllers/epic/AppleNewsEpicTestsController.scala
```diff
@@ -20,10 +24,10 @@ class AppleNewsEpicTestsController(
   authAction,
   components,
   stage,
-  name = LiveblogEpicTestsController.name,
+  name = AppleNewsEpicTestsController.name,
```
paulbrown1982, line 27: Good spot.

paulbrown1982 (approved):
👍  LGTM, much better having the controllers namespaced.

tomrf1:
Just pushed up a change after I realised the targeting needs to be different than Apple News.
In V1, for AMP we target by region but not tag.
Tags will come later

PR #182, opened 2020-11-17
Refactor testEditor 

TomPretty:
## What does this change?

Refactor the `testEditor` higher order component. This started out as [this card](https://trello.com/c/NZACHuoB/2405-epic-banner-tool-prevent-save-if-there-are-invalid-fields) but ended up being a little more!

The initial bug was caused by legacy logic in `testEditor` that allowed for editing multiple tests in one session before saving. To fix this I removed all that logic - we now just use the flags `selectedTestIsValid` and `selectedTestHasBeenModified` to keep track of the selected test status.

I also took this as an opportunity to refactor `testEditor` to return a functional component that uses hooks instead of a class component.

### DIFF
--- a/public/src/components/channelManagement/testEditor.tsx
+++ b/public/src/components/channelManagement/testEditor.tsx
@@ -103,7 +103,7 @@ const TestEditor = <T extends Test>(
       fetchStateFromServer();
     }, []);
 
-    const save = (updatedTests: T[]): Promise<void> => {
+    const save = async (updatedTests: T[]): Promise<void> => {
       const postData = {
         version: version,
         value: {
@@ -111,7 +111,7 @@ const TestEditor = <T extends Test>(
         },
       };
 
-      return saveFrontendSettings(settingsType, postData)
+      saveFrontendSettings(settingsType, postData)
         .then(resp => {
           if (!resp.ok) {
             resp.text().then(msg => alert(msg));
--- a/public/src/components/channelManagement/testEditor.tsx
+++ b/public/src/components/channelManagement/testEditor.tsx
@@ -22,7 +22,7 @@ export interface InnerComponentProps<T extends Test> {
   onTestArchive: () => void;
   onTestCreate: (newTest: T) => void;
   onTestPriorityChange: (newPriority: number, oldPriority: number) => void;
-  onSelectedTestName: (testName: string) => void;
+  onTestSelected: (testName: string) => void;
   onTestErrorStatusChange: (isValid: boolean) => void;
   cancel: () => void;
   requestTakeControl: () => void;
@@ -258,7 +258,7 @@ const TestEditor = <T extends Test>(
               onTestDelete={onTestDelete}
               onTestArchive={onTestArchive}
               onTestCreate={onTestCreate}
-              onSelectedTestName={onSelectedTestName}
+              onTestSelected={onSelectedTestName}
               requestTakeControl={requestTestsTakeControl}
               requestLock={requestTestsLock}
               lockStatus={lockStatus}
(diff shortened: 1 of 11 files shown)


public/src/components/channelManagement/testEditor.tsx
```diff
@@ -41,279 +40,240 @@ interface DataFromServer<T extends Test> {
...
+    }
+  }, [editMode]);
+};
 
 /**
  * A stateful higher-order component for fetching/saving test data.
  * Takes care of locks
  */
-// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
 const TestEditor = <T extends Test>(
```
tomrf1, line 74: too late now but this would have been much easier to follow if the functional component refactor had been done separately from the modified/valid logic change

TomPretty: yeah for sure - I got carried away. I'll be sure to be more thoughtful of that going forward!

public/src/components/channelManagement/testEditor.tsx
```diff
@@ -41,279 +40,240 @@ interface DataFromServer<T extends Test> {
...
-          );
-          this.setState({ timeoutAlertId: null });
-        }, 60 * 20 * 1000);
-
-        this.setState({ timeoutAlertId });
-      } else if (this.state.timeoutAlertId) {
-        window.clearTimeout(this.state.timeoutAlertId);
-        this.setState({ timeoutAlertId: null });
-      }
+    const save = async (updatedTests: T[]): Promise<void> => {
```
tomrf1, line 106: does this need to be `async`?

public/src/components/channelManagement/testEditor.tsx
```diff
@@ -41,279 +40,240 @@ interface DataFromServer<T extends Test> {
...
-      lockStatus: { locked: false },
-      modifiedTests: {},
-      timeoutAlertId: null,
-    };
+): React.FC => {
+  const WrappedInnerComponent: React.FC = () => {
+    const [tests, setTests] = useState<T[] | null>(null);
+    const [version, setVersion] = useState<string | null>(null);
+    const [editMode, setEditMode] = useState<boolean>(false);
+    const [selectedTestName, setSelectedTestName] = useState<string>('');
```
tomrf1, line 82: would `string | null` be more consistent?

public/src/components/channelManagement/testEditor.tsx
TomPretty: Originally I had 
```js
    const save = (updatedTests: T[]): Promise<void> => {
      const postData = {
        version: version,
        value: {
          tests: updatedTests,
        },
      };

      return saveFrontendSettings(settingsType, postData)
        .then(resp => {
          if (!resp.ok) {
            resp.text().then(msg => alert(msg));
          }
        })
        .catch(() => {
          alert('Error while saving');
        })
        .then(fetchStateFromServer);
    };
```
But then vs code gave me a squiggly underline saying it could be made async - I'm not 100% sure if that's coming from how I have vs code set up or if that's part of the linting on the project

public/src/components/channelManagement/testEditor.tsx
TomPretty: Yeah makes sense - I'll do that now 👍

public/src/components/channelManagement/testEditor.tsx
tomrf1: Weird, I just tried it in intellij and it's happy.
Returning a Promise from a non-async function is perfectly fine. I don't see the benefit in marking something as async if you're not using await.

tomrf1 (approved):
great work! It's so nice to have the modifiedTests stuff gone, it was a big cause of complexity in the original implementation!

public/src/components/channelManagement/testEditor.tsx
TomPretty: I think it's a vs code thing - I'll remove the async 👍

PR #186, opened 2020-12-10
Update template validation 

TomPretty:
## What does this change?
Iona ran into an issue this morning that turned out to be due to accidently using`%%ARTICLE COUNT%%` instead of `%%ARTICLE_COUNT%%`. I figured the easiest thing to do would be to just match on anything between two sets of `%%`s. 

I very briefly played around with the [string-similarity](https://www.npmjs.com/package/string-similarity) npm package to figure out the closest match to provide a `did you mean...` type error message. It was neat, but as it was running on every character input it made the editing experience really laggy. I think we could maybe solve this with a simpler algorithm/debouncing the input to avoid accessive calls (which shouldn't be too difficult with react-hook-form), but didn't seem worth the effort for now!


tomrf1 (approved):
👍

PR #187, opened 2020-12-11
Make CODE a little safer

TomPretty:
## What does this change?
It was pointed out that we should maybe make out CODE tests a little more realistic, just incase they ever slip into production. There are two things in this PR to hopefully make things a little safer:

- custom colour scheme in DEV/CODE to make it more distinct from PROD
- default tests/variants in DEV/CODE to allow devs to quickly save realistic looking tests whilst testing

TODO:
- After merging, delete the existing DEV/CODE tests and replace them with out more realistic placeholders

## Images
Screenshot 2020-12-11 at 11 51 07

### DIFF
--- a/app/views/index.scala.html
+++ b/app/views/index.scala.html
@@ -12,9 +12,9 @@
   </head>
   <body>
     <div id="root"></div>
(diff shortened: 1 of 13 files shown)


tomrf1 (approved):
this is great!

PR #190, opened 2021-01-13
Add amounts testing 

TomPretty:
## What does this change?
Allows for amounts tests on the contributions LP to be configured from support-admin-console, as per [this card](https://trello.com/c/5KqrDNqc/2340-implement-amounts-testing-on-the-rrcp). The matching support-frontend PR is [here](https://github.com/guardian/support-frontend/pull/2917).

It's just going to be Jesse using this, so UX isn't a massive priority, but it might be worth adding some sort of saving indicator as there's nothing atm!

## Images
Screenshot 2021-01-13 at 15 49 00


public/src/hocs/withS3Data.tsx
```diff
@@ -0,0 +1,59 @@
+import React, { useState, useEffect } from 'react';
```
TomPretty, line 1: I was thinking that we could reuse this. I think it would be quite straight forward to get the banner deploy dashboard using this. Might even be possible to get the banner/epic editors using this too.

app/controllers/AmountsController.scala
```diff
@@ -1,13 +1,13 @@
 package controllers
 
 import com.gu.googleauth.AuthAction
-import models.AmountsRegions
+import models.ConfiguredAmounts
+import models.ConfiguredAmounts._
 import play.api.mvc.{AnyContent, ControllerComponents}
-import io.circe.generic.auto._
```
tomrf1, line 6: it turns out that asking circe to do its magic here is a bit unreliable, so we now define the Encoder/Decoder as implicit vals and import them here

public/src/components/amounts/amountsEditorRow.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+const AmountsEditorRow: React.FC<AmountsEditorRowProps> = ({
+  label,
+  amountsSelection,
+  updateSelection,
+}: AmountsEditorRowProps) => {
+  const classes = useStyles();
+
+  const addAmount = (amount: number): void => {
+    // this is already a sorted list, so we could be cleverer here, but this sure takes less code!
```
tomrf1, line 56: Given the tiny array size I think we can live without optimisation 😄

public/src/hocs/withS3Data.tsx
tomrf1: Interesting, it feels like a more general form of `TestEditor`?

public/src/components/shared/liveSwitch.tsx
```diff
@@ -0,0 +1,58 @@
...
+}));
+
+interface LiveSwitchProps {
+  isLive: boolean;
+  label: string;
+  isDisabled: boolean;
+  onChange: (isLive: boolean) => void;
+}
+
+const LiveSwitch: React.FC<LiveSwitchProps> = ({
```
tomrf1, line 34: 👍

public/src/utils/forms.ts
```diff
@@ -0,0 +1,20 @@
...
+export const VALID_CHARACTERS_REGEX = /^[\w-]+$/;
+
+export const EMPTY_ERROR_HELPER_TEXT = 'Field cannot be empty - please enter some text';
+
+export const DUPLICATE_ERROR_HELPER_TEXT = 'Name already exists - please try another';
+
+export const duplicateValidator = (existingNames: string[]): Validator => (
+  name: string,
+): string | boolean => {
+  const existingNamesLowerCased = existingNames.map(n => n.toLowerCase());
```
tomrf1, line 15: not at all important, but `some` can save making 2 passes here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

tomrf1 (approved):
This is great!
And the fact that it maintains a different file in s3 means you can merge before updating support-frontend 👍

public/src/hocs/withS3Data.tsx
TomPretty: Yeah exactly!

public/src/utils/forms.ts
TomPretty: oh nice, so...
```js
export const duplicateValidator = (existingNames: string[]): Validator => (
  name: string,
): string | boolean => {
  if (existingNames.some(n => n.toLowerCase() === name.toLowerCase())) {
    return DUPLICATE_ERROR_HELPER_TEXT;
  }
  return true;
};
```

PR #191, opened 2021-01-20
Add Epic Holdback Tests

TomPretty:
## What does this change?
Adds a new channel in the tool for epic holdback tests. These are treated as a normal list of tests, with the restriction that a user has no control over the variants. A single variant named 'CONTROL' is created when a new test is created (same as for Apple News too). Over on the dotcom-components side, based on a users mvtId we will funnel 1% of our audience through this holdback stack instead of the regular one. 

I think the logic for hiding/showing different parts of the epic editor is getting a little tricky to read. I think this is because we encode the type of the epic through an `EpicType` and then make the editor handle the logic based on that type. I reckon we can invert this, and instead provide and `EpicConfig` that will configure which parts of the tool to show/hide e.g.

```ts
interface EpicConfig {
  onlyShowOneVariant: boolean;
  allowCustomVariantSplit: boolean;
  /* ... */
}

const ARTICLE_EPIC_CONFIG : EpicConfig = {
  onlyShowOneVariant: false,
  allowCustomVariantSplit: true,
  /* ... */
}

const LIVEBLOG_EPIC_CONFIG : EpicConfig = {
  onlyShowOneVariant: true,
  allowCustomVariantSplit: false,
  /* ... */
}
```

I'm happy to do this refactor, but it felt like it made sense to do in a follow up PR - what do we think? 

## TODO
- create PROD/CODE s3 files (tests + locks)

## Images
Screenshot 2021-01-20 at 13 45 30

### DIFF
--- a/app/controllers/epic/EpicHoldbackTestsController.scala
+++ b/app/controllers/epic/EpicHoldbackTestsController.scala
@@ -1,0 +1,38 @@
+package controllers.epic
+
+import com.gu.googleauth.AuthAction
+import controllers.LockableS3ObjectController
+import models.EpicTests
+import play.api.libs.circe.Circe
+import play.api.libs.ws.WSClient
+import play.api.mvc._
+import services.FastlyPurger
+import services.S3Client.S3ObjectSettings
+import zio.DefaultRuntime
+
+import scala.concurrent.ExecutionContext
+
+object EpicHoldbackTestsController {
+  val name = "epic-holdback-tests"
+}
+
+class EpicHoldbackTestsController(
+  authAction: AuthAction[AnyContent],
+  components: ControllerComponents,
+  ws: WSClient, stage: String,
+  runtime: DefaultRuntime
+)(implicit ec: ExecutionContext) extends LockableS3ObjectController[EpicTests](
+    authAction,
+    components,
+    stage,
+    name = EpicHoldbackTestsController.name,
+    dataObjectSettings = S3ObjectSettings(
+      bucket = "gu-contributions-public",
+      key = s"epic/$stage/${EpicHoldbackTestsController.name}.json",
+      publicRead = true,  // This data will be requested by dotcom
+      cacheControl = Some("max-age=30"),
+      surrogateControl = Some("max-age=86400")  // Cache for a day, and use cache purging after updates
+    ),
+    fastlyPurger = FastlyPurger.fastlyPurger(stage, s"${EpicHoldbackTestsController.name}.json", ws),
+    runtime = runtime
+  ) with Circe
--- a/app/controllers/epic/EpicHoldbackTestsController.scala
+++ b/app/controllers/epic/EpicHoldbackTestsController.scala
@@ -29,10 +29,10 @@ class EpicHoldbackTestsController(
     dataObjectSettings = S3ObjectSettings(
       bucket = "gu-contributions-public",
       key = s"epic/$stage/${EpicHoldbackTestsController.name}.json",
-      publicRead = true,  // This data will be requested by dotcom
-      cacheControl = Some("max-age=30"),
-      surrogateControl = Some("max-age=86400")  // Cache for a day, and use cache purging after updates
+      publicRead = false,
+      cacheControl = None,
+      surrogateControl = None
     ),
-    fastlyPurger = FastlyPurger.fastlyPurger(stage, s"${EpicHoldbackTestsController.name}.json", ws),
+    fastlyPurger = None,
     runtime = runtime
   ) with Circe
--- a/public/src/components/channelManagement/epicTests/epicTestEditor.tsx
+++ b/public/src/components/channelManagement/epicTests/epicTestEditor.tsx
@@ -302,7 +302,7 @@ const EpicTestEditor: React.FC<EpicTestEditorProps> = ({
         </div>
       )}
 
-      {epicEditorConfig.allowContentTargetting && (
+      {epicEditorConfig.allowContentTargeting && (
         <div className={classes.sectionContainer}>
           <Typography variant={'h3'} className={classes.sectionHeader}>
             Target content
@@ -319,7 +319,7 @@ const EpicTestEditor: React.FC<EpicTestEditorProps> = ({
         </div>
       )}
 
-      {epicEditorConfig.allowLocationTargetting && (
+      {epicEditorConfig.allowLocationTargeting && (
         <div className={classes.sectionContainer}>
           <Typography variant={'h3'} className={classes.sectionHeader}>
             Target audience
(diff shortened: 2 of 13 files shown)


public/src/components/channelManagement/epicTests/epicTestsForm.tsx
```diff
@@ -82,16 +82,27 @@ export interface EpicTest extends Test {
...
+  ...EPICS_THAT_ONLY_ALLOW_A_SINGLE_VARIANT,
+  'AMP',
+];
+
+export const isOnlyAllowedOneVariant = (epicType: EpicType): boolean =>
+  EPICS_THAT_ONLY_ALLOW_A_SINGLE_VARIANT.includes(epicType);
+
+export const isNotAllowedVariantSplit = (epicType: EpicType): boolean =>
+  EPICS_THAT_DONT_ALLOW_VARIANT_SPLIT.includes(epicType);
+
```
TomPretty, line 96: This was a temporary solution to help with the logic. I'd advocate for removing this in the long run in favour of an `EpicConfig` option per epic type

tomrf1: I like your suggestion of the `EpicConfig` 👍

app/controllers/epic/EpicHoldbackTestsController.scala
```diff
@@ -0,0 +1,38 @@
...
+    authAction,
+    components,
+    stage,
+    name = EpicHoldbackTestsController.name,
+    dataObjectSettings = S3ObjectSettings(
+      bucket = "gu-contributions-public",
+      key = s"epic/$stage/${EpicHoldbackTestsController.name}.json",
+      publicRead = true,  // This data will be requested by dotcom
+      cacheControl = Some("max-age=30"),
+      surrogateControl = Some("max-age=86400")  // Cache for a day, and use cache purging after updates
```
tomrf1, line 34: now that we're fetching through the s3 api, we no longer need `publicRead`, `cacheControl`, `surrogateControl`

app/controllers/epic/EpicHoldbackTestsController.scala
```diff
@@ -0,0 +1,38 @@
...
+    ),
+    fastlyPurger = FastlyPurger.fastlyPurger(stage, s"${EpicHoldbackTestsController.name}.json", ws),
```
tomrf1, line 36: `fastlyPurger` is also redundant

app/controllers/epic/EpicHoldbackTestsController.scala
TomPretty: ahhh nice of course! I guess we should do that for all the others too, but perhaps in a follow up?

tomrf1: yep all the public/cache stuff can be removed in a separate PR

TomPretty: I've pushed that up!

public/src/components/channelManagement/epicTests/epicTestsForm.tsx
tomrf1: Would it be much work? It feels like we're half way there with this code!

TomPretty: I've started work on it now - If it ends up being to much code I can always split it out, if not I'll just push it up to here as well

TomPretty: I've pushed up all that refactor! I'll update the PR description tomorrow though!

public/src/components/channelManagement/helpers/shared.ts
```diff
@@ -14,7 +14,112 @@ export interface Test {
...
 }
 
-export type EpicType = 'ARTICLE' | 'LIVEBLOG' | 'APPLE_NEWS' | 'AMP';
+export type EpicType = 'ARTICLE' | 'ARTICLE_HOLDBACK' | 'LIVEBLOG' | 'APPLE_NEWS' | 'AMP';
+
+export interface EpicEditorConfig {
+  // test level settings
+  allowMultipleVariants: boolean;
+  allowCustomVariantSplit: boolean;
+  allowContentTargetting: boolean;
```
tomrf1, line 23: s/Targetting/Targeting

public/src/components/channelManagement/helpers/shared.ts
```diff
@@ -14,7 +14,112 @@ export interface Test {
   variants: Variant[];
 }
 
-export type EpicType = 'ARTICLE' | 'LIVEBLOG' | 'APPLE_NEWS' | 'AMP';
+export type EpicType = 'ARTICLE' | 'ARTICLE_HOLDBACK' | 'LIVEBLOG' | 'APPLE_NEWS' | 'AMP';
+
+export interface EpicEditorConfig {
```
tomrf1, line 19: I had no idea there'd be so many settings, but this is great!

public/src/components/channelManagement/epicTests/epicTestsForm.tsx
```diff
@@ -82,16 +87,24 @@ export interface EpicTest extends Test {
 
 type Props = InnerComponentProps<EpicTest>;
 
-const getEpicTestForm = (epicType: EpicType): React.FC<Props> => {
-  const isLiveBlog = epicType === 'LIVEBLOG';
-  const isOffPlatform = epicType === 'APPLE_NEWS' || epicType === 'AMP';
+const EPICS_THAT_ONLY_ALLOW_A_SINGLE_VARIANT: EpicType[] = ['ARTICLE_HOLDBACK', 'APPLE_NEWS'];
```
tomrf1, line 90: can this stuff be deleted now?

public/src/components/channelManagement/helpers/shared.ts
TomPretty: 🙈 good spot!

public/src/components/channelManagement/helpers/shared.ts
TomPretty: Yeah me neither haha - turns out we have a lot of customization so far!

public/src/components/channelManagement/epicTests/epicTestsForm.tsx
TomPretty: ahhh yeah great spot!

public/src/components/channelManagement/epicTests/epicTestEditor.tsx
```diff
@@ -267,7 +263,7 @@ const EpicTestEditor: React.FC<EpicTestEditorProps> = ({
         </div>
       )}
 
-      {!isOffPlatform && test.variants.length > 0 && (
+      {epicEditorConfig.allowCustomVariantSplit && test.variants.length > 1 && (
```
TomPretty, line 266: I think we might want `> 1` here as that means we have a control + at least one variant

public/src/components/channelManagement/helpers/shared.ts
TomPretty: fixed!

public/src/components/channelManagement/epicTests/epicTestsForm.tsx
TomPretty: deleted!

public/src/components/channelManagement/epicTests/epicTestEditor.tsx
tomrf1: Agreed

tomrf1 (approved):
Great work!

PR #194, opened 2021-02-17
Make epic heading mandatory

tomrf1:


public/src/components/channelManagement/epicTests/utils/defaults.ts
```diff
@@ -11,7 +11,6 @@ export const DEFAULT_MAX_EPIC_VIEWS: MaxEpicViews = {
 
 const DEV_AND_CODE_DEFAULT_VARIANT: EpicVariant = {
   name: 'CONTROL',
-  heading: "Since you're here ...",
```
TomPretty, line 14: It might be worth keeping this?

TomPretty (approved):
LGTM 👍

PR #196, opened 2021-02-24
Make mobile banner copy configurable

tomrf1:
Backend PR: https://github.com/guardian/support-dotcom-components/pull/373

This PR changes the banner test variant model to include:
```
  bannerContent: Option[BannerContent],
  mobileBannerContent: Option[BannerContent],
```

For the migration it will support both the old and new models.

### Screenshots
By default there is no mobile-specific content:
Screen Shot 2021-02-24 at 11 47 38

A new content editor is added below if radio button changes:
Screen Shot 2021-02-24 at 11 47 53

And the label above changes as well:
Screen Shot 2021-02-24 at 11 48 13

### DIFF
--- a/app/models/BannerTests.scala
+++ b/app/models/BannerTests.scala
@@ -17,9 +17,21 @@ object BannerTemplate extends Enum[BannerTemplate] with CirceEnum[BannerTemplate
   case object GuardianWeeklyBanner extends BannerTemplate
 }
 
+case class BannerContent(
+  heading: Option[String],
+  messageText: String,
+  highlightedText: Option[String],
+  cta: Option[Cta],
+  secondaryCta: Option[Cta]
+)
+
 case class BannerVariant(
   name: String,
   template: BannerTemplate,
+  bannerContent: Option[BannerContent],
+  mobileBannerContent: Option[BannerContent],
+
+  // Deprecated - use bannerContent / mobileBannerContent
   heading: Option[String],
   body: String,
   highlightedText: Option[String],
--- a/app/models/BannerTests.scala
+++ b/app/models/BannerTests.scala
@@ -33,7 +33,7 @@ case class BannerVariant(
 
   // Deprecated - use bannerContent / mobileBannerContent
   heading: Option[String],
-  body: String,
+  body: Option[String],
   highlightedText: Option[String],
   cta: Option[Cta],
   secondaryCta: Option[Cta]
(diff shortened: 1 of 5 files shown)


app/models/BannerTests.scala
```diff
@@ -17,11 +17,23 @@ object BannerTemplate extends Enum[BannerTemplate] with CirceEnum[BannerTemplate
...
 
+case class BannerContent(
+  heading: Option[String],
+  messageText: String,
+  highlightedText: Option[String],
+  cta: Option[Cta],
+  secondaryCta: Option[Cta]
+)
+
 case class BannerVariant(
```
tomrf1, line 28: Once migration is complete, `bannerContent` will become mandatory and the deprecated fields removed

app/models/BannerTests.scala
```diff
@@ -17,11 +17,23 @@ object BannerTemplate extends Enum[BannerTemplate] with CirceEnum[BannerTemplate
   case object GuardianWeeklyBanner extends BannerTemplate
 }
 
+case class BannerContent(
```
tomrf1, line 20: This model matches the one in dotcom-components

public/src/components/channelManagement/bannerTests/bannerTestVariantEditor.tsx
```diff
@@ -88,96 +109,202 @@ const BannerTestVariantEditor: React.FC<BannerTestVariantEditorProps> = ({
...
+            onValidationChange={onValidationChange}
+            supportSecondaryCta={true}
+          />
+        </div>
+      </div>
+    </>
+  );
+};
+
+const BannerTestVariantContentEditorWithStyles = withStyles(styles)(BannerTestVariantContentEditor);
```
TomPretty, line 205: we could potentially refactor to use the `useStyles` way of styling instead of the `withStyles` to avoid having to do this (it's also generally a little nicer) e.g: https://github.com/guardian/support-admin-console/blob/fe9cf64d613a9d913a6e1b8f21a4d8fad285dc61/public/src/components/shared/liveSwitch.tsx#L4

tomrf1: oh yes that is much nicer, thanks!

TomPretty (approved):
Just a tiny suggestion about potentially using `useStyles`, but nonetheless looks great to me! 🚀

PR #197, opened 2021-02-25
Use t4g instance type

tomrf1:


TomPretty (commented):
does this need to change in the riffraff yaml too?

tomrf1:
> does this need to change in the riffraff yaml too?

yes - can you not see it in the diff?!

TomPretty (approved):
oh now I can - idk if I was just being blind before 😂 looks good! 👍

PR #198, opened 2021-02-26
Update node + jest

TomPretty:
## What does this change
Just some small updates. Updated the version of node to the latest LTS then noticed npm was complaining about jest having a security vulnerability so updated that too.


paulbrown1982 (approved):
👍

PR #201, opened 2021-04-12
Add separate article count settings to variants

TomPretty:
## What does this change?
Adds a variant level config to turn on/off the separate article count feature. For now it is effectively either on/off, but we've given ourselves some leeway in the model to add new fields/allow new types of article count. 

The corresponding SDC PR is: https://github.com/guardian/support-dotcom-components/pull/417. 

## Images
Screenshot 2021-04-12 at 15 06 13

### DIFF
--- a/app/models/EpicTests.scala
+++ b/app/models/EpicTests.scala
@@ -39,6 +39,18 @@ case class TickerSettings(
   copy: TickerCopy
 )
 
+sealed trait SeparateArticleCountType extends EnumEntry
+object SeparateArticleCountType extends Enum[SeparateArticleCountType] with CirceEnum[SeparateArticleCountType] {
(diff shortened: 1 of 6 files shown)


app/models/EpicTests.scala
```diff
@@ -39,6 +39,18 @@ case class TickerSettings(
   copy: TickerCopy
 )
 
+sealed trait SeparateArticleCountType extends EnumEntry
```
tomrf1, line 42: This pattern works and is how we've been doing enums elsewhere, so no need to change it now - but I just discovered there is now a better way:
https://stackoverflow.com/a/66967469

tomrf1 (approved):
great!

PR #202, opened 2021-04-15
Update the epic secondary cta options

TomPretty:
## What does this change?
Update the epic secondary cta options - we now have the option to select either: `None`, `Custom` (which opens the CTA fields editor), or `Contributions reminder`. The previous behaviour was that when the secondary button was unchecked, it would always show a contributions reminder.

Before the banner and epic editors shared the same `VariantEditorButtonsEditor` component but now they require different components so I've separated that into two components: `BannerTestVariantEditorCtasEditor` and `EpicTestEditorCtasEditor`. Additionally I've renamed any mention of `Button(s)` to `Cta(s)` as that lines up better with the types.

The corresponding SDC change is here: https://github.com/guardian/support-dotcom-components/pull/424

**TODO**

Having looked through the PROD data, there aren't any tests with a custom secondary CTA field, so it won't break anything deploying to PROD. However, I imagine essentially all of the existing tests will want to be updated to use the `Contributions reminder` option.

- [x] test in CODE

** AFTER DEPLOY **

I will go through the tests in the PROD tool with Jesse to update the tests. There are currently no tests in PROD that use the `secondaryCta` field (i.e. it's always undefined). This means, after we deploy all tests will show as `None`. We probably want most (all?) to actually be set to `Contributions reminder`.  

## Images

Screenshot 2021-04-15 at 13 27 00

Screenshot 2021-04-15 at 13 27 04

Screenshot 2021-04-15 at 13 27 08


public/src/components/channelManagement/variantEditorCtaFieldsEditor.tsx
```diff
@@ -0,0 +1,91 @@
+import React, { useEffect } from 'react';
```
TomPretty, line 1: This was factored out of the `VariantEditorCtaEditor` so that it can be shared by the new `VariantEditorSecondaryCtaEditor`

public/src/components/channelManagement/variantEditorSecondaryCtaEditor.tsx
```diff
@@ -0,0 +1,90 @@
...
+const VariantEditorSecondaryCtaEditor: React.FC<VariantEditorSecondaryCtaEditorProps> = ({
+  classes,
+  label,
+  cta,
+  updateCta,
+  onValidationChange,
+  isDisabled,
+}: VariantEditorSecondaryCtaEditorProps) => {
+  const handleChange = (event: React.ChangeEvent<{ value: unknown }>): void => {
+    const value = event.target.value as SecondaryCtaKind | 'None';
```
TomPretty, line 48: If instead of making `secondaryCta` and optional field on the `Variant` type, we added that as a new `SecondaryCtaKind` i.e.

```ts
export enum SecondaryCtaKind {
  None = 'NoneSecondaryCta',
  Custom = 'CustomSecondaryCta',
  ContributionsReminder = 'ContributionsReminderSecondaryCta',
}
```
it would make the logic generally a little cleaner

public/src/components/channelManagement/helpers/shared.ts
```diff
@@ -89,7 +89,7 @@ export const LIVEBLOG_EPIC_CONFIG: EpicEditorConfig = {
   allowVariantImageUrl: false,
   allowVariantFooter: false,
   allowVariantCustomPrimaryCta: true,
-  allowVariantCustomSecondaryCta: true,
+  allowVariantCustomSecondaryCta: false,
```
TomPretty, line 92: I noticed we don't use this in SDC so I turned this off

app/models/EpicTests.scala
```diff
@@ -10,6 +10,24 @@ case class ControlProportionSettings(proportion: Float, offset: Int)
...
+  `type`: String = "CustomSecondaryCta",
+  cta: Cta,
+) extends SecondaryCta
+
+case class ContributionsReminderSecondaryCta(
+  `type`: String = "ContributionsReminderSecondaryCta",
+) extends SecondaryCta
+
+object SecondaryCta {
+  implicit val customConfig: Configuration = Configuration.default.withDiscriminator("type")
```
tomrf1, line 25: 💯

public/src/components/channelManagement/bannerTests/bannerTestVariantEditorCtasEditor.tsx
```diff
@@ -33,7 +33,7 @@ interface VariantEditorButtonsEditorProps extends WithStyles<typeof styles> {
   supportSecondaryCta: boolean;
 }
 
-const VariantEditorButtonsEditor: React.FC<VariantEditorButtonsEditorProps> = ({
+const BannerTestVariantEditorCtasEditor: React.FC<BannerTestVariantEditorCtasEditorProps> = ({
```
tomrf1, line 36: just a name, but I wonder if this could just be `BannerCtasEditor`

tomrf1 (approved):
great!

public/src/components/channelManagement/bannerTests/bannerTestVariantEditorCtasEditor.tsx
TomPretty: yeah the naming convention is a little clunky here. I think it might be best to do a follow up PR to clean that up a little (being thorough will touch quite a few files!)

tomrf1: 👍

PR #203, opened 2021-04-20
Make cta fields required

TomPretty:
## What does this change?
Make the `Cta` fields required to match with dotcom-components.

### DIFF
--- a/public/src/components/channelManagement/epicTests/epicTestVariantEditorCtasEditor.tsx
+++ b/public/src/components/channelManagement/epicTests/epicTestVariantEditorCtasEditor.tsx
@@ -19,6 +19,11 @@ const DEFAULT_PRIMARY_CTA: Cta = {
   baseUrl: 'https://support.theguardian.com/contribute',
 };
 
+const DEFAULT_SECONDARY_CTA: Cta = {
+  text: '',
+  baseUrl: '',
+};
+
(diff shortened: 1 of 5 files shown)


public/src/components/channelManagement/epicTests/epicTestVariantEditorCtasEditor.tsx
```diff
@@ -19,6 +19,11 @@ const DEFAULT_PRIMARY_CTA: Cta = {
   baseUrl: 'https://support.theguardian.com/contribute',
 };
 
+const DEFAULT_SECONDARY_CTA: Cta = {
```
tomrf1, line 22: what happens if we don't have default values?

TomPretty: Yeah we could leave them empty, then when a user selects `Custom` the fields would be empty by default, maybe that makes more sense?

tomrf1: I think empty is preferable for 2nd cta as it could be anything! It should still error when they try to save

TomPretty: Yeah that makes sense - I've made that change now!

tomrf1 (approved):
👍

PR #207, opened 2021-04-27
Banner editor padding

TomPretty:
## What does this change?
I noticed we'd lost the padding on the banner variant editor - this adds it back!

### before

Screenshot 2021-04-27 at 09 03 58

### after

Screenshot 2021-04-27 at 09 03 29


tomrf1 (approved):
good spot

PR #209, opened 2021-05-04
Channel switches

tomrf1:
A new page for switches relating to epics/banners.

This is primarily for emergencies when we need to disable the channels on dotcom

Screen Shot 2021-05-04 at 14 10 39
Screen Shot 2021-05-04 at 14 10 49

### DIFF
--- a/public/src/components/channelManagement/ChannelSwitches.tsx
+++ b/public/src/components/channelManagement/ChannelSwitches.tsx
@@ -1,0 +1,109 @@
+import React, {useEffect, useState} from 'react';
+import {makeStyles, Theme} from "@material-ui/core";
+import {fetchFrontendSettings, FrontendSettingsType, saveFrontendSettings} from "../../utils/requests";
+import Switch from "@material-ui/core/Switch";
+import FormControlLabel from "@material-ui/core/FormControlLabel";
+import Button from "@material-ui/core/Button";
+
+const useStyles = makeStyles(({ palette }: Theme) => ({
+  container: {
+    margin: '30px',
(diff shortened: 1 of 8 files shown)


public/src/components/channelManagement/ChannelSwitches.tsx
```diff
@@ -0,0 +1,130 @@
...
+        value={name}
+      />
+    }
+  />
+);
+
+const ChannelSwitches: React.FC = () => {
+  const classes = useStyles();
+  const [switches, setSwitches] = useState<ChannelSwitches>(defaultSwitches);
+  const [version, setVersion] = useState<string | null>(null);
```
TomPretty, line 66: Looks great! This might not actually save too much code, but we do have a [withS3Data hoc](https://github.com/guardian/support-admin-console/blob/master/public/src/hocs/withS3Data.tsx) which I think would fit here. We're currently using it for the [amounts editor](https://github.com/guardian/support-admin-console/blob/master/public/src/components/amounts/configuredAmountsEditor.tsx#L116), but it was designed (🤞) to be generic!

tomrf1: ahh I forgot about this!

tomrf1: I've pushed that up - much nicer

TomPretty: oh nice - yeah that worked out great! 🚀

TomPretty (approved):
Looks great! 🔥 🔥

PR #210, opened 2021-05-05
remove bintray dependency

tomrf1:
simple-configuration has moved to maven, as bintray is being binned

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -5,16 +5,13 @@ version := "1.0-SNAPSHOT"
 
 scalaVersion := "2.12.8"
 
-//Bintray for simple-configuration dependency
-resolvers += "Guardian Platform Bintray" at "https://dl.bintray.com/guardian/platforms"
-
 val circeVersion = "0.10.0"
 val jacksonVersion = "2.10.0"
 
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -5,16 +5,13 @@ version := "1.0-SNAPSHOT"
...
-resolvers += "Guardian Platform Bintray" at "https://dl.bintray.com/guardian/platforms"
-
 val circeVersion = "0.10.0"
 val jacksonVersion = "2.10.0"
 
 libraryDependencies ++= Seq(
   "com.typesafe.scala-logging" %% "scala-logging" % "3.7.2",
   "com.gu.play-googleauth" %% "play-v26" % "1.0.7",
-  "com.gu" %% "simple-configuration-ssm" % "1.4.1",
+  "com.gu" %% "simple-configuration-ssm" % "1.5.6",
```
johnduffell, line 14: you don't need to update to the v2 credential provider? or it just uses a default?

johnduffell (approved):
👍 just one question/comment

build.sbt
tomrf1: no, that happens internally

PR #211, opened 2021-05-06
Mj add banner ab audience percentages

michaelbjacobson:
### What does this PR do?
Implements control/variant audience proportion assignment, as we already do on contributions `epics`.

TODO: There is definitely scope for further refactoring between `epic` and `banner` test config to reduce code repetition. I suggest we make this a health task to pick up in the future.

### DIFF
--- a/app/utils/ControlProportionSettings.scala
+++ b/app/utils/ControlProportionSettings.scala
@@ -1,0 +1,3 @@
+package utils
+
+case class ControlProportionSettings(proportion: Float, offset: Int)
--- a/public/src/components/channelManagement/bannerTests/bannerTestEditor.tsx
+++ b/public/src/components/channelManagement/bannerTests/bannerTestEditor.tsx
@@ -20,7 +20,7 @@ import TestEditorVariantSummary from '../testEditorVariantSummary';
 import BannerVariantPreview from './bannerVariantPreview';
 import { ControlProportionSettings } from '../helpers/controlProportionSettings';
 import TestVariantsSplitEditor from '../testVariantsSplitEditor';
-import { testEditorStyles as styles } from '../helpers/testEditorStyles';
+import { useStyles } from '../helpers/testEditorStyles';
 
 const copyHasTemplate = (content: BannerContent, template: string): boolean =>
   (content.heading && content.heading.includes(template)) || content.messageText.includes(template);
@@ -31,7 +31,7 @@ const testCopyHasTemplate = (test: BannerTest, template: string): boolean =>
       (variant.mobileBannerContent && copyHasTemplate(variant.mobileBannerContent, template)),
   );
 
-interface BannerTestEditorProps extends WithStyles<typeof styles> {
+interface BannerTestEditorProps {
   test: BannerTest;
   hasChanged: boolean;
   onChange: (updatedTest: BannerTest) => void;
@@ -47,7 +47,6 @@ interface BannerTestEditorProps extends WithStyles<typeof styles> {
 }
 
 const BannerTestEditor: React.FC<BannerTestEditorProps> = ({
-  classes,
   test,
   onChange,
   onValidationChange,
@@ -60,6 +59,8 @@ const BannerTestEditor: React.FC<BannerTestEditorProps> = ({
   testNicknames,
   createTest,
 }: BannerTestEditorProps) => {
+  const classes = useStyles();
+
   const setValidationStatusForField = useValidation(onValidationChange);
 
   const getArticlesViewedSettings = (test: BannerTest): ArticlesViewedSettings | undefined => {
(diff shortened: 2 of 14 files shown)


app/utils/ControlProportionSettings.scala
```diff
@@ -0,0 +1,3 @@
+package utils
```
tomrf1, line 1: 👍 for a new file, but this can live in the `models` package

public/src/components/channelManagement/bannerTests/bannerTestEditor.tsx
```diff
@@ -114,20 +110,39 @@ const BannerTestEditor: React.FC<BannerTestEditorProps> = ({
...
+  const updateTest = (update: (test: BannerTest) => BannerTest): void => {
+    if (test) {
+      const updatedTest = update(test);
+
+      onChange({
+        ...updatedTest,
+        // To save dotcom from having to work this out
+        articlesViewedSettings: getArticlesViewedSettings(updatedTest),
+      });
+    }
```
TomPretty, line 122: How come we're making this change?

michaelbjacobson: Basically to better mirror the implementation on epics. My thinking is that this will help pave the way to share as much code as possible between the two.

app/utils/ControlProportionSettings.scala
michaelbjacobson: Done! ✅

public/src/components/channelManagement/bannerTests/bannerTestEditor.tsx
TomPretty: Yeah that makes sense - I think it might be best to refactor the other way around. The epic manager still has some older logic from the old style UX of the tool e.g the `if (test)` isn't really needed as the type of `test` is `BannerTest` so it can't be `null/undefined`.

public/src/components/channelManagement/bannerTests/bannerTestEditor.tsx
```diff
@@ -114,20 +72,35 @@ const BannerTestEditor: React.FC<BannerTestEditorProps> = ({
     return undefined;
   };
 
-  const updateTest = (updatedTest: BannerTest): void => {
+  const updateTest = (update: (test: BannerTest) => BannerTest): void => {
+    const updatedTest = update(test);
+
```
TomPretty, line 77: Yeah nice I think it's good to remove the `if` here as it wasn't needed. I think it's maybe slightly unnecessary here for the argument to be a function that updates the test - it might be simpler to just pass in the updated test! For some context, I think the reason it was like this in the `epicTestEditor` was because `test` could potentially be `undefined` so having `updateTest` take an `update` function removed the repetition of having to check `if (test)` in all the `onChange` handlers (`onVariantsChange`, `onRegionsChange`, etc). But now `test` is always defined I think it actually adds a little extra code instead of saving some!

TomPretty: Oh! I've just got a little farther and I can see that the `epicTestEditor` `onChange` handlers *did* check it `if(test)` anyways! I really don't know why we have this `update` function then! It's safe to remove though!

public/src/components/channelManagement/epicTests/epicTestEditor.tsx
```diff
@@ -18,46 +18,11 @@ import EpicTestTargetContentEditor from './epicTestTargetContentEditor';
...
-  sectionHeader: {
-    fontSize: 16,
-    fontWeight: 500,
-    color: palette.grey[700],
-  },
-  buttonsContainer: {
-    paddingTop: spacing(4),
-    paddingBottom: spacing(12),
-  },
-}));
```
TomPretty, line 60: I think it's great to share styles! But the `useStyles` approach is actually the newer, preferred method. Really we should refactor all the components to use `useStyles` over the `withStyles` approach but we haven't gotten around to it! I'd say we'd be better off updating the `bannerTestEditor` instead of this one.

public/src/components/channelManagement/helpers/testEditorStyles.ts
```diff
@@ -0,0 +1,40 @@
+import { createStyles, Theme } from '@material-ui/core';
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+export const testEditorStyles = ({ spacing, palette }: Theme) =>
+  createStyles({
```
TomPretty, line 5: Yeah again I think it's great to share styles! But probably best to go for the `useStyles` method. We could rename this file to something like `useTestEditorStyles` and have it export that as a hook

TomPretty (approved):
Yeah great stuff! 🔥

PR #212, opened 2021-05-07
Fix highlighted text

TomPretty:
## What does this change?
Make highlighted text field show for contributions banner. We had our parenthesis slightly in the wrong place which meant the field was hidden when the contributions banner template was selected. This change fixes the logic.


paulbrown1982 (approved):
👍

PR #213, opened 2021-05-14
Add reminders to banner

TomPretty:
## What does this change?
Update the banner model + UI to support having the option to have a contributions reminder as the secondary cta.

## Migration plan

Merging this will break the tool in PROD. We will have to manually update the the data in s3 for any test that **has a secondary cta**. For tests **without** a secondary cta, the field is **undefined** so will be fine after the merge.

## Images
Screenshot 2021-05-14 at 13 17 08


tomrf1 (approved):
🚢

PR #214, opened 2021-05-24
Update template validation + cleanup

TomPretty:
## What does this change?
Update template validation. I made two updates to the template validation regex. 

1. we now match on `\S` = any none whitespace character i.e `%%HELL*O%%` will now be detected as an invalid template
2. we now match on just one `%` either side i.e `%ARTICLE_COUNT%`, `%%ARTICLE_COUNT%%`, `%%ARTICLE_COUNT%` will all be detected as invalid templates

Additionally I removed some duplicated code and removed some old tests.


build-tc
```diff
@@ -11,6 +11,5 @@ nvm use
 
 npm install
 npm run lint
-npm run test
```
TomPretty, line 14: We now have no tests, which causes this line to fail the build

public/src/components/channelManagement/epicTests/__tests__/epicTestVariantEditor.spec.tsx
```diff
@@ -1,19 +0,0 @@
...
-      expect(getInvalidTemplateError('I am from %%COUNTRY_NAME%%')).toEqual(null);
-    });
-
-    it('should return an error message if template is not ok', () => {
-      expect(getInvalidTemplateError('I am from %%COUNTRY_NAM%%')).toEqual(
-        'Invalid template: %%COUNTRY_NAM%%',
-      );
-    });
-  });
-});
```
TomPretty, line 19: We could move these into a new file e.g `validation.spec.tsx` and test the function in `validation.tsx`?

tomrf1: yes - I think test cases are still useful here

TomPretty: Yeah makes sense - I'll add them back. I can add a few new cases as `if template is not ok` could mean a few things

tomrf1 (approved):
👍 great

public/src/components/channelManagement/epicTests/__tests__/epicTestVariantEditor.spec.tsx
TomPretty: I've pushed that up!

PR #216, opened 2021-05-27
Upgrade to modules v2

TomPretty:
## What does this change?
Upgrade to use the v2 version of the modules. This also required upgrading to emotion 11 + react 17.

### DIFF
--- a/public/src/components/channelManagement/bannerTests/bannerVariantPreview.tsx
+++ b/public/src/components/channelManagement/bannerTests/bannerVariantPreview.tsx
@@ -1,8 +1,7 @@
 import React, { useEffect, useState } from 'react';
 import { Theme, makeStyles, Button } from '@material-ui/core';
-import * as emotion from 'emotion';
-import * as emotionCore from '@emotion/core';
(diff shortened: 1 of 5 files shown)


public/src/components/channelManagement/bannerTests/bannerVariantPreview.tsx
```diff
@@ -120,17 +119,16 @@ const BannerVariantPreview: React.FC<BannerVariantPreviewProps> = ({
...
+      emotionReactJsxRuntime,
     };
 
     const stage = getStage();
 
     const baseUrl =
       stage === 'PROD'
-        ? 'https://contributions.guardianapis.com/modules/v1/banners'
-        : 'https://contributions.code.dev-guardianapis.com/modules/v1/banners';
+        ? 'https://contributions.guardianapis.com/modules/v2/banners'
```
tomrf1, line 130: doesn't have to be now, but it would be nice to move the version into a shared const somewhere

tomrf1 (approved):
thanks!

PR #218, opened 2021-06-02
Update banner body copy limit when there is a secondary cta

TomPretty:
## What does this change?
Update banner body copy limit when there is a secondary cta


tomrf1 (approved):
👍

PR #225, opened 2021-06-30
Show AC above in live preview

TomPretty:
## What does this change?
Show AC above in live preview

Screenshot 2021-06-30 at 11 16 44


tomrf1 (approved):
thanks!

PR #229, opened 2021-07-23
Super mode switch

tomrf1:
This will be used by SDC
Screen Shot 2021-07-23 at 12 46 21


.sbtopts
```diff
@@ -0,0 +1,2 @@
+-J-Xss8m
```
tomrf1, line 1: I was getting stack overflows during compilation

paulbrown1982 (approved):
👍

PR #230, opened 2021-08-31
Remove aus moment code

TomPretty:
## What does this change?
Remove aus moment code

## TODO

- [x] wait for tests to be removed from the tool


tomrf1 (approved):
perhaps check with BRR that they're happy for them to be archived

TomPretty:
> perhaps check with BRR that they're happy for them to be archived

Yeah! Just got confirmation that Paddy is happy, so I've gone ahead and archived all those tests

PR #231, opened 2021-09-01
Add region selector for apple news epics

TomPretty:
## What does this change?

Add a region selector for apple news epics. We're launching apple news in Aus so we want to be able to target epics specific to the US or Australia. 

## Images

Screenshot 2021-09-01 at 11 55 33

### DIFF
--- a/public/src/components/channelManagement/helpers/shared.ts
+++ b/public/src/components/channelManagement/helpers/shared.ts
@@ -1,3 +1,4 @@
+import { Region } from '../../../utils/models';
 import { ValidationStatus } from './validation';
 
 export interface Variant {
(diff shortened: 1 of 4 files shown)


public/src/components/channelManagement/helpers/shared.ts
```diff
@@ -99,8 +101,9 @@ export const APPLE_NEWS_EPIC_CONFIG: EpicEditorConfig = {
   allowMultipleVariants: false,
   allowCustomVariantSplit: false,
   allowContentTargeting: true,
-  allowLocationTargeting: false,
-  allowSupporterStatusTargeting: true,
+  allowLocationTargeting: true,
+  customRegions: [Region.UnitedStates, Region.AUDCountries],
+  allowSupporterStatusTargeting: false,
```
TomPretty, line 106: This should have always been `false`, but the supporter status targeting was actually never shown if `allowLocationTargeting` was `false`

tomrf1 (approved):
great!

PR #235, opened 2021-09-29
Fix control proportion bug

TomPretty:
## What does this change?
Fix control proportion bug where a test with 0 or 1 variant(s) can have `controlProportionSettings` set. This caused a bug in support-dotcom-components that led to 500s. We've fixed the SDC bug in this [PR](https://github.com/guardian/support-dotcom-components/pull/532), but this PR ensures that we reset the `controlProportionSettings` settings when dropping down to just a single variant.

### DIFF
--- a/public/src/components/channelManagement/epicTests/epicTestEditor.tsx
+++ b/public/src/components/channelManagement/epicTests/epicTestEditor.tsx
@@ -21,7 +21,10 @@ import { ARTICLE_COUNT_TEMPLATE, COUNTRY_NAME_TEMPLATE } from '../helpers/valida
 import TestVariantsSplitEditor from '../testVariantsSplitEditor';
 import { getDefaultVariant } from './utils/defaults';
 import LiveSwitch from '../../shared/liveSwitch';
-import { ControlProportionSettings } from '../helpers/controlProportionSettings';
+import {
+  canHaveCustomVariantSplit,
+  ControlProportionSettings,
+} from '../helpers/controlProportionSettings';
 import { useStyles } from '../helpers/testEditorStyles';
 
 const copyHasTemplate = (test: EpicTest, template: string): boolean =>
@@ -107,7 +110,11 @@ const EpicTestEditor: React.FC<EpicTestEditorProps> = ({
 
   const onVariantDelete = (deletedVariantName: string): void => {
     const updatedVariantList = test.variants.filter(variant => variant.name !== deletedVariantName);
-    onVariantsChange(updatedVariantList);
+    const controlProportionSettings = canHaveCustomVariantSplit(updatedVariantList)
+      ? test.controlProportionSettings
+      : undefined;
+
+    updateTest({ ...test, variants: updatedVariantList, controlProportionSettings });
   };
 
   const createVariant = (name: string): void => {
@@ -233,7 +240,7 @@ const EpicTestEditor: React.FC<EpicTestEditorProps> = ({
         </div>
       )}
 
-      {epicEditorConfig.allowCustomVariantSplit && test.variants.length > 1 && (
+      {epicEditorConfig.allowCustomVariantSplit && canHaveCustomVariantSplit(test.variants) && (
         <div className={classes.sectionContainer}>
           <Typography variant={'h3'} className={classes.sectionHeader}>
             Variants split
--- a/public/src/components/channelManagement/testVariantsSplitEditor.tsx
+++ b/public/src/components/channelManagement/testVariantsSplitEditor.tsx
@@ -11,12 +11,9 @@ import {
 import { Variant } from './helpers/shared';
 import { EMPTY_ERROR_HELPER_TEXT } from './helpers/validation';
 import { useForm } from 'react-hook-form';
-import { ControlProportionSettings } from './helpers/controlProportionSettings';
+import { ControlProportionSettings, hasControl } from './helpers/controlProportionSettings';
 
 const MAX_MVT = 1_000_000;
-
-const hasControl = (variants: Variant[]): boolean =>
-  !!variants.find(v => v.name.toLowerCase() === 'control');
 
 const randomMvt = (): number => Math.round(Math.random() * MAX_MVT);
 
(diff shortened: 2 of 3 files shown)


public/src/components/channelManagement/epicTests/epicTestEditor.tsx
```diff
@@ -107,7 +107,10 @@ const EpicTestEditor: React.FC<EpicTestEditorProps> = ({
 
   const onVariantDelete = (deletedVariantName: string): void => {
     const updatedVariantList = test.variants.filter(variant => variant.name !== deletedVariantName);
-    onVariantsChange(updatedVariantList);
+    const controlProportionSettings =
+      updatedVariantList.length >= 2 ? test.controlProportionSettings : undefined;
+
+    updateTest({ ...test, variants: updatedVariantList, controlProportionSettings });
```
TomPretty, line 113: I considered updating `onVariantsChange` and adding the `controlProportionSettings` logic there. However, it's only on deleting variants that we need to do this, so it seemed to make sense to just update here (but does mean we need to use `updateTest` here instead of `onVariantsChange`)

tomrf1 (approved):
great!

public/src/components/channelManagement/epicTests/epicTestEditor.tsx
```diff
@@ -107,7 +107,10 @@ const EpicTestEditor: React.FC<EpicTestEditorProps> = ({
 
   const onVariantDelete = (deletedVariantName: string): void => {
     const updatedVariantList = test.variants.filter(variant => variant.name !== deletedVariantName);
-    onVariantsChange(updatedVariantList);
+    const controlProportionSettings =
+      updatedVariantList.length >= 2 ? test.controlProportionSettings : undefined;
```
johnduffell, line 111: can we be a bit more explicit with something like this (pseudo typescript)
```suggestion
      updatedVariantList.exists(!_.equalsIgnoreCase("control")) ? test.controlProportionSettings : undefined;
```

johnduffell (approved):
great work. 👍 just a comment for consistency with SDC. I guess there are no unit tests for this area to update?

public/src/components/channelManagement/epicTests/epicTestEditor.tsx
TomPretty: Yeahh this is a great point - actually we should probably check that there is also a variant left called `control`

TomPretty: Perhaps (going all out)

```ts
  function hasControl(variants: EpicVariant[]): boolean {
    return variants.findIndex(v => v.name.toLowerCase() === 'control') > -1;
  }

  function hasAtLeastOneNonControlVariant(variants: EpicVariant[]): boolean {
    return variants.filter(v => v.name.toLowerCase() !== 'control').length >= 1;
  }

  const onVariantDelete = (deletedVariantName: string): void => {
    const updatedVariantList = test.variants.filter(variant => variant.name !== deletedVariantName);
    const controlProportionSettings =
      hasControl(updatedVariantList) && hasAtLeastOneNonControlVariant(updatedVariantList)
        ? test.controlProportionSettings
        : undefined;

    updateTest({ ...test, variants: updatedVariantList, controlProportionSettings });
  };
```

Or potentially that check could just be 

```ts
      hasControl(updatedVariantList) && updatedVariantList.length >= 2
```

which should also work

TomPretty:
> great work. 👍 just a comment for consistency with SDC. I guess there are no unit tests for this area to update?

No unit tests I'm afraid :( We've spoken about adding tests in the past - this would be great bit of logic to have tested. I think we decided we didn't need to go as far as having component tests (e.g with `React Testing Library`). But pulling logic in the components out into there own pure `.ts` files and then having those unit tested would be great

public/src/components/channelManagement/epicTests/epicTestEditor.tsx
```diff
@@ -233,22 +238,24 @@ const EpicTestEditor: React.FC<EpicTestEditorProps> = ({
...
-            <TestVariantsSplitEditor
-              variants={test.variants}
-              controlProportionSettings={test.controlProportionSettings}
-              onControlProportionSettingsChange={onControlProportionSettingsChange}
-              onValidationChange={onVariantsSplitSettingsValidationChanged}
-              isDisabled={!editMode}
-            />
+      {epicEditorConfig.allowCustomVariantSplit &&
+        hasControl(test.variants) &&
+        test.variants.length >= 2 && (
```
TomPretty, line 243: Just updated this logic as we shouldn't show the variant split editor if there isn't a control

public/src/components/channelManagement/testVariantsSplitEditor.tsx
```diff
@@ -169,7 +166,7 @@ const TestVariantsSplitEditor: React.FC<TestVariantsSplitEditorProps> = ({
           />
         </RadioGroup>
 
-        {controlProportionSettings && hasControl(variants) && (
+        {controlProportionSettings && (
```
TomPretty, line 169: I've moved the `hasControl` check outside this component so we don't render the whole component if we don't have a control

public/src/components/channelManagement/epicTests/epicTestEditor.tsx
johnduffell: this logic is repeated on line 111 and 243, should we have a `hasNonControl` function as well?

johnduffell (approved):
looks good, just a suggestion to make the logic a bit more transparent

public/src/components/channelManagement/epicTests/epicTestEditor.tsx
johnduffell: sorry didn't spot this as it was "outdated" but yeah I duplicated your suggestion before I saw this :)  So go for it!

johnduffell: I think the naming of the functions here is the really useful part, as it enables the intention/why of the code to be clear, rather than just the mechanics.

PR #240, opened 2021-10-11
Add the contributions banner with sign in as a selectable template

i-hardy:
## What does this change?
See guardian/support-dotcom-components#539

This adds the 'contributions banner with sign-in link' as a selectable template, to facilitate A/B testing this option against the existing yellow contributions banner which does not have a sign-in link.

This is part of the set up work to allow colleagues in BRR to A/B test this version with a sign-in link against the existing template. It is anticipated that after this test this template option will be removed, either due to incorporating the link into the banner by default, or because the test demonstrated a significant negative impact on key metrics.

## Images

**Template selection options**
Screenshot 2021-10-11 at 16 04 06

See linked PR in sdc for images of the banner itself


TomPretty (approved):
LGTM 👍

PR #242, opened 2021-10-15
Add option to show a sign in link on an epic variant

i-hardy:
## What does this change?
This adds the ability to show a sign-in link on an epic variant, to facilitate A/B testing the effect of showing readers the option to sign in. See guardian/support-dotcom-components#543 for images, and #240 for a similar PR relating to testing the same option on the banner.

### DIFF
--- a/public/src/components/channelManagement/helpers/shared.ts
+++ b/public/src/components/channelManagement/helpers/shared.ts
@@ -106,7 +106,7 @@ export const LIVEBLOG_EPIC_CONFIG: EpicEditorConfig = {
   allowVariantSeparateArticleCount: false,
   allowVariantTicker: false,
   allowVariantChoiceCards: false,
-  allowVariantSignInLink: true,
+  allowVariantSignInLink: false,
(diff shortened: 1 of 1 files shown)


public/src/components/channelManagement/helpers/shared.ts
```diff
@@ -103,6 +106,7 @@ export const LIVEBLOG_EPIC_CONFIG: EpicEditorConfig = {
   allowVariantSeparateArticleCount: false,
   allowVariantTicker: false,
   allowVariantChoiceCards: false,
+  allowVariantSignInLink: true,
```
TomPretty, line 109: It's probably worth setting the liveblog, apple news, and amp epic configs to `false` here. Those are all different components so we'll have to add the sign in link functionality separately there too. Whilst we don't have it, it's probs best to hide the option from users

i-hardy: 👍 Makes sense!

PR #243, opened 2021-10-18
add a filter-by-region selector at top of tasks list, for read-only mode

KaliedaRik:
## What does this change?
Adds a simple filter-by-region selector at the top of the tests list on the left hand side of the page. Selector only appears when in READ-ONLY mode. Replaced by static text whenever user moves into EDIT mode.

## How to test
Create a set of tests which are restricted by region. 
+ Check that when a region is selected, only tests which include that region are listed beneath the selector. 
+ Check that when the 'Show all tests' option is selected, all tests display beneath the selector
+ Check that the selector is replaced by static text, and all tests display beneath the text, when moving into EDIT mode
+ Check that the selector displays the 'Show all tests' option, and all tests display beneath it, when user navigates to a different page

A small cosmetic change which will help users more quickly find the tests they want to edit

Risks are minimal, no data is changed by this cosmetic update

### DIFF
--- a/public/src/components/channelManagement/bannerTests/bannerTestsForm.tsx
+++ b/public/src/components/channelManagement/bannerTests/bannerTestsForm.tsx
@@ -34,6 +34,8 @@ const getBannerTestsForm = (isFirstChannel: boolean): React.FC<Props> => {
     requestLock,
     cancel,
     editMode,
+    regionFilter,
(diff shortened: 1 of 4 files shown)


tomrf1 (approved):
Great work! Just had another play in CODE and it's a really useful feature

PR #248, opened 2021-11-04
Change default primary CTA

samanthagottlieb:
Co-authored-by: Tom Pretty <TomPretty@users.noreply.github.com>
Co-authored-by: Maria Olanipekun <marialani@users.noreply.github.com>
Co-authored-by: Michael Jacobson <michaelbjacobson@users.noreply.github.com>

## What does this change?

This PR changes the default primary epic CTA to 'Continue'.

We have added default primary and secondary CTA objects for both the banner and epic to defaults.ts so these can be reused and to avoid dupication.

We have removed an unused object from shared.ts.

## Images

Screenshot 2021-11-04 at 15 51 14


tomrf1 (approved):
great!

PR #249, opened 2021-11-09
Capi proxy endpoints

tomrf1:
Adds endpoints for querying capi for tags and sections.
Simply proxies the client requests, adding an api-key.

The client will need to extract the array of results, e.g.
```
fetch(`/capi/tags?q=${term}`).then( response => response.json()).then(body => body.response.results)
```


KaliedaRik:
Looks good to me. Thanks for showing me how to build the new endpoint in the backend

KaliedaRik (approved):

PR #254, opened 2021-11-22
add enableQuantumMetric to the switchboard

tomrf1:
In its own section for now because it's not in the `experiments` object.
We need to redesign/reimplement this page


TomPretty (approved):
👍

PR #256, opened 2021-11-24
Migrate switches config

tomrf1:
Co-authored-by: Rik Roots KaliedaRik@users.noreply.github.com

Currently support-frontend and support-admin-console need to agree on the model for the switches:
https://github.com/guardian/support-frontend/blob/b94e1e1c8a6539c666b5ed57198afd0eebcde38b/support-frontend/app/admin/settings/Switches.scala#L6

This causes issues if admin-console's model is out of date.
Instead, this PR proposes loosening the data model on the admin-console side, so that users can type in new switch names and assign them to arbitrary groups of switches.

This would be a backwards incompatible change, so we'd probably want to introduce a new version of the switches.json file in S3 for the migration.


app/models/SupportFrontendSwitches.scala
```diff
@@ -3,56 +3,26 @@ package models
...
 
-/**
-  * Based on https://github.com/guardian/support-frontend/blob/master/support-frontend/app/admin/settings/
-  */
+import scala.collection.immutable.IndexedSeq
 
-sealed trait SwitchState
-case object On extends SwitchState
-case object Off extends SwitchState
+sealed trait SwitchState extends EnumEntry
```
tomrf1, line 10: this bit is just (slightly) simplifying the `SwitchState` json support by using enumeratum

TomPretty (commented):
Yeah makes sense to me! 👍

app/models/SupportFrontendSwitches.scala
johnduffell: it's a nice idea but I'm not sure it's a big saving - I guess it just maps On to "On" and Off to "Off" transparently?

johnduffell (approved):
sounds sensible to me on the format

app/models/SupportFrontendSwitches.scala
tomrf1: I've just discovered circe now has a nicer way to do this, I've pushed up a commit

app/models/SupportFrontendSwitches.scala
```diff
@@ -2,58 +2,24 @@ package models
...
+  case class Switch(description: String, state: SwitchState)
+  type SwitchName = String
+  type SwitchGroup = Map[SwitchName,Switch]
+  type GroupName = String
+  type SupportFrontendSwitches = Map[GroupName,SwitchGroup]
 
-object SupportFrontendSwitches {
   implicit val customConfig: Configuration = Configuration.default.withDefaults
+  implicit val stateDecoder: Decoder[SwitchState] = deriveEnumerationDecoder[SwitchState]
+  implicit val stateEncoder: Encoder[SwitchState] = deriveEnumerationEncoder[SwitchState]
```
TomPretty, line 22: This is neat!

tomrf1: it's the dream!

public/src/components/switchboard.tsx
```diff
@@ -1,456 +1,211 @@
...
 
-  render(): React.ReactNode {
-    const { classes } = this.props;
+  const createGroupsFromData = (group: [string, SwitchGroup]): JSX.Element => {
+    const [groupId, groupData] = group;
 
     return (
+      <FormControl className={classes.formControl} key={groupId}>
+        <FormLabel>{groupData.description}</FormLabel>
+        {Object.entries(groupData.switches).map(mySwitch =>
```
tomrf1, line 136: I think it's best to destructure tuples asap, it saves having to do `switchData[0]` etc elsewhere. E.g.
```
Object.entries(groupData.switches).map(([switchName, switchData]) =>
  createSwitchesFromGroupData(switchName, switchData, groupId),
)
```

public/src/components/switchboard.tsx
```diff
@@ -1,456 +1,211 @@
...
+      </FormControl>
+    );
+  };
+
+  const createSwitchFields = (): JSX.Element => {
+    return <>{Object.entries(data).map(group => createGroupsFromData(group))}</>;
+  };
+
+  // This isn't working because original state doesn't get conserved in this component
+  // - Need to consider whether we need a 'reset' button?
```
tomrf1, line 148: We'd need to add a `resetData` callback to `withS3Data` to enable this, to re-fetch from S3.
Perhaps it's not necessary - the user can just refresh the page?

public/src/components/switchboard.tsx
```diff
@@ -1,456 +1,211 @@
...
+    groupId: string,
+  ): JSX.Element => {
+    const [switchId, switchData] = mySwitch;
+
+    return (
+      <FormControlLabel
+        label={switchData.description}
+        checked={switchData.state === 'On' ? true : false}
+        control={<SwitchUI />}
+        onChange={(event): void => updateSwitchSetting(groupId, mySwitch, event)}
```
tomrf1, line 123: you can avoid the `event: any` param in `updateSwitchSetting` if you just pass in a boolean here:
```
const checked = switchData.state === 'On';

return (
  <FormControlLabel
    label={switchData.description}
    checked={checked}
    control={<SwitchUI />}
    onChange={(event): void => updateSwitchSetting(groupId, mySwitch, !checked)}
    value={switchId}
    key={switchId}
  />
);
```

public/src/components/switchboard.tsx
KaliedaRik: I've taken out the refresh functionality - like you say, there's no need for it as user can just refresh page or navigate away and return

public/src/components/switchboard.tsx
KaliedaRik: Done and pushed fix

KaliedaRik:
This PR is ready for review, but should not be merged until related work on Support Frontend (updating the code to use v2 of the JSON object) is also ready to be merged.

### Purpose of PR
We have identified a need to move switch management functionality to a Single Source of Truth. Currently we are able to make changes to switches in both AWS S3 and in the RRCP. This is because switches have been hard-coded into the page. If someone adds a new switch to the JSON object in S3, that switch will not appear on the RRCP panel. If the RRCP switches page is saved, the change in S3  gets written over.

This change moves the JSON object to v2. The object now includes descriptions of the switches, and descriptions of the groups in which the switches are displayed. We can now use the JSON file to construct all the switches and their groups dynamically, removing all hard-coded switches from the code base. Thus the Single Source of Truth is achieved.

No significant changes have been made to the switches page display. Minor changes are:
+ The "refresh" button has been removed as it is redundant.
+ When a user updates a switch's state, a warning alert appears above the save button to remind the user that they need to shave their changes before they can take effect.

-

KaliedaRik:
### Important!
Before merging this PR we must create a new JSON object (v2) for the PROD environment and upload it to AWS S3

PR #257, opened 2021-11-26
Add US EOY Giving Tuesday banner to RRCP

marialani:
## What does this change?
This PR adds the US End of Year moment banner to the Reader Revenue Control Panel.

## Images
Screenshot 2021-11-26 at 12 24 52
Screenshot 2021-11-26 at 12 26 34

### DIFF
--- a/app/models/BannerTests.scala
+++ b/app/models/BannerTests.scala
@@ -23,37 +23,37 @@ object BannerTemplate extends Enum[BannerTemplate] with CirceEnum[BannerTemplate
 }
 
 case class BannerContent(
-    heading: Option[String],
-    messageText: String,
-    highlightedText: Option[String],
-    cta: Option[Cta],
-    secondaryCta: Option[SecondaryCta]
+  heading: Option[String],
+  messageText: String,
+  highlightedText: Option[String],
+  cta: Option[Cta],
+  secondaryCta: Option[SecondaryCta]
 )
 
 case class BannerVariant(
-    name: String,
-    template: BannerTemplate,
-    bannerContent: Option[BannerContent],
-    mobileBannerContent: Option[BannerContent],
+  name: String,
+  template: BannerTemplate,
+  bannerContent: Option[BannerContent],
+  mobileBannerContent: Option[BannerContent],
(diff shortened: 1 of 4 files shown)


app/models/BannerTests.scala
```diff
@@ -7,9 +7,10 @@ import io.circe.{Decoder, Encoder}
 
 import scala.collection.immutable.IndexedSeq
 
-
 sealed trait BannerTemplate extends EnumEntry
-object BannerTemplate extends Enum[BannerTemplate] with CirceEnum[BannerTemplate] {
+object BannerTemplate
+    extends Enum[BannerTemplate]
+    with CirceEnum[BannerTemplate] {
```
TomPretty, line 13: I think we'll probably want to drop the formatting changes in this file

app/models/BannerTests.scala
```diff
@@ -19,40 +20,41 @@ object BannerTemplate extends Enum[BannerTemplate] with CirceEnum[BannerTemplate
   case object InvestigationsMomentBanner extends BannerTemplate
   case object EnvironmentMomentBanner extends BannerTemplate
   case object UsEoyMomentBanner extends BannerTemplate
+  case object UsEoyMomentGivingTuesdayBanner extends BannerTemplate
```
TomPretty, line 23: just keep this line!

app/models/BannerTests.scala
marialani: Think this is a consequence of format on save, do we have lint rules for these scala files or would I need to disable my IDE settings to prevent changes?

TomPretty: We don't have any linting setup for scala. Scala devs seem to not like it so much 🤷‍♂️

marialani: Ah okay, no worries

app/models/BannerTests.scala
TomPretty: I think we've just got a few more formatting changes that we want to drop!

TomPretty (approved):
LGTM 👍

PR #259, opened 2021-12-10
Rjr add header tests page

KaliedaRik:
## What does this change?
This PR adds functionality which, together with complementary PRs in main site repositories, will allow non-engineers to update and AB test the main site's top-left header copy and CTAs via RRCP.

header-main-site

The functionality for this new feature draws heavily on existing functionality for banners - specifically the ability to: 
+ set geographical regions and supporter status for where the test will appear to the user
+ create controls/variants for each test, and to set the variant split for AB testing purposes
+ define different copy for desktop vs mobile devices
+ the ability to archive or delete each test, together with batch archive/delete functionality

The header area on the main page is very limited in size. The test form defines just two text fields: `heading` and `sub-heading`. It also allows the user to define between 0 and 2 CTA buttons with each button having copy and URL destination fields

New navigation
header-new-nav

Left-hand-side listings - these can be filtered by region
header-lhs-listings

The variants form (top) for heading/subheading and CTA fields
header-view-variants

The variants form (bottom) for mobile alternative copy and CTAs
header-view-mobile-alt

The CTA fields
header-cta-fields

For multi-variant tests, the split controls
header-view-split


public/src/components/channelManagement/headTests/headTestEditor.tsx
```diff
@@ -0,0 +1,211 @@
...
+  editMode: boolean;
+  onDelete: () => void;
+  onArchive: () => void;
+  onTestSelected: (testName: string) => void;
+  testNames: string[];
+  testNicknames: string[];
+  createTest: (newTest: HeadTest) => void;
+}
+
+const HeadTestEditor: React.FC<HeadTestEditorProps> = ({
```
tomrf1, line 39: why `HeadTest` on the client? It's inconsistent with the scala and SDC, where it's `HeaderTest`

public/src/components/channelManagement/headTests/headTestNewVariantButton.tsx
```diff
@@ -0,0 +1,146 @@
+import React from 'react';
```
tomrf1, line 1: this file isn't used

public/src/components/channelManagement/headTests/headTestVariantEditorSecondaryCtaEditor.tsx
```diff
@@ -0,0 +1,86 @@
+import React from 'react';
```
tomrf1, line 1: this file looks the same as `variantEditorCtaEditor.ts`

KaliedaRik: This new file is a copy of `../variantEditorCtaEditor.ts` with just the interface name changed. I ran into some compile issues around primary and secondary CTA types (they differ in banners) which got solved when I made a local copy of the variantEditor file for the header secondaryCTA buttons. I'm not proud of the change but it fixed the types issue; I'm happy to get rid of this fix if you can help me solve those type errors.

tomrf1: ah I see. I'll have a think...

public/src/components/channelManagement/headTests/headTestEditor.tsx
KaliedaRik: Happy to revert to `Header...` (and `header...`) across all files (including filenames)

public/src/components/channelManagement/headTests/headTestNewVariantButton.tsx
KaliedaRik: That's interesting! Checking through, I can see that `testNewVariantButton.tsx` is a copy of`banner/bannerTestNewVariantButton.tsx` (only file paths adjusted for folder nesting) which is possibly why I didn't catch this. Happy to delete `head/headTestNewVariantButton.tsx`, but if someone deletes `testNewVariantButton.tsx` in the future then that would break our new header tests page display.

tomrf1: Just checked this - `banner/bannerTestNewVariantButton.tsx` is also not used and can be deleted!
`testNewVariantButton.tsx` (used by `testVariantsEditor.tsx`) is used by epic/banner/header.
We have tried to share as much as possible between the epic + banner tools, and the header should also.

tomrf1: > but if someone deletes testNewVariantButton.tsx in the future then that would break our new header tests page display.
not sure why anyone would do this, but TS would catch it!

public/src/components/channelManagement/headTests/headTestVariantEditorSecondaryCtaEditor.tsx
tomrf1: it looks like this is no longer an issue, I just swapped `HeadTestVariantEditorSecondaryCtaEditor` for `VariantEditorCtaEditor` and it works fine, so I think you can delete this file.

public/src/components/channelManagement/headTests/headTestNewVariantButton.tsx
KaliedaRik: Will delete

public/src/components/channelManagement/headTests/headTestVariantEditorSecondaryCtaEditor.tsx
KaliedaRik: I'll give it a go

KaliedaRik: Done

public/src/components/channelManagement/headerTests/headerTestEditor.tsx
```diff
@@ -0,0 +1,211 @@
...
+        setValidationStatusForField(variant.name, isValid)
+      }
+    />
+  );
+
+  const renderVariantSummary = (variant: HeaderVariant): React.ReactElement => (
+    <TestEditorVariantSummary
+      name={variant.name}
+      testName={test.name}
+      testType="BANNER"
```
tomrf1, line 117: `HEADER`

public/src/components/channelManagement/headerTests/headerTestVariantEditor.tsx
```diff
@@ -0,0 +1,286 @@
...
+const HeaderTestVariantEditor: React.FC<HeaderTestVariantEditorProps> = ({
+  variant,
+  editMode,
+  onValidationChange,
+  onVariantChange,
+}: HeaderTestVariantEditorProps) => {
+  const classes = useStyles();
+  const setValidationStatusForField = useValidation(onValidationChange);
+
+  const content: HeaderContent = variant.content || {
```
tomrf1, line 213: the `content` field is mandatory in the model, so the `|| {...}` stuff can be removed

public/src/components/channelManagement/headerTests/utils/defaults.ts
```diff
@@ -0,0 +1,66 @@
...
+export const DEFAULT_SECONDARY_CTA: Cta = {
+  text: 'Support the Guardian',
+  baseUrl: 'https://support.theguardian.com/contribute',
+};
+
+const DEV_AND_CODE_DEFAULT_VARIANT: HeaderVariant = {
+  name: 'CONTROL',
+  content: {
+    heading: 'Lorem Ipsum',
+    subheading: 'Exam desut lineas buteram loas',
```
tomrf1, line 20: We use realistic copy for default values in all environments, just in case...
Let's go with:
```
heading: 'Support the Guardian',
subheading: 'Available for everyone, funded by readers',
```

public/src/components/channelManagement/headerTests/headerTestEditor.tsx
KaliedaRik: Fixed, alongside adding 'HEADER' to the TestType type in helpers/shared.ts file

public/src/components/channelManagement/headerTests/headerTestVariantEditor.tsx
KaliedaRik: Done

public/src/components/channelManagement/headerTests/utils/defaults.ts
KaliedaRik: Updated both variants where LoremIpsum copy was being used

tomrf1 (approved):
great work!

PR #261, opened 2021-12-16
snyk action

tomrf1:


kelvin-chappell (approved):
👍

public/src/components/channelManagement/epicTests/sectionsEditor.tsx
```diff
@@ -23,7 +23,7 @@ export const SectionsEditor: React.FC<SectionEditorProps> = ({
   disabled,
   ids,
   onUpdate,
-}) => {
+}: SectionEditorProps) => {
```
tomrf1, line 26: I'm not sure why these only just started causing the build to fail

PR #270, opened 2021-12-20
implement per variant default choice card contribution frequency

michaelbjacobson:
## What does this change?
Allows per variant configuration of default contribution amount choice card.


app/models/EpicTests.scala
```diff
@@ -23,6 +23,15 @@ object TickerCountType extends Enum[TickerCountType] with CirceEnum[TickerCountT
   case object people extends TickerCountType
 }
 
+sealed trait ContributionFrequency extends EnumEntry
```
tomrf1, line 26: btw circe can now do enums with slightly neater syntax, see https://github.com/guardian/support-admin-console/blob/main/app/models/SupportFrontendSwitches.scala#L8
I realise lots of places still do it this way though!
E.g.
```
sealed trait ContributionFrequency
case object ONE_OFF extends ContributionFrequency
case object MONTHLY extends ContributionFrequency
case object ANNUAL extends ContributionFrequency
```
and then further down next to the Encoders/Decoders:
```
implicit val stateDecoder: Decoder[ContributionFrequency] = deriveEnumerationDecoder[ContributionFrequency]
implicit val stateEncoder: Encoder[ContributionFrequency] = deriveEnumerationEncoder[ContributionFrequency]
```

tomrf1 (approved):
looks good. Will this work with the new way of rendering choice cards from an iframe?

michaelbjacobson:
> looks good. Will this work with the new way of rendering choice cards from an iframe?

I've no idea, I think I must have been off when that change was made! I manually tested this change in `CODE` using the web preview functionality and it all worked perfectly. Do you know who did the iframe work?

tomrf1:
> > looks good. Will this work with the new way of rendering choice cards from an iframe?
> 
> I've no idea, I think I must have been off when that change was made! I manually tested this change in `CODE` using the web preview functionality and it all worked perfectly. Do you know who did the iframe work?

I don't think the iframe stuff is live yet, @TomPretty will know more!

app/models/EpicTests.scala
michaelbjacobson: @tomrf1 thanks, this is v good to know for the future, but I realised that this type had already been created elsewhere in the codebase and named `ContributionType`.

tomrf1: oh, good spot!

PR #271, opened 2021-12-24
Update epic default highlighted text

KaliedaRik:
## What does this change?
A small copy change to the Epic default highlighted text


tomrf1 (approved):
👍

PR #276, opened 2022-01-18
Rjr update user guide links

KaliedaRik:
The link to the User Guide currently resides at the bottom of the links in the hamburger navigation. Marketing colleagues rarely remember that the UG exists. This PR aims to fix the situation by moving the UG link to the right hand side of the top navigation. 

The change will also act as a visual reminder to engineering colleagues that the UG exists, and needs to be updated as changes are made to the RRCP functionality and UX.

Furthermore, an additional button link (to an incorrect document) appears in the bottom right of the screen when the user edits tests in certain channels. This PR removes this button as it is no longer required. It also fixes the layout of the sticky bar messaging as users no longer require text prompts on how to view a test in Read Only mode - this information is given in better detail in the User Guide.

**Before screenshots:**

Screenshot 2022-01-18 at 12 17 56

Screenshot 2022-01-18 at 12 18 13

Screenshot 2022-01-18 at 12 18 38

**After screenshots:**

Screenshot 2022-01-18 at 12 21 26

Screenshot 2022-01-18 at 12 21 52


tomrf1 (approved):
great! Much clearer

PR #277, opened 2022-01-26
Add Remirror Rich Text Editor to RRCP forms

KaliedaRik:
## Current situation
The current Epic and Banner variant forms include a number of text fields where Marketing colleagues can enter copy. Some of these fields are text <input> fields; others are <TextArea> fields. In all cases the functionality of these fields is non-existent: if Marketing colleagues want to highlight phrases, or wrap a phrase in a link, then they need to insert raw HTML code into the copy to make this happen.

The fields also support three %%PLACEHOLDERS%% which are converted into page view counts, appropriate currency symbols, etc down the line when the Epic or Banner displays in a reader's browser/device. Again, Marketing colleagues need to insert these placeholders manually into the copy

Screenshot 2022-02-04 at 14 38 55

Screenshot 2022-02-04 at 14 39 40

## What does this PR change?
[Scoping document, for context](https://docs.google.com/document/d/1knxb1s7QIgHpk_4Ml6TV8rntdw2qqMZ0Fxfj8dq-pIg/edit#heading=h.j5pltcq996sd)

We are updating the current text input fields with a Rich Text Editor (ProseMirror, as wrapped for React by the ReMirror library). The RTE will make it easier for Marketing colleagues to add **bold** and _italic_ markup to their copy, and more easily add and manage inline links within the copy. Placeholders can also be added to the copy using dedicated placeholder buttons.

The new functionality also includes checking the fields for excessive (or a lack of) copy in a given field.

Screenshot 2022-02-04 at 14 59 20

Screenshot 2022-02-04 at 14 59 39


public/src/components/channelManagement/richTextEditor.tsx
```diff
@@ -0,0 +1,407 @@
...
+      new LinkExtension({ autoLink: true }),
+      new TextHighlightExtension(),
+    ],
+    content: parseCopyForParagraphs(copyData),
+    selection: 'start',
+    stringHandler: 'html',
+  });
+
+  // Handle to the ProseMirror <div> element containing all the RTE text
+  let ref: HTMLDivElement | null = document.createElement('div');
```
tomrf1, line 361: This feels like it goes against the spirit of remirror. I need to read the docs more but I think we should use their api to listen for events and get the content.
The docs are a bit light but it seems you can use the `EventsExtension` to listen for the blur event, e.g:
```
const eventsExtension = new EventsExtension();
eventsExtension.addHandler('blur', () => { })
// add eventsExtension to extensions array
```

And I think you can use getHTML -
https://remirror.io/docs/getting-started/commands-and-helpers/#helpers

KaliedaRik: The current extraction code is the code I'm least happy with. Will look at the EventsExtension as suggested.

public/src/components/channelManagement/richTextEditor.tsx
```diff
@@ -0,0 +1,407 @@
...
+  const active = useActive();
+
+  return (
+    <div>
+      <span className={classes.fieldLabel}>{label != null ? label : 'Editable field'}</span>
+      {!disabled && (
+        <>
+          <button
+            className={active.bold() ? 'remirror-button-active' : 'remirror-button'}
+            // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
tomrf1, line 251: the linter doesn't appear to care about these (`npm run lint`) - can they be removed?

public/src/components/channelManagement/richTextEditor.tsx
```diff
@@ -0,0 +1,407 @@
...
+// ReMirror/ProseMirror LINK functionality
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+function useLinkShortcut() {
+  const [linkShortcut, setLinkShortcut] = useState<ShortcutHandlerProps | undefined>();
+  const [isEditing, setIsEditing] = useState(false);
+
+  useExtension(
+    LinkExtension,
+    ({ addHandler }) =>
+      addHandler('onShortcut', props => {
```
tomrf1, line 74: what is the shortcut key?

public/src/components/channelManagement/richTextEditor.tsx
```diff
@@ -0,0 +1,407 @@
...
+          : [
+              // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+              { type: ComponentItem.ToolbarButton, onClick: () => clickEdit(), icon: 'link' },
+            ],
+      },
+    ],
+    [clickEdit, onRemove, activeLink],
+  );
+
+  const items: ToolbarItemUnion[] = useMemo(() => linkEditItems, [linkEditItems]);
```
tomrf1, line 195: I realise this is taken from the example, but do you know why we need `items` here? It's just a copy of `linkEditItems`, which itself is memoized...

public/src/components/channelManagement/epicTests/epicTestVariantEditor.tsx
```diff
@@ -101,33 +107,43 @@ const EpicTestVariantEditor: React.FC<EpicTestVariantEditorProps> = ({
   useEffect(() => {
     const isValid = Object.keys(errors).length === 0;
     onValidationChange(isValid);
-  }, [
```
tomrf1, line 104: with this change we've lost validation, e.g. I can now save an invalid template in the text. I'll have a think about how we can make the new component work with the existing validation

public/src/components/channelManagement/richTextEditor.tsx
KaliedaRik: I'll see if they can be removed without linter complaints

public/src/components/channelManagement/richTextEditor.tsx
KaliedaRik: Good question. I'll find an answer as this is the sort of thing we'd want to document in the User Guide.

public/src/components/channelManagement/epicTests/epicTestVariantEditor.tsx
KaliedaRik: For context, validating the copy in the RTE (for either zero length or excessive length) caused me endless headaches because the current code uses react-hook-form for this work. Sadly RHF expects the field it is checking to be some sort of input element (with a .value attribute, I think) whereas ProseMirror avoids <input> and <textarea> and uses contetEditable <div> elements instead. I'll check to make sure I've correctly plumbed in the replacement validation code - for four of those five fields - into this useEffect hook.

public/src/components/channelManagement/richTextEditor.tsx
KaliedaRik: I'll look into this and see if there's efficiencies to be made in the links-related code

public/src/components/channelManagement/epicTests/epicTestVariantEditor.tsx
tomrf1: react-hook-form provides a way to handle custom inputs - https://react-hook-form.com/api/usecontroller/controller#main
If we are to continue using react-hook-form in this codebase then we should wrap the new input with this.

I've just had a play and this works for `paragraphs` (note the use of `templateValidator`):
```
troller
  name="paragraphs"
  control={control}
  rules={{
    required: true,
    validate: (pars: string[]) => pars.map(templateValidator).find(result => result !== true),
  }}
  render={(data) => {
    return <RichTextEditor
      error={errors.paragraphs !== undefined}
      // @ts-ignore -- react-hook-form doesn't believe it has a message field
      helperText={errors?.paragraphs?.message || getParagraphsHelperText()}
      copyData={data.value}
      updateCopy={pars => {
        data.onChange(pars);
        handleSubmit(onSubmit)()
      }}
      name="paragraphs"
      label="Body copy"
      disabled={!editMode}
    />
  }}
/>
```

Happy to pair on this - it's quite fiddly

public/src/components/channelManagement/richTextEditor.tsx
KaliedaRik: 11 warning lines removed, linter not complaining.

public/src/components/channelManagement/richTextEditor.tsx
KaliedaRik: Looking at the code, they're using two floating toolbars, and they have a useMemo for each of them. One appears when text gets selected and offers the user a chance to create a link; the other shows up when the user clicks in an already-created link and offers them the chance to edit/delete the link.

`      <FloatingToolbar items={items} positioner="selection" placement="top" enabled={!isEditing} />
      <FloatingToolbar
        items={linkEditItems}
        positioner={linkPositioner}
        placement="bottom"
        enabled={!isEditing && empty}
      />
`

public/src/components/channelManagement/richTextEditor.tsx
tomrf1: In terms of getting the content out, here's the example hook which you pass into the Remirror component:
```
const hooks = [
  () => {
    const { getHTML } = useHelpers();

    const handleSaveShortcut = useCallback(
      ({ state }) => {
        console.log(JSON.stringify(getHTML(state)));
        return true; // Prevents any further key handlers from being run.
      },
      [getText],
    );

    // "Mod" means platform agnostic modifier key - i.e. Ctrl on Windows, or Cmd on MacOS
    useKeymap('Mod-s', handleSaveShortcut);
  },
];
```
The challenge is to trigger this onBlur rather than on key press, and I'm not sure how to do that with the `EventsExtension`

public/src/components/channelManagement/richTextEditor.tsx
KaliedaRik: I'll play around with deleting things and see what occurs

public/src/components/channelManagement/richTextEditor.tsx
tomrf1: Actually, I've found a way to use the EventsExtension in the hook-
```
const hooks = [
  () => {
    const { getText, getHTML } = useHelpers();

    const handleSaveShortcut = useCallback(
      ({ state }) => {
        console.log(JSON.stringify(getHTML(state)));
        return true; // Prevents any further key handlers from being run.
      },
      [getHTML],
    );
    manager.getExtension(EventsExtension).addHandler('blur', handleSaveShortcut);
  },
];
```

public/src/components/channelManagement/richTextEditor.tsx
```diff
@@ -0,0 +1,407 @@
...
+});
+
+// Typescript
+interface RichTextEditorProps extends WithStyles<typeof styles> {
+  disabled: boolean;
+  label?: string;
+  helperText?: string;
+  name?: string;
+  error: boolean;
+  updateCopy: (item?: string[]) => void;
```
tomrf1, line 56: It would be nice if the types of the input value and the update value were consistent, and also if we could be explicit about whether we're working with `string` or `string[]`. This would avoid the need for e.g. `highlightedText: updatedHighlightedText[0]` throughout the codebase.

What about if the `RichTextEditor` deals only with `string`, which may contain html such as `<p>`.
And then for the special case of the `paragraphs` fields there's a wrapper component (`ParagraphsRichTextEditor`?) which deals with `string[]` and just does the transformations for the `<p>` tags, and wraps `RichTextEditor`.

tomrf1: Just to illustrate what I mean -
```
interface ParagraphsRichTextEditor {
  paragraphs: string[];
  onUpdate: (value: string[]) => void;
  ...
}
const ParagraphsRichTextEditor: React.FC<ParagraphsRichTextEditor> = ({
  paragraphs,
  onUpdate,
  ...
}) => {
  return (<RichTextEditor
    value={parseCopyForParagraphs(paragraphs)}
    onUpdate={(updatedValue: string) => onUpdate(htmlToParagraphs(updatedValue))}
    ...
  />)
}
```

public/src/components/channelManagement/richTextEditor.tsx
KaliedaRik: Fix applied. I've also had to go through the code base adding in the `paragraphs[]` attribute alongside the deprecated `messageText` attribute - which will need to be removed from the code base after we update all the LIVE/CODE/DEV banner test JSON files to use`paragraphs[]`

public/src/components/channelManagement/epicTests/epicTestVariantEditor.tsx
KaliedaRik: I think I've implemented this correctly for `epicTestVariantEditor.tsx` - needs more testing. Also need to implement this fix for `bannerTestVariantEditor.tsx`

public/src/components/channelManagement/richTextEditor.tsx
KaliedaRik: Given that I've already done work to update banners to accept a replacement `paragraphs` attribute to (eventually) replace `messageText` in the SDC repo, I've gone with a solution that gets rid of code like `highlightedText: updatedHighlightedText[0]` to `highlightedText: updatedHighlightedText.join(' ')`.  It could be refactored, but I'd rather get this solution working first.

public/src/models/banner.ts
```diff
@@ -23,7 +23,8 @@ export enum BannerTemplate {
 
 export interface BannerContent {
   heading?: string;
-  messageText: string;
+  messageText?: string;
+  paragraphs?: string[];
```
tomrf1, line 27: why is `paragraphs` being added to the banner model in this PR?

public/src/components/channelManagement/richTextEditor.tsx
```diff
@@ -0,0 +1,431 @@
...
+            }
+            if (p.tagName === 'P') {
+              return true;
+            }
+            return false;
+          });
+
+          const paragraphsCopy = paragraphs.map(p => p.innerHTML);
+
+          // Get rid of unwanted markup created by ProseMirror
```
tomrf1, line 358: this is unnecessary now we're using `getHTML`

public/src/components/channelManagement/epicTests/epicTestVariantEditor.tsx
```diff
@@ -70,6 +74,15 @@ interface EpicTestVariantEditorProps {
   onValidationChange: (isValid: boolean) => void;
 }
 
+interface EpicTestMuiTextFields {
```
tomrf1, line 77: we should be able to go back to the original `FormData` type now

public/src/components/channelManagement/richTextEditor.tsx
tomrf1: I don't think the work to add `paragraphs` to banners is very relevant to this suggestion.
The various uses of the RTE will be simpler, and the types clearer, if `RichTextEditor` deals only with `string`. A simple wrapper component that does the transformation for <p> tags can be used where a `string[]` is required.
I'm happy to pair on this.

public/src/components/channelManagement/richTextEditor.tsx
tomrf1: As far as I can tell there is no default shortcut for links. I suggest deleting `useLinkShortcut` from this file

public/src/components/channelManagement/epicTests/epicTestVariantEditor.tsx
```diff
@@ -79,17 +92,18 @@ const EpicTestVariantEditor: React.FC<EpicTestVariantEditorProps> = ({
 }: EpicTestVariantEditorProps) => {
   const classes = getUseStyles(epicEditorConfig.allowMultipleVariants)();
 
-  const templateValidator = templateValidatorForPlatform(epicEditorConfig.platform);
+  // Handling MUI TextField updates
+  const templateValidator = templateValidatorForPlatform('DOTCOM');
```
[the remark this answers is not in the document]
KaliedaRik: I was confused? This could be an error introduced by me copying the equivalent code in bannerTestVariantEditor into this file.

tomrf1 (approved):
Wonderful! A great leap forward

PR #278, opened 2022-01-27
Enable viewing of other tests while editing

tomrf1:
Currently when you start editing a test, you cannot view other tests.
This is because the Save button applies to only that test, but in reality all tests are saved to a single json file, so we cannot allow a user to selectively save individual tests without making a larger architectural change.

A quick compromise is to enable read-only viewing of other tests, but the Save button still applies to the single test being edited.

### Viewing the test that is being edited:
Screen Shot 2022-01-27 at 15 13 18

### Viewing another test (unable to edit):
Screen Shot 2022-01-27 at 15 13 23


KaliedaRik (approved):
All seems fine to me

PR #279, opened 2022-01-31
Header tests - mobile content

tomrf1:
The code is largely lifted from the banner tool.

Note - although the model allows for everything to be customed on mobile (e.g. heading), only the primary cta is ever displayed on mobile in the header. So the tool only allows users to configure the primary cta.

Screen Shot 2022-02-01 at 14 43 59

PR to use this field in the SDC component: https://github.com/guardian/support-dotcom-components/pull/622

### DIFF
--- a/public/src/components/channelManagement/headerTests/headerTestVariantEditor.tsx
+++ b/public/src/components/channelManagement/headerTests/headerTestVariantEditor.tsx
@@ -50,6 +50,7 @@ const useStyles = makeStyles(({ palette, spacing }: Theme) => ({
   },
   buttonsContainer: {
     marginTop: spacing(2),
+    marginLeft: spacing(2),
   },
 }));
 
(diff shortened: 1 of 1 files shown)


public/src/components/channelManagement/headerTests/headerTestVariantEditor.tsx
```diff
@@ -219,6 +246,39 @@ const HeaderTestVariantEditor: React.FC<HeaderTestVariantEditorProps> = ({
...
+        <FormControlLabel
+          value="enabled"
+          key="enabled"
+          control={<Radio />}
+          label="Show different copy on mobile"
+          disabled={!editMode}
+        />
+      </RadioGroup>
+      {variant.mobileContent && (
+        <HeaderTestVariantContentEditor
```
tomrf1, line 270: I've just realised that we only show the primary cta on mobile. So the tool should not show fields for heading/subheading + 2nd cta

tomrf1: I've pushed a commit for this

KaliedaRik (approved):
All sensible stuff

PR #284, opened 2022-02-09
Remove deprecated banner fields

tomrf1:
The `BannerVariant` model has an optional seperate field for mobile content:
```
bannerContent: BannerContent;
mobileBannerContent?: BannerContent;
```

Historically this was not the case, and in the migration to support both we left some legacy fields in the model.
This PR removes them - they've not been used for a while.

I've archived some very old draft tests which still had the old model


KaliedaRik (approved):
Sensible updates.

PR #286, opened 2022-02-16
RJR - support paragraphs in banner body copy

KaliedaRik:
## What does this change?
All Banners currently use a `messageText` String field to hold the copy to be used in the Banner's body. Different Banners also use different approaches to adding a sentence of highlighted text to the body copy. Compare this to the situation in Epics, where body copy is supplied in an array of Strings (in a `paragraphs` attribute), which allows Epics to break their body copy into paragraphs.

This PR aims to bring the handling of body copy into alignment between Banners and Epics, with copy supplied in a new `paragraphs` array. It also introduces a shared function to handle the addition of highlighted text to the last paragraph of the body copy.

### Risks
Body copy is generated by Marketing colleagues using the RRCP. This PR is supported by [an equivalent PR in the support.dotcom.components repo](https://github.com/guardian/support-dotcom-components/pull/632) to update the Banner variant forms so that they save their body content as an array of Strings in the new `paragraphs` attribute. (link to SAC PR required)

A major risk for this PR is handling the transition of existing Banner tests from the old way of doing things (using `messageText`) to the new way (using `paragraphs`). We mitigate that risk in this PR by including temporary functionality so that Banner tests that do not include the `paragraphs` copy automatically fall back to using the `messageText` copy.

Once we are satisfied that all Banner tests on the LIVE site have been updated to include their copy in the `paragraphs` attribute we will be able to remove the temporary functionality and simplify the code base.

### Testing
This PR can be tested using the RRCP in DEV or CODE.

### Images
(Add some images to compare existing vs new paragraph functionality)

### DIFF
--- a/public/src/components/channelManagement/bannerTests/bannerTestEditor.tsx
+++ b/public/src/components/channelManagement/bannerTests/bannerTestEditor.tsx
@@ -23,7 +23,10 @@ import TestVariantsSplitEditor from '../testVariantsSplitEditor';
 import { useStyles } from '../helpers/testEditorStyles';
 
 const copyHasTemplate = (content: BannerContent, template: string): boolean =>
-  (content.heading && content.heading.includes(template)) || content.messageText.includes(template);
+  (content.heading && content.heading.includes(template)) ||
+  (content.paragraphs != null && content.paragraphs[0].includes(template)) ||
+  (content.messageText != null && content.messageText.includes(template));
+
 const testCopyHasTemplate = (test: BannerTest, template: string): boolean =>
   test.variants.some(
     variant =>
--- a/public/src/components/channelManagement/bannerTests/bannerTestEditor.tsx
+++ b/public/src/components/channelManagement/bannerTests/bannerTestEditor.tsx
@@ -24,7 +24,7 @@ import { useStyles } from '../helpers/testEditorStyles';
 
 const copyHasTemplate = (content: BannerContent, template: string): boolean =>
   (content.heading && content.heading.includes(template)) ||
-  (content.paragraphs != null && content.paragraphs[0].includes(template)) ||
+  (content.paragraphs && content.paragraphs.some(para => para.includes(template))) ||
   (content.messageText != null && content.messageText.includes(template));
 
 const testCopyHasTemplate = (test: BannerTest, template: string): boolean =>
--- a/public/src/components/channelManagement/bannerTests/bannerTestVariantEditor.tsx
+++ b/public/src/components/channelManagement/bannerTests/bannerTestVariantEditor.tsx
@@ -97,6 +97,12 @@ interface BannerTestVariantContentEditorProps {
   deviceType: DeviceType;
 }
 
+interface FormData {
+  heading?: string;
+  body: string;
+  highlightedText?: string;
+}
+
 const BannerTestVariantContentEditor: React.FC<BannerTestVariantContentEditorProps> = ({
   content,
   template,
@@ -109,13 +115,26 @@ const BannerTestVariantContentEditor: React.FC<BannerTestVariantContentEditorPro
 
   const templateValidator = templateValidatorForPlatform('DOTCOM');
 
-  const defaultValues: BannerContent = {
+  const getParagraphsString = (
+    paragraphs: string[] | null | undefined,
+    messageText: string | null | undefined,
+  ): string => {
+    if (Array.isArray(paragraphs)) {
+      return paragraphs.join('\n');
+    }
+    if (paragraphs != null) {
+      return paragraphs;
+    }
+    return messageText || '';
+  };
+
+  const defaultValues: FormData = {
     heading: content.heading || '',
-    paragraphs: content.paragraphs || [content.messageText] || [],
+    body: getParagraphsString(content.paragraphs, content.messageText),
     highlightedText: content.highlightedText || '',
   };
 
-  const { register, handleSubmit, errors, trigger } = useForm<BannerContent>({
+  const { register, handleSubmit, errors, trigger } = useForm<FormData>({
     mode: 'onChange',
     defaultValues,
   });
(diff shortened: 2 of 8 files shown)


public/src/components/channelManagement/bannerTests/bannerTestEditor.tsx
```diff
@@ -23,7 +23,10 @@ import TestVariantsSplitEditor from '../testVariantsSplitEditor';
 import { useStyles } from '../helpers/testEditorStyles';
 
 const copyHasTemplate = (content: BannerContent, template: string): boolean =>
-  (content.heading && content.heading.includes(template)) || content.messageText.includes(template);
+  (content.heading && content.heading.includes(template)) ||
+  (content.paragraphs != null && content.paragraphs[0].includes(template)) ||
```
KaliedaRik, line 27: Not happy with this line - I think that we need to be testing for the template variable against all paragraphs joined into a string?

public/src/components/channelManagement/bannerTests/bannerTestVariantEditor.tsx
```diff
@@ -184,10 +199,14 @@ const BannerTestVariantContentEditor: React.FC<BannerTestVariantContentEditorPro
               required: EMPTY_ERROR_HELPER_TEXT,
               validate: templateValidator,
             })}
-            error={errors.messageText !== undefined}
-            helperText={errors.messageText ? errors.messageText.message : BODY_DEFAULT_HELPER_TEXT}
+            error={errors.paragraphs !== undefined}
+            helperText={
+              errors.paragraphs && errors.paragraphs[0]
```
KaliedaRik, line 204: Another 'not happy with' comment - this time I'm not sure about how the error checker records error messages for an array of copy. Need to check with how it works with the Epic paragraphs array code in main branch.

public/src/components/channelManagement/bannerTests/utils/defaults.ts
```diff
@@ -30,6 +33,7 @@ const PROD_DEFAULT_VARIANT: BannerVariant = {
   template: BannerTemplate.ContributionsBanner,
   bannerContent: {
     messageText: '',
+    paragraphs: [''],
```
KaliedaRik, line 36: Need to check to see if setting paragraphs to an empty array would be the better thing to do here.

public/src/components/channelManagement/bannerTests/bannerTestEditor.tsx
tomrf1: Just check all paras in the array -
`(content.paragraphs && content.paragraphs.some(para => para.includes(template))`

public/src/components/channelManagement/bannerTests/bannerTestVariantEditor.tsx
tomrf1: For the epic it's actually maintained as a string up until `onSubmit` - see [here](https://github.com/guardian/support-admin-console/blob/main/public/src/components/channelManagement/epicTests/epicTestVariantEditor.tsx#L119).

We'll change that when the new RichTextEditor component is added, but for now I suggest following what the epic does. This unfortunately involves having a [separate type](https://github.com/guardian/support-admin-console/blob/main/public/src/components/channelManagement/epicTests/epicTestVariantEditor.tsx#L58) for the `useForm`.

public/src/components/channelManagement/bannerTests/utils/defaults.ts
tomrf1: I would go with empty array, and remove `messageText` entirely

KaliedaRik: Done

public/src/components/channelManagement/bannerTests/bannerTestVariantEditor.tsx
KaliedaRik: Banner body copy now follows Epic functionality

public/src/components/channelManagement/bannerTests/bannerTestEditor.tsx
KaliedaRik: Done

public/src/components/channelManagement/bannerTests/utils/defaults.ts
```diff
@@ -30,6 +33,7 @@ const PROD_DEFAULT_VARIANT: BannerVariant = {
   template: BannerTemplate.ContributionsBanner,
   bannerContent: {
     messageText: '',
```
tomrf1, line 35: `messageText` needs to be removed from the defaults now

KaliedaRik: Done!

public/src/components/channelManagement/bannerTests/bannerTestVariantEditor.tsx
```diff
@@ -127,10 +146,10 @@ const BannerTestVariantContentEditor: React.FC<BannerTestVariantContentEditorPro
   useEffect(() => {
     const isValid = Object.keys(errors).length === 0;
     onValidationChange(isValid);
-  }, [errors.messageText, errors.heading, errors.highlightedText]);
+  }, [errors.body, errors.heading, errors.highlightedText]);
 
-  const onSubmit = ({ heading, messageText, highlightedText }: BannerContent): void => {
-    onChange({ ...content, heading, messageText, highlightedText });
+  const onSubmit = ({ heading, body, highlightedText }: FormData): void => {
+    onChange({ ...content, heading, paragraphs: body.split('\n'), highlightedText });
```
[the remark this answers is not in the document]
KaliedaRik: Would this change lead to the removal of cta and secondaryCta data each time someone updates the copy fields? An alternative approach to getting rid of `messageText` may be to explicitly (and temporarily) set it to null?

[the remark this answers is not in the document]
tomrf1: No - this component updates the content only, see the type of `onChange` -
`onChange: (updatedContent: BannerContent) => void;`

Also, TS would tell us if we were removing any required fields.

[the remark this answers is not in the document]
tomrf1: oh sorry, just realised ctas are on the content model!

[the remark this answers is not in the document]
tomrf1: ok in that case I agree with explicitly removing messageText (for now)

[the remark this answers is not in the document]
tomrf1: side note - it is odd that `onSubmit` doesn't handle updates to the ctas, but let's not worry about that now!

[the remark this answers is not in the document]
KaliedaRik: Done

tomrf1 (approved):
great work!

PR #287, opened 2022-02-25
Do not cache the page

tomrf1:
There's not really any benefit to browser caching of the page - traffic is tiny.
And currently users may need to hard-refresh to see the latest features.


KaliedaRik (approved):
Happy to approve

PR #288, opened 2022-02-25
Campaigns editor

tomrf1:
Simple tool for creating/deleting campaigns.

Still to do - make these campaigns available when creating epic/banner/header tests, so that we can automatically add the name as a prefix.

Main view:
Screen Shot 2022-02-25 at 14 39 57

Create dialog:
Screen Shot 2022-02-25 at 14 41 00

Delete dialog:
Screen Shot 2022-02-25 at 14 41 10


public/src/components/drawer.tsx
```diff
@@ -177,6 +177,11 @@ export default function NavDrawer(): React.ReactElement {
             <ListItemText primary="Channel Switches" />
           </ListItem>
         </Link>
+        <Link key="Campaigns" to="/campaigns" className={classes.link}>
+          <ListItem className={classes.listItem} button key="Campaigns">
+            <ListItemText primary="Campaigns" />
+          </ListItem>
+        </Link>
       </div>
```
KaliedaRik, line 185: I was wondering if it would be better to put the 'Campaigns' link at the top of the list? Possibly under its own 'Campaigns management' heading?
Screenshot 2022-03-07 at 12 15 30

tomrf1: I'll move it to the top. Let's see where this goes before giving it its own section - it's currently a very simple feature

PR #292, opened 2022-03-02
Campaign selector

tomrf1:
[PR for new campaigns tool](https://github.com/guardian/support-admin-console/pull/288).
Adds a campaigns selector to the modal for creating new tests.
If a campaign is selected then it is added as a prefix, with `__` in between. Tests with the same prefix will be grouped together in the AB test dashboard for analysis.
Screen Shot 2022-03-02 at 15 36 09
Screen Shot 2022-03-02 at 15 36 17
Screen Shot 2022-03-02 at 15 36 28

### DIFF
--- a/public/src/components/channelManagement/createTestDialog.tsx
+++ b/public/src/components/channelManagement/createTestDialog.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import { useForm } from 'react-hook-form';
 import {
   Button,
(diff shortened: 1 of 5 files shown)


public/src/components/channelManagement/createTestDialog.tsx
```diff
@@ -51,7 +62,7 @@ interface CreateTestDialogProps extends WithStyles<typeof styles> {
   existingNames: string[];
   existingNicknames: string[];
   mode: Mode;
-  testNamePrefix?: string;
+  testNamePrefix?: string; // set if all tests must have the same prefix
```
tomrf1, line 65: this field is only used for the 'holdback' epic, which has prefix `HOLDBACK__`

public/src/components/channelManagement/createTestDialog.tsx
```diff
@@ -90,6 +119,33 @@ const CreateTestDialog: React.FC<CreateTestDialogProps> = ({
         </IconButton>
       </div>
       <DialogContent dividers>
+        {testNamePrefix === undefined && (
```
tomrf1, line 122: if `testNamePrefix` is defined then we don't allow a campaign (for 'holdback' epic only)

KaliedaRik (approved):
Looks fine to me!

PR #297, opened 2022-03-18
Fix js bundle caching

tomrf1:
Play sets an [etag header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) for assets. But it's not working well for our js bundle - the etag value doesn't tend to change between releases, which means users have to hard-refresh to see updates.

The Play etag behaviour is specified [here](https://www.playframework.com/documentation/2.8.x/AssetsOverview#Etag-support). I'm not sure why the fallback behaviour doesn't work (involving file date), but this PR adds [sbt-digest](https://github.com/sbt/sbt-digest), which gives Play a hash for its etags.

### Is it working?
In CODE we can see that Play is now serving a version of the js bundle with the etag hash in the filename:
Screen Shot 2022-03-18 at 09 37 29

Screen Shot 2022-03-18 at 09 42 13


johnduffell (commented):
nice work 👍

johnduffell:
See https://github.com/sbt/sbt/issues/6235 for why the timestamp might not be working

PR #299, opened 2022-03-22
Functionality to clone test variants

KaliedaRik:
## What does this change?

This PR will add functionality to allow user to clone an existing test variant, rather than having to create a new variant and copy text between fields in different variants.

Work in progress. Details will be added in due course

### DIFF
--- a/public/src/components/channelManagement/variantCloneButton.tsx
+++ b/public/src/components/channelManagement/variantCloneButton.tsx
@@ -1,62 +1,8 @@
 import React from 'react';
-import { useForm } from 'react-hook-form';
-import {
-  Button,
-  Dialog,
-  DialogActions,
-  DialogContent,
-  DialogTitle,
-  IconButton,
-  makeStyles,
-  TextField,
-  Theme,
-} from '@material-ui/core';
-
-import CloseIcon from '@material-ui/icons/Close';
-
+import { Button } from '@material-ui/core';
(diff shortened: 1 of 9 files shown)


public/src/components/channelManagement/variantCloneButton.tsx
```diff
@@ -0,0 +1,133 @@
...
+import useOpenable from '../../hooks/useOpenable';
+
+import {
+  EMPTY_ERROR_HELPER_TEXT,
+  INVALID_CHARACTERS_ERROR_HELPER_TEXT,
+  VALID_CHARACTERS_REGEX,
+  createDuplicateValidator,
+} from './helpers/validation';
+
+const useStyles = makeStyles(({ spacing }: Theme) => ({
```
tomrf1, line 28: this file copies a lot of css and logic from `testNewVariantButton.tsx`. We can refactor slightly to share a `createVariantDialog.tsx`, like we do for tests [here](https://github.com/guardian/support-admin-console/blob/main/public/src/components/channelManagement/createTestDialog.tsx#L69).
I've pushed a branch to show how this could be done: https://github.com/guardian/support-admin-console/pull/301

KaliedaRik: Done

tomrf1 (approved):
👍 great feature!

PR #303, opened 2022-03-29
Migrate uses of withStyles to makeStyles

tomrf1:
There are 2 ways to do css-in-js with MUI:
- higher-order component API (`withStyles`)
- hook API (`makeStyles`)

Currently the codebase uses a mix of the two, which is confusing.
We now prefer `makeStyles`. This is because `withStyles` creates more complexity - we have to wrap our components, and this can confuse the TS compiler when type parameters are involved.

This PR replaces all instances of `withStyles` with `makeStyles`.

### DIFF
--- a/public/src/utils/requests.tsx
+++ b/public/src/utils/requests.tsx
@@ -71,7 +71,7 @@ export function fetchSupportFrontendSettings(
 
 export function saveSupportFrontendSettings(
   settingsType: SupportFrontendSettingsType,
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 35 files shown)


public/src/utils/requests.tsx
```diff
@@ -71,7 +71,7 @@ export function fetchSupportFrontendSettings(
 
 export function saveSupportFrontendSettings(
   settingsType: SupportFrontendSettingsType,
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/explicit-module-boundary-types
```
tomrf1, line 74: just silencing some old warnings

PR #304, opened 2022-03-31
Use radios for supporter status and device type

tomrf1:
Before - we have a strange UX for these, involving checkboxes:
Screen Shot 2022-03-31 at 14 59 32

After - simpler UX with radio buttons:
Screen Shot 2022-03-31 at 15 00 46

I've deleted the two old components:
- testEditorTargetDeviceType.tsx
- testEditorTargetSupporterStatusSelector.tsx

And created a generic component (`TypedRadioGroup`) for radio buttons, for use where we have a union of strings (e.g. `type DeviceType = 'Mobile' | 'Desktop' | 'All'`).
There are probably more cases in this repo where it can be used but I'll just do these 2 for now. I've deleted much more code than I've added


KaliedaRik (approved):
I really like this UX improvement!

PR #305, opened 2022-04-05
Populate new test modal fields with copied test data

KaliedaRik:
## What does this change?
This is a small PR which will pre-populate the 'new test' modal's form fields with data when the user copies an existing test to create a new one.

**Screenshots: creating a new test with default values**
Screenshot 2022-04-05 at 14 30 23

Screenshot 2022-04-05 at 14 30 38

**Screenshots: creating a test from an existing test's data**
Screenshot 2022-04-05 at 14 30 56

Screenshot 2022-04-05 at 14 31 10


tomrf1 (approved):
great!

PR #307, opened 2022-04-13
Do not allow html on AMP/Apple News

tomrf1:
The Rich Text Editor allows arbitrary html. If a user pastes text from a google doc then it may include html.
AMP + Apple News do not support html.
This PR adds a `allowHtml` flag to the RTE to prevent html accidentally being published on these platforms.

Separately, we should explore how we can prevent copying of markup from google docs.

### DIFF
--- a/public/src/components/channelManagement/richTextEditor/richTextEditor.tsx
+++ b/public/src/components/channelManagement/richTextEditor/richTextEditor.tsx
@@ -36,6 +36,7 @@ interface RichTextEditorProps<T> {
   error: boolean;
   updateCopy: (item?: T) => void;
   copyData?: T;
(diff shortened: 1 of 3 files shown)


public/src/components/channelManagement/richTextEditor/richTextEditor.tsx
```diff
@@ -291,6 +292,25 @@ const getRteCopyLength = (copy: string[]): number => {
   return paragraphsCheck.length;
 };
 
+const paragraphsToArray = (html: string): string[] => {
```
tomrf1, line 295: refactored from below

KaliedaRik (approved):
Seems sensible

PR #308, opened 2022-04-13
Add cdk

tomrf1:
The first step in using cdk is to add it to the project, and just point it at our existing cloudformation.yaml.
Later, we can redefine our resources in TS using the cdk api, and remove the cloudformation.yaml.


cloudformation.yaml
```diff
@@ -203,8 +203,6 @@ Resources:
       Name: !Sub ${Stack}-${Stage}-${App}
       Port: 9000
       Protocol: HTTP
-      VpcId:
-        Ref: VpcId
```
tomrf1, line 207: this is the only change. Here `VpcId` appeared twice (by mistake), which now produces an error:
`Map keys must be unique; "VpcId" is repeated`

jest.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
tomrf1, line 1: the jest config in package.json wasn't working properly, I've moved it here and fixed the path condition

.eslintignore
```diff
@@ -0,0 +1,12 @@
+cdk/
```
tomrf1, line 1: eslint was previously using the .gitignore file, but we now need to exclude the cdk directory from the main linter, so I've added a .eslintignore

cloudformation.yaml
TomPretty: good to fix this!

TomPretty (commented):
Looking good! I think we'll need to update https://github.com/guardian/support-admin-console/blob/main/build.sbt#L41 to use the new cloudformation file produced by cdk. Given it produces json it might be worth renaming to `cfn.json` and also updating riff-raff: https://github.com/guardian/support-admin-console/blob/main/conf/riff-raff.yaml#L8

tomrf1:
> Looking good! I think we'll need to update https://github.com/guardian/support-admin-console/blob/main/build.sbt#L41 to use the new cloudformation file produced by cdk. Given it produces json it might be worth renaming to `cfn.json` and also updating riff-raff: https://github.com/guardian/support-admin-console/blob/main/conf/riff-raff.yaml#L8

Good spot, thanks!

TomPretty (approved):
LGTM! Thanks for picking this one up in marketing tools!

PR #309, opened 2022-04-14
Remove html from pasted text

tomrf1:
Our users often paste copy from Google Docs, which may inadvertently include markup.
This PR adds a Remirror Extension to add a prosemirror plugin which removes html from pasted text.


KaliedaRik (approved):
Happy to approve this

PR #312, opened 2022-04-20
Add a new copy template for digisub price data

KaliedaRik:
## What does this change?
Marketing colleagues have identified a requirement to add a set of %%PRICE_XYZ%% string templates to the system. These will be used while creating copy for Banner (and Epic) test variants in the Reader Revenue Control Panel.

This PR adds functionality to validate %%PRICE_XYZ%% strings in RRCP, replacing them with price strings for our various products (monthly and annual) which will be supplied to components as props supplied by the SDC/Support-Frontend servers. A subsequent PR will add a UI to the Rich Text Editor menu to make it easier for Marketing colleagues to add the templates to their copy.

This PR will need to be deployed alongside a sibling PR in the support-dotcom-components repo, and a PR for support-frontend, to add functionality for SDR to poll/cache the frontend on a regular basis to obtain correct pricing information for each product, in each region at the current time.

support-frontend PR: https://github.com/guardian/support-frontend/pull/3653

support-dotcom-components PR: https://github.com/guardian/support-dotcom-components/pull/652


tomrf1 (approved):
nice! Best to merge after the SDC one goes live

public/src/components/channelManagement/helpers/validation.tsx
```diff
@@ -100,11 +100,23 @@ export const createDuplicateValidator = (
 export const CURRENCY_TEMPLATE = '%%CURRENCY_SYMBOL%%';
 export const COUNTRY_NAME_TEMPLATE = '%%COUNTRY_NAME%%';
 export const ARTICLE_COUNT_TEMPLATE = '%%ARTICLE_COUNT%%';
+export const PRICE_DIGISUB_MONTHLY = '%%PRICE_DIGISUB_MONTHLY%%';
+export const PRICE_DIGISUB_ANNUAL = '%%PRICE_DIGISUB_ANNUAL%%';
+export const PRICE_GUARDIANWEEKLY_MONTHLY = '%%PRICE_GUARDIANWEEKLY_MONTHLY%%';
+export const PRICE_GUARDIANWEEKLY_ANNUAL = '%%PRICE_GUARDIANWEEKLY_ANNUAL%%';
 
 const VALID_TEMPLATES = {
```
tomrf1, line 108: Beyond the scope of this PR, but I think we need a refactor of template rules.
These new templates are currently specific to DOTCOM + banner, but this PR enables them on epic/header.
And really the RTE should only display the allowed templates in the UI (e.g. AMP currently shows the article count template even though it's not allowed).
There is a card for this

KaliedaRik: > but this PR enables them on epic/header

I've removed the `prices` attribute from both header and epic models in SDC which (I hope) will prevent those channels displaying a message should the test include a price template - but I haven't tested this with the recent changes.

For SAC (this PR), I'm thinking maybe we can make the RTE fields more aware of their environment? For instance, when we build the UI for these templates, we only want the menu/buttons to display in Banner forms. Similarly, we want to suppress the `bold` button in the highlighted copy field (as bold styling gets added automatically by SDC).

tomrf1: Yes, showing the correct templates for the platform/channel is what I meant. I'm ok with this being done in a separate PR though

PR #313, opened 2022-04-20
Remove RTE

tomrf1:
The RTE has introduced a bug that prevents multiple variants from being edited.
It's a very strange bug and we need time to investigate.
For now, this PR removes all uses of the RTE. It instead uses the MUI `TextField` component


KaliedaRik (approved):
It breaks my heart, but I approve this temporary regression

PR #314, opened 2022-04-20
Only render the selected variant

tomrf1:
This fixes a bug that was introduced with the RTE.
The bug:
1. edit any RTE field in variantA
2. edit any RTE field in variantB
3. changes to variantA are overwritten

If RTE fields are involved, the `onVariantChange` function ([here](https://github.com/guardian/support-admin-console/blob/main/public/src/components/channelManagement/epicTests/epicTestEditor.tsx#L109)) ends up with a stale version of `test.variants`, despite a re-render. This doesn't happen if RTE fields are not involved, so there's something very strange going on.

The quick fix in this PR is to only render the VariantEditor for the currently selected variant. This means the UI is _slightly_ slower when going between variants, but it's barely noticable.


KaliedaRik (approved):
LGTM!

PR #316, opened 2022-04-21
Improve html validation on AMP/Apple News

tomrf1:
A previous change to pass an `allowHtml` prop into the RTE prevents users from adding styling, by using `getText` instead of `getHTML`. But I've noticed it doesn't actually prevent them from typing in their own html.

This PR:
1. adds a validator to check for html if the epic is on AMP/AN
2. Hides the bold/italics buttons on the RTE if `!allowHtml` (to avoid confusion)
3. Tidies up validation.ts by removing unused code
4. Makes the return type of the validator functions consistant.  They now all return `string | undefined`, where a `string` contains an error message and `undefined` means no error. ([expected type](https://github.com/react-hook-form/react-hook-form/blob/master/src/types/validator.ts#L18))


KaliedaRik (approved):
LGTM

PR #317, opened 2022-04-21
Add dynamodb tables

tomrf1:
This adds new dynamodb tables for channel tests.
They're in the same stack as the support-admin-console app. I did consider having a separate stack for the tables, but decided there isn't a good reason for it as the `removalPolicy: 'retain'` will prevent accidental deletions during cloudformation updates.

This PR creates 3 dynamodb tables, for epic/banner/header. It adds an index to each for querying by `status` - most of the time we'll be filtering out archived tests.

The read/write capacity (which affects the cost) is a bit of a guess at the moment - we can tune this later.

Note that because dynamodb is schema-less, rows will only require a `name` and `status`, to satisfy the main `partitionKey` and the index `partitionKey`. The backend will validate the json against the model, as it currently does. This makes migrating `json-in-S3` to `json-in-dynamodb` very simple, and we can gradually make use of dynamo's features to improve the app.


build.sbt
```diff
@@ -38,7 +38,10 @@ riffRaffManifestProjectName := "support:admin-console"
 riffRaffPackageName := "admin-console"
 riffRaffUploadArtifactBucket := Option("riffraff-artifact")
 riffRaffUploadManifestBucket := Option("riffraff-builds")
-riffRaffArtifactResources += (file("cdk/cdk.out/AdminConsole-PROD.template.json"), "cfn/AdminConsole-PROD.template.json")
-riffRaffArtifactResources += (file("cdk/cdk.out/AdminConsole-CODE.template.json"), "cfn/AdminConsole-CODE.template.json")
-
+riffRaffArtifactResources ++= Seq(
```
tomrf1, line 41: this gets the cloudformation templates into the right place during the build, ready for riffraff to pick them up

cdk/bin/cdk.ts
```diff
@@ -1,7 +1,10 @@
 import 'source-map-support/register';
 import { App } from 'aws-cdk-lib';
 import { AdminConsole } from '../lib/admin-console';
+import { AdminConsoleDynamo } from '../lib/admin-console-dynamo';
 
 const app = new App();
 new AdminConsole(app, 'AdminConsole-CODE', { stack: 'support', stage: 'CODE' });
 new AdminConsole(app, 'AdminConsole-PROD', { stack: 'support', stage: 'PROD' });
+new AdminConsoleDynamo(app, 'AdminConsoleDynamo-CODE', { stack: 'support', stage: 'CODE' });
+new AdminConsoleDynamo(app, 'AdminConsoleDynamo-PROD', { stack: 'support', stage: 'PROD' });
```
tomrf1, line 10: This adds the new stack, separate from the existing one that contains the loadbalancer, autoscaling group etc

cdk/lib/__snapshots__/admin-console-dynamo.test.ts.snap
```diff
@@ -0,0 +1,200 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
tomrf1, line 1: autogenerated

cdk/lib/admin-console-dynamo.ts
```diff
@@ -0,0 +1,46 @@
...
+    const buildTable = (channel: string) => {
+      const id = `${channel[0].toUpperCase() + channel.slice(1)}TestsDynamoTable`;
+
+      const table = new Table(this, id, {
+        tableName: `${channel}-tests-${this.stage}`,
+        removalPolicy: RemovalPolicy.RETAIN,
+        readCapacity: 4,
+        writeCapacity: 4,
+        partitionKey: {
+          name: 'name',
```
tomrf1, line 20: this makes the test name the primary key - we already require each test to have a unique name

cdk/bin/cdk.ts
TomPretty: so the reason to have this as a separate stack is to decouple the updates between the two, given the dynamo stack will almost never change?

cdk/lib/admin-console-dynamo.ts
```diff
@@ -0,0 +1,46 @@
...
+import type { App, CfnElement } from 'aws-cdk-lib';
+import { RemovalPolicy } from 'aws-cdk-lib';
+import { AttributeType, ProjectionType, Table } from 'aws-cdk-lib/aws-dynamodb';
+
+export class AdminConsoleDynamo extends GuStack {
+  constructor(scope: App, id: string, props: GuStackProps) {
+    super(scope, id, props);
+
+    const buildTable = (channel: string) => {
+      const id = `${channel[0].toUpperCase() + channel.slice(1)}TestsDynamoTable`;
```
TomPretty, line 12: I saw this and thought js must have a capitalise function - but it doesn't!

cdk/bin/cdk.ts
tomrf1: It's more my instinct to keep the persistence separate from the app...but I'm not sure it's necessary. The `removalPolicy: RemovalPolicy.RETAIN` part should ensure the tables are never deleted by an accidental cloudformation update.
I'm going to go away and get opinions on the devex channel before going ahead with this!

tomrf1: I ended up doing it all in one stack

cdk/lib/__snapshots__/admin-console.test.ts.snap
```diff
@@ -158,6 +158,70 @@ Object {
       },
       "Type": "AWS::AutoScaling::LaunchConfiguration",
     },
+    "BannerTestsDynamoTable": Object {
```
tomrf1, line 161: this file is generated when you run `yarn test -u`

cdk/lib/admin-console.ts
```diff
@@ -5,25 +5,86 @@ import { GuStack } from '@guardian/cdk/lib/constructs/core';
...
     const { domainName } = props;
 
     const app = 'admin-console';
 
+    const dynamoTables = [
+      this.buildTestsTable('header'),
+      this.buildTestsTable('epic'),
+      this.buildTestsTable('banner'),
+    ];
+    const dynamoPolicies = this.buildDynamoPolicies(dynamoTables);
```
TomPretty, line 86: this is so much nicer than having to write things out manually in cloudformation!

TomPretty (approved):
LGTM 👍

cdk/lib/admin-console.ts
tomrf1: Agreed!

PR #318, opened 2022-04-22
CDK phase 2: dual stack

tomrf1:
This step creates the new stack, and leaves the old stack in place.
It also creats a CNAME record which we will later use to route traffic to the new stack's loadbalancer.

### DIFF
--- a/cdk/lib/admin-console.test.ts
+++ b/cdk/lib/admin-console.test.ts
@@ -1,15 +1,12 @@
 import { App } from 'aws-cdk-lib';
 import { Template } from 'aws-cdk-lib/assertions';
+import { prodProps } from '../bin/cdk';
 import { AdminConsole } from './admin-console';
 
 describe('The AdminConsole stack', () => {
   it('matches the snapshot', () => {
     const app = new App();
-    const stack = new AdminConsole(app, 'AdminConsole', {
-      stack: 'support',
-      stage: 'TEST',
-      domainName: 'support.code.dev-gutools.co.uk',
-    });
+    const stack = new AdminConsole(app, 'AdminConsole', prodProps);
     const template = Template.fromStack(stack);
     expect(template.toJSON()).toMatchSnapshot();
   });
--- a/cdk/lib/admin-console.test.ts
+++ b/cdk/lib/admin-console.test.ts
@@ -5,7 +5,11 @@ import { AdminConsole } from './admin-console';
 describe('The AdminConsole stack', () => {
   it('matches the snapshot', () => {
     const app = new App();
-    const stack = new AdminConsole(app, 'AdminConsole', { stack: 'support', stage: 'TEST' });
+    const stack = new AdminConsole(app, 'AdminConsole', {
+      stack: 'support',
+      stage: 'TEST',
+      domainName: 'support.code.dev-gutools.co.uk',
+    });
     const template = Template.fromStack(stack);
     expect(template.toJSON()).toMatchSnapshot();
   });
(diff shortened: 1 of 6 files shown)


cdk/lib/admin-console.ts
```diff
@@ -1,15 +1,96 @@
...
+      monitoringConfiguration: {
+        noMonitoring: true,
+      },
+      userData,
+      roleConfiguration: {
+        additionalPolicies: policies,
+      },
+      scaling: { minimumInstances: 1, maximumInstances: 2 },
+      instanceType: InstanceType.of(InstanceClass.T4G, InstanceSize.MICRO),
+      withoutImdsv2: true,
```
tomrf1, line 74: necessary for cloudwatch logging, see https://github.com/guardian/support-frontend/commit/76a015bc2461af10295c7ddb3dd0651ee92c23ae

cdk/lib/admin-console.test.ts
```diff
@@ -5,7 +5,11 @@ import { AdminConsole } from './admin-console';
 describe('The AdminConsole stack', () => {
   it('matches the snapshot', () => {
     const app = new App();
-    const stack = new AdminConsole(app, 'AdminConsole', { stack: 'support', stage: 'TEST' });
+    const stack = new AdminConsole(app, 'AdminConsole', {
+      stack: 'support',
```
jacobwinch, line 9: Minor thing (and personal preference really): I have started pulling my [`PROD` props into a const](https://github.com/guardian/amigo/blob/3f0bb6b1c75d2029eeddac0d6fcc27ee850c75e8/cdk/bin/cdk.ts#L20-L24) and [passing these into the test](https://github.com/guardian/amigo/blob/3f0bb6b1c75d2029eeddac0d6fcc27ee850c75e8/cdk/lib/amigo.test.ts#L9) as well as the [stack instantiation](https://github.com/guardian/amigo/blob/3f0bb6b1c75d2029eeddac0d6fcc27ee850c75e8/cdk/bin/cdk.ts#L26).

This helps to keep your snapshot as close as possible to the template that you will actually upload (i.e. the thing that you really care about not breaking!) - perhaps there are a few subtle bugs that can be caught this way as the number of stage-specific props grows.

cdk/lib/admin-console.ts
```diff
@@ -1,15 +1,96 @@
...
+      }),
+      new GuPutS3ObjectsPolicy(this, 'PublicSettingsBucketPut', {
+        bucketName: 'gu-contributions-public',
+        paths: [`epic/${this.stage}/*`, `banner/${this.stage}/*`, `header/${this.stage}/*`],
+      }),
+    ];
+
+    const ec2App = new GuEc2App(this, {
+      applicationPort: 9000,
+      app: 'admin-console',
```
jacobwinch, line 60: Minor: I think you can use your `app` const (from line 28) here

cdk/lib/admin-console.ts
```diff
@@ -1,15 +1,96 @@
...
+      access: { scope: AccessScope.PUBLIC },
```
jacobwinch, line 61: This feels a little odd for an internal tool, but it looks like the old load balancer is also configured in this way?

cdk/lib/admin-console.ts
```diff
@@ -1,15 +1,96 @@
...
+    const template = new CfnInclude(this, 'YamlTemplate', {
       templateFile: yamlTemplateFilePath,
     });
+
+    // TODO Migrate to new LB
+    const oldElb = template.getResource('ElasticLoadBalancer') as CfnLoadBalancer;
+    new GuCname(this, 'cname', {
+      app,
+      domainName,
+      ttl: Duration.minutes(60),
```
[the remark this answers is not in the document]
tomrf1: yep that's right!

cdk/lib/admin-console.ts
tomrf1: yes it actually used to be restricted, but was changed [here](https://github.com/guardian/support-admin-console/pull/118)!

PR #320, opened 2022-04-27
Route traffic to new stack

tomrf1:
Also, because this is the first time I've been able to test the app in CODE, I discovered it was missing the `s3:PutObjectAcl` permission on a bucket, so I've added this.

The final step after this is live will be to delete the old stack.

### DIFF
--- a/cdk/lib/admin-console.ts
+++ b/cdk/lib/admin-console.ts
@@ -48,9 +48,13 @@ export class AdminConsole extends GuStack {
         bucketName: 'support-admin-console',
         paths: [`${this.stage}/*`],
       }),
-      new GuPutS3ObjectsPolicy(this, 'PublicSettingsBucketPut', {
-        bucketName: 'gu-contributions-public',
-        paths: [`epic/${this.stage}/*`, `banner/${this.stage}/*`, `header/${this.stage}/*`],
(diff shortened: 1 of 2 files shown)


cdk/lib/admin-console.ts
```diff
@@ -80,17 +83,16 @@ export class AdminConsole extends GuStack {
...
     // TODO Legacy cloudformation stack - to be removed after migration to cdk is complete
     const yamlTemplateFilePath = join(__dirname, '../..', 'cloudformation.yaml');
-    const template = new CfnInclude(this, 'YamlTemplate', {
+    new CfnInclude(this, 'YamlTemplate', {
       templateFile: yamlTemplateFilePath,
     });
 
-    // TODO Migrate to new LB
-    const oldElb = template.getResource('ElasticLoadBalancer') as CfnLoadBalancer;
+    // TODO lower ttl after migration
```
jacobwinch, line 90: Minor: ```suggestion
    // TODO increase ttl again after migration
```

Comment (jacobwinch) [inline cdk/lib/admin-console.ts:90]:
```suggestion
    // TODO increase ttl again after migration
```

jacobwinch (approved):

cdk/lib/admin-console.ts
tomrf1: good spot!

PR #321, opened 2022-04-27
Complete CDK migration

tomrf1:
Following https://github.com/guardian/cdk/blob/main/docs/migration-guide-ec2.md#stage-3-cleanup

1. Removes old cloudformation, deleting the old stack
2. Removes `gu:riffraff:new-asg` tag
3. Remove the `asgMigrationInProgress` parameter from the riff-raff.yaml
4. Adds `amiParameter` to the riff-raff.yaml

This should be merged when we're confident the new stack is ok and no traffic is going to the old one


jacobwinch (approved):
🎉

PR #324, opened 2022-04-29
Make epic heading optional

tomrf1:
The heading field is optional in the model, and these days it's mostly left empty. Currently users have to enter a space because the form requires it to be non-empty.

Tested with AMP and Apple News to ensure they're happy with `undefined` heading


public/src/components/channelManagement/epicTests/utils/defaults.ts
```diff
@@ -21,7 +21,6 @@ export const DEFAULT_SECONDARY_CTA: Cta = {
 
 const DEV_AND_CODE_DEFAULT_VARIANT: EpicVariant = {
   name: 'CONTROL',
-  heading: "Since you're here ...",
```
tomrf1, line 24: this is pre-populated in CODE/DEV, but it's best to keep it closer to how the tool is used in PROD

KaliedaRik (approved):
Approved

PR #325, opened 2022-05-04
Add new pricing templates to the Rich Text Editor

KaliedaRik:
## What does this change?
We've added new pricing copy templates to SDC and SAC repos - see PR #312 for further info.

This PR adds UX to the Rich Text Editor to make it easy for users to add the new templates to their copy

**Screenshots:**
Screenshot 2022-05-04 at 16 17 48
Screenshot 2022-05-04 at 16 18 01
Screenshot 2022-05-04 at 16 18 19
Screenshot 2022-05-04 at 16 18 34

### DIFF
--- a/public/src/components/channelManagement/richTextEditor/richTextEditor.tsx
+++ b/public/src/components/channelManagement/richTextEditor/richTextEditor.tsx
@@ -341,22 +341,13 @@ const RichTextMenu: React.FC<RichTextMenuProps> = ({
               <span className={classes.remirrorButtonSpacer}>&nbsp;</span>
             </>
           )}
-          <button
-            className="remirror-button"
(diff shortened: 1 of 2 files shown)


tomrf1:
A note about validation -
Currently the RRCP allows users to enter price templates on epic + header on dotcom. And with this PR, price templates will show in the RTE toolbar. SDC does not currently give pricing data to epic + header, and so if a price template is set on these channels then it will fail to render (the user will see nothing).

In a follow-up PR we should make validation and the RTE toolbar smarter:
- validation should only allow price templates on the banner
- RTE should only show price template buttons on the banner tool
- RTE should not show ARTICLE_COUNT button for AMP epic
- ...etc

tomrf1 (approved):
nice simple solution 👍

PR #326, opened 2022-05-05
Single dynamo table for channel tests

tomrf1:
In the 1st PR I added dynamodb tables for epic, banner, and header. In the world of relational databases this feels a normal thing to do.
I then thought about it a bit harder and realised that with dynamo, a table per channel does not make sense.
Firstly, we'd need a lot of tables (apple news epics, liveblog epics, banner1, banner2...).
Secondly, with dynamo there are typically no benefits to defining lots of tables (no joins/normalisation).

With this PR we have just one table: `channel-tests`.
It has a composite primary key:
- the partition key is `channel` (e.g. "epic", "banner1", "banner2"...)
- the sort key is `name`

This means:
1. The test name is still unique within a channel.
2. Querying for tests is efficient because the data is partitioned by channel. Typically SDC/SAC query for tests in a single channel. (Dynamodb uses the partition key value as the input to a hash function).
3. In the future we can add secondary indexes for new use cases, for example for querying by `campaign` across all channels.

See [this article](https://aws.amazon.com/blogs/database/choosing-the-right-dynamodb-partition-key/) for an explanation of dynamodb primary keys.

### DIFF
--- a/cdk/lib/__snapshots__/admin-console.test.ts.snap
+++ b/cdk/lib/__snapshots__/admin-console.test.ts.snap
@@ -158,70 +158,6 @@ Object {
       },
       "Type": "AWS::AutoScaling::LaunchConfiguration",
     },
-    "BannerTestsDynamoTable": Object {
-      "DeletionPolicy": "Retain",
-      "Properties": Object {
-        "AttributeDefinitions": Array [
-          Object {
-            "AttributeName": "name",
-            "AttributeType": "S",
-          },
-          Object {
-            "AttributeName": "status",
-            "AttributeType": "S",
-          },
-        ],
-        "GlobalSecondaryIndexes": Array [
-          Object {
-            "IndexName": "status-index",
-            "KeySchema": Array [
-              Object {
-                "AttributeName": "status",
-                "KeyType": "HASH",
-              },
-            ],
-            "Projection": Object {
-              "ProjectionType": "ALL",
-            },
-            "ProvisionedThroughput": Object {
-              "ReadCapacityUnits": 4,
-              "WriteCapacityUnits": 4,
-            },
-          },
-        ],
-        "KeySchema": Array [
-          Object {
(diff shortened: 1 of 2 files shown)


cdk/lib/__snapshots__/admin-console.test.ts.snap
```diff
@@ -158,49 +158,15 @@ Object {
       },
       "Type": "AWS::AutoScaling::LaunchConfiguration",
     },
-    "BannerTestsDynamoTable": Object {
```
tomrf1, line 161: this file is autogenerated

cdk/lib/admin-console.ts
```diff
@@ -12,39 +12,31 @@ import {
...
       writeCapacity: 4,
       partitionKey: {
-        name: 'name',
+        name: 'channel',
         type: AttributeType.STRING,
       },
-    });
-
-    // Add an index for querying by status (LIVE/DRAFT/ARCHIVED)
-    table.addGlobalSecondaryIndex({
```
tomrf1, line 39: for now let's try without any indexes

rupertbates (approved):
One suggestion about the table name

cdk/lib/admin-console.ts
TomPretty: yeah makes sense to add an index alongside building the feature that will use it 👍

TomPretty (approved):
Great! I guess you'll need to deploy the cfn change manually as it'll delete the existing tables?

tomrf1:
> > Great! I guess you'll need to deploy the cfn change manually as it'll delete the existing tables?
> 
> Yes, I'm currently in cloudformation hell trying to do this!

For the record, I had to manually edit the *stack policy* to allow deletes on `AWS::DynamoDB::Table`, as well as change the `removalPolicy` on the tables themselves

PR #327, opened 2022-05-06
Dynamodb service

tomrf1:
We have a new dynamo table: https://github.com/guardian/support-admin-console/pull/326

This is step 1 of the migration plan. The plan is:
1. Migrate SAC backend to write to dynamo and s3 at the same time, but still read from s3. No change to the client. Data is now automatically duplicated between s3/dynamo.
2. Migrate SDC to query dynamo instead of using s3.
3. Migrate SAC client + server to stop using s3, and enable test-level locks.

**Changes in this PR:**
- Introduces the `DynamoChannelTests` class, with basic functionality for writing tests to the table.
- The existing controller for channel tests (`LockableS3ObjectController`) now takes a `dynamoWrite` function, and calls this immediately after writing to s3.
- When writing to dynamo it now includes a `channel` field (e.g. "Epic", "Banner1") - this is the partition key of the table and will be used for querying.
- It also includes a new `priority` field, for test ordering. This is based on the ordering of the array sent by the client.
- I've also added some optional fields for use later on: `status`, `lockStatus`. These aren't used yet.


app/models/LockStatus.scala
```diff
@@ -0,0 +1,10 @@
+package models
+
+import java.time.OffsetDateTime
+
+case class LockStatus(locked: Boolean, email: Option[String], timestamp: Option[OffsetDateTime])
```
tomrf1, line 5: contents of this file is copied from `LockableS3ObjectController.scala` for sharing

app/utils/Circe.scala
```diff
@@ -1,11 +1,67 @@
...
 
 object Circe {
   def decodeStringAndCollect[T](pf: PartialFunction[String,T]): Decoder[T] = Decoder.decodeString.emap { s =>
     pf.lift(s)
       .map(Right.apply)
       .getOrElse(Left(s"Unexpected value: $s"))
   }
+
+  private val printer = Printer.spaces2.copy(dropNullValues = true)
+  def noNulls(json: Json): String = printer.print(json)
```
tomrf1, line 17: copied from `utils/Circe/scala`

app/controllers/AmountsController.scala
```diff
@@ -4,10 +4,10 @@ import com.gu.googleauth.AuthAction
 import models.ConfiguredAmounts
 import models.ConfiguredAmounts._
 import play.api.mvc.{AnyContent, ControllerComponents}
-import zio.DefaultRuntime
+import zio.ZEnv
```
tomrf1, line 7: In this PR I've updated the zio version, which unfortunately changes the `Runtime` definition slightly. So there are lots of import/type changes like this

app/services/DynamoChannelTests.scala
```diff
@@ -0,0 +1,128 @@
...
+    * the stream when the list of batches is empty.
+    */
+  private val BATCH_SIZE = 25
+  def putAllBatched(writeRequests: List[WriteRequest]): ZIO[ZEnv, DynamoPutError, Unit] = {
+    val batches = writeRequests.grouped(BATCH_SIZE).toList
+    ZStream(())
+      .forever
+      .fixed(2.seconds) // wait 2 seconds between batches
+      .timeoutError(DynamoPutError(new Throwable("Timed out writing batches to dynamodb")))(1.minute)
+      .foldWhileM(batches)(_.nonEmpty) {
```
TomPretty, line 87: This is v niceee

app/models/ChannelTest.scala
```diff
@@ -0,0 +1,48 @@
...
+  case object Banner1 extends Channel
+  case object Banner2 extends Channel
+  case object Header extends Channel
+
+  implicit val customConfig: Configuration = Configuration.default.withDefaults
+  implicit val statusEncoder = deriveEnumerationEncoder[Channel]
+  implicit val statusDecoder = deriveEnumerationDecoder[Channel]
+}
+
+trait ChannelTest[T] {
```
tomrf1, line 35: The type parameter here is a bit odd. `T` will be the same type, e.g. `case class BannerTest(...) extends ChannelTest[BannerTest]`.
It's necessary for the `withChannel` and `withPriority` methods, which need to specifically return `T` rather than just `ChannelTest`

app/utils/Circe.scala
```diff
@@ -1,11 +1,67 @@
...
+
+  // Converts Circe Json to Dynamodb Attributes
+  def jsonToDynamo(json: Json): AttributeValue =
+    json.deepDropNullValues.fold(
+      jsonNull = AttributeValue.builder().nul(true).build,
```
TomPretty, line 22: feels like a very amazon-ey move to go for `nul` instead of `null`

tomrf1: yes I'm not really a fan of the method names in the sdk!

app/models/ChannelTest.scala
TomPretty: It's a shame we can't share the implementation of the `withChannel` and `withPriority` methods - but I can't think of a way to do that!

tomrf1: Yeah - each implementation just calls `.copy(...)`. But `copy` is only available on case classes, and `ChannelTest` has to be a trait

app/models/ChannelTest.scala
```diff
@@ -0,0 +1,48 @@
...
+  val name: String
+  val channel: Option[Channel] // optional only for the migration
+  val status: Option[Status]
+  val lockStatus: Option[LockStatus]
+  val priority: Option[Int]
```
TomPretty, line 40: I've been having a think if we could avoid using an `Int` here - it just means in the worst case we'll actually have to update every single test when we reorder the list. 

I've been trying to google a specific concept, but I can't seem to find the technical term for it, so I'll try and explain it!

If we had `priority: String`. Then if we had one test with `priority = 'b'` and one with `priority = 'c'`, then we can insert a test between those with e.g `priority = 'bm'`. 

That way we can always do an `O(1)` insert. We would have to be smart generating the value of order to keep the strings as short as possible (I've used a python library for this exact thing in the past but I can't remember what it was called).

As an extra bonus, we could actually use this as the sort key in dynamo and then we'd be able to query the table in priority order. (You could actually do this now even if we kept priority as an `Int`).

tomrf1: This is an interesting idea!
The `priority` is the least satisfying part of this solution, compared with the array in S3.
The plan was:
1. decouple test ordering and test editing in the UI. Updating the ordering of the list of tests is a separate action.
2. client submits an array of test names, and the server updates just the `priority` field of every live/draft test in dynamo. It doesn't matter what the old priorities were, we just use the indices from the array. 

But I'll have a think about your suggestion.

Regarding the sort key - I think we still want `name` as the sort key in the primary key, to keep it unique per channel. But we can define a secondary index `priority` as the sort key.

TomPretty: Ahh yeah completely forgot about the `name` part haha - yeah definitely want to keep that as the primary sort key! Could definitely add a secondary index to query in priority order though 👍

TomPretty: Yeah makes sense!

tomrf1:
I've just pushed a commit to add a `replaceChannelTests` function.
This is because we need to handle deletes in dynamodb as well.
It does this by fetching all existing tests in dynamo for the channel, deleting any which aren't in the updated list of tests, then batch writing the updated list.

PR #328, opened 2022-05-06
add field environment awareness functionality

KaliedaRik:
## What does this change?
This PR adds some limited awareness for each RTE field in a form, so that its menu buttons adapt to that environment.

For example:

- RTE fields in an Epic test form will not display prices copy template buttons
- RTE fields in an Apple news Epic form will only display the currency copy template, and hide the bold and italic buttons
- etc

TODO: add screenshots


tomrf1 (approved):
great! I'm very happy that the validation rules flow into the RTE this way

PR #329, opened 2022-05-09
Batch stream

tomrf1:


app/services/DynamoChannelTests.scala
```diff
@@ -47,32 +49,56 @@ class DynamoChannelTests(stage: String) extends StrictLogging {
...
+        .get(tableName)
+        .map(items => items.asScala.toList)
+        .getOrElse(Nil)
 
     }.mapError(error =>
       DynamoPutError(error)
     )
 
-  def getAllTests[T <: ChannelTest[T] : Decoder](channel: Channel): ZIO[Blocking, DynamoGetError, List[T]] =
+  /**
```
tomrf1, line 73: @adamnfish any thoughts on `putAllBatched`? The use of an empty stream feels a bit odd, but it gives me all the features I wanted

app/services/DynamoChannelTests.scala
```diff
@@ -47,32 +49,56 @@ class DynamoChannelTests(stage: String) extends StrictLogging {
...
+          logger.info(s"next batch: $nextBatch")
+          putAll(nextBatch).map {
+            case Nil => remainingBatches
+            case unprocessed => unprocessed :: remainingBatches
+          }
+        case (Nil, _) =>
+          logger.info(s"finished batches")
+          ZIO.succeed(Nil)
+      }
+      .map(_ => ()) // on success, the result value isn't meaningful
```
kelvin-chappell, line 98:
```suggestion
      .unit // on success, the result value isn't meaningful
```

kelvin-chappell (approved):
It's a nice solution.  Quite easy to see what's going on 👍

app/services/DynamoChannelTests.scala
tomrf1: nice!

PR #331, opened 2022-05-12
Show 'Saving...' during save

tomrf1:
Currently in the channel test tools there is no UI feedback after clicking the save button until the save completes.
This PR changes the button to say "Saving..."
Screen Shot 2022-05-12 at 08 47 31

I've implemented this by adding a new `saving` boolean state to the `TestEditor` higher-order component.

Saving tests may get a bit slower during the dynamodb migration while we write all tests to S3 + dynamo.


KaliedaRik (approved):
Nice UX improvement

PR #332, opened 2022-05-13
Alarm for 5XX errors

tomrf1:
Currently we have no alarms.
With this PR, an email alert will be sent any time a 5XX response is returned to the client.
We always return 5XXs when there's a backend error, so this should be enough.

### DIFF
--- a/cdk/lib/__snapshots__/admin-console.test.ts.snap
+++ b/cdk/lib/__snapshots__/admin-console.test.ts.snap
@@ -514,6 +514,112 @@ Object {
       },
       "Type": "AWS::IAM::Policy",
     },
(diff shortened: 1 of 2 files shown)


cdk/lib/__snapshots__/admin-console.test.ts.snap
```diff
@@ -514,6 +514,112 @@ Object {
       },
       "Type": "AWS::IAM::Policy",
     },
+    "High5xxPercentageAlarmAdminconsole274ADE37": Object {
```
tomrf1, line 517: this file is autogenerated when you run `yarn test -u` under `/cdk`

MaelGNM (approved):
Looks good!

PR #336, opened 2022-05-16
Fix epic holdback access

tomrf1:
Test config files are public in s3 (for historical reasons - dotcom used to fetch them directly).
Currently when the Holdback epic config is updated in S3, it removes read access. This doesn't affect SDC (which has full read permissions), but prevents the RRCP from reading the data again.

None of this is relevant once we've finished migrating to dynamodb...


KaliedaRik (approved):
Like the fix!

PR #337, opened 2022-05-17
Bump aws sdk version to 2.17.188

KaliedaRik:
## What does this change?

This PR fixes an 'information exposure' vulnerability identified in the `io.netty:netty-common` library, which is used extensively by the AWS SDK suite of libraries. The recommended fix is to bump the SDK versions to 2.17.188 as they have updated their dependency on io.netty:netty-common to the latest fixed version

## Screenshots

Screenshot 2022-05-17 at 16 05 52

Screenshot 2022-05-17 at 16 37 23

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,7 +6,7 @@ version := "1.0-SNAPSHOT"
 scalaVersion := "2.13.7"
 
 val circeVersion = "0.14.1"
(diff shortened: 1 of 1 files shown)


tomrf1 (approved):
great!

PR #338, opened 2022-05-19
update/fix banner test variant preview

KaliedaRik:
## What does this change?

Existing 'live' preview functionality does not accurately display Banner and Epic test variants to the user - some critical parts of the tests are missing (eg choice cards). Marketing colleagues have to rely on the Web preview (where available) to capture screenshots of their work, which is inefficient and not particularly 'real-time live'.

This PR attempts to fix these issues for the various Banner channel 'live' previews.

### Screenshots

TODO - add screenshots


tomrf1 (approved):
👍

PR #339, opened 2022-05-20
Stop RTE updates overwriting other changes

tomrf1:
In the epic + banner variant editors we use the RTE. Each use of the RTE is wrapped in a react-hook-form `Controller`, and managed by `useForm`.
Currently updates to fields managed by `useForm` will cause other updates to be overwritten. This is because the Controllers each call a callback on update, and that callback has a stale version of the full variant data.

One solution is to somehow force a re-rendering of each Controller whenever the variant updates.
My solution instead adds a `useState` for storing the latest validated fields, and a `useEffect` to combine them with the rest of the variant fields.

I often wonder if react-hook-form is really the best solution for our validation, but that's a decision for another day...


KaliedaRik (approved):
Looks good - and agree we should review the use of react-hook-form going forward

PR #340, opened 2022-05-23
Increase dynamodb capacity

tomrf1:
I've noticed we're slightly exceeding the write capacity of 4 units/second. This is probably because the list of epic tests is huge and we're currently writing the whole lot when a user saves (in the long term this will not be the case).
The AWS sdk does automatically take care of retries if there's any throttling but I think it's good to have a correct capacity set.


KaliedaRik (approved):
All good to go

PR #341, opened 2022-05-24
Use on-demand billing for dynamodb during migration

tomrf1:
We are in the process of migrating from S3 to dynamodb for channel tests data.
While this is happening, we're batch writing updates for all tests in a channel when a user clicks save. In the long term we can just update the test that a user has changed.
Because of this, the writes to dynamo use a lot of capacity units when a user clicks save. This is infrequent, but can use hundreds of units at a time:
Screen Shot 2022-05-23 at 15 17 51

This PR switches to on-demand rather than provisioned capacity. We can go back to provisioned when the migration is over, and we have more predictable writes.

Docs: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.ReadWriteCapacityMode.html

### DIFF
--- a/app/services/DynamoChannelTests.scala
+++ b/app/services/DynamoChannelTests.scala
@@ -6,7 +6,7 @@ import io.circe.syntax._
 import models.{Channel, ChannelTest}
 import DynamoChannelTests._
 import software.amazon.awssdk.services.dynamodb.DynamoDbClient
-import software.amazon.awssdk.services.dynamodb.model.{AttributeValue, BatchWriteItemRequest, DeleteRequest, PutRequest, QueryRequest, WriteRequest}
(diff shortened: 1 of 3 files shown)


app/services/DynamoChannelTests.scala
```diff
@@ -56,6 +56,7 @@ class DynamoChannelTests(stage: String) extends StrictLogging {
         BatchWriteItemRequest
           .builder
           .requestItems(Map(tableName -> writeRequests.asJava).asJava)
+          .returnConsumedCapacity(ReturnConsumedCapacity.TOTAL)
```
tomrf1, line 59: this is just to get it to show in our logs how many units it used during batch writes

PR #342, opened 2022-05-24
Point-in-time backups

tomrf1:
This PR enables continuous backups in PROD only.
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/PointInTimeRecovery.html
The table isn't very write-heavy so I don't expect this to cost much, but I'll keep any eye on this


MaelGNM (approved):
👍

PR #343, opened 2022-05-25
Stop writing channel tests to S3

tomrf1:
We're in the process of migrating from S3 to Dynamodb for channel tests data.
This is step 3.
Currently the `LockableS3ObjectController` reads from S3 and writes to both S3 and Dynamodb.

[SDC is now reading from dynamodb.](https://github.com/guardian/support-dotcom-components/pull/706#event-6675501424)

With this PR we stop writing to S3, and read from dynamodb.
- I've renamed `LockableS3ObjectController` -> `ChannelTestsController`.
- Users still have to lock all tests in a channel when editing, and this still uses a lock file in S3.
- We still batch update all tests when the user clicks save. A follow up change will need to update the UI + backend to enable editing of a single test at a time. At that point we will no longer need the S3 lock files.
- Deletes the `FastlyPurger`, which was still being used, though isn't actually needed anymore


app/models/ChannelTest.scala
```diff
@@ -43,6 +45,8 @@ trait ChannelTest[T] {
   def withPriority(priority: Int): T
 }
 
-trait ChannelTests[T <: ChannelTest[T]] {
-  val tests: List[T]
+case class ChannelTests[T : Decoder : Encoder](tests: List[T])
```
tomrf1, line 48: the client still expects this model in the response, but there's no longer any need to inherit from `ChannelTests` on the backend

app/controllers/ChannelTestsController.scala
```diff
@@ -1,47 +1,42 @@
...
-    * But we still read from S3, which is the source of truth for now.
-    */
-  dynamoWrite: CT => ZIO[ZEnv, DynamoError, Unit],
+  dynamo: DynamoChannelTests
 )(implicit ec: ExecutionContext) extends AbstractController(components) with Circe {
 
   private val lockObjectSettings = S3ObjectSettings(
     bucket = "support-admin-console",
-    key = s"$stage/locks/$name.lock",
+    key = s"$stage/locks/$lockFileName.lock",
```
tomrf1, line 39: we still use S3 for locking the tests for editing, for now

app/controllers/ChannelTestsController.scala
```diff
@@ -71,11 +66,15 @@ abstract class LockableS3ObjectController[CT <: ChannelTests[_] : Decoder : Enco
...
-      S3Json
-        .getFromJson[CT](s3Client)
-        .apply(dataObjectSettings)
-        .map { case VersionedS3Data(value, version) =>
-          Ok(noNulls(LockableS3ObjectResponse(value, version, lockStatus, request.user.email).asJson))
+      dynamo.getAllTests[T](channel)
+        .map { channelTests =>
+          val response = LockableS3ObjectResponse(
+            ChannelTests(channelTests),
+            "versioning-deprecated",
```
tomrf1, line 74: the client still expects a version field, though this isn't used now that we're not using s3. This can be removed later

MaelGNM:
[Update SAC to only write to dynamo, and to read from dynamo](https://trello.com/c/MeQPsid1/336-update-sac-to-only-write-to-dynamo-and-to-read-from-dynamo)

MaelGNM (approved):
It looks good to me, but would be good to have someone with more scala knowledge to review as well

PR #345, opened 2022-06-06
Set status field in the client

tomrf1:
Part of the migration to dynamodb involves adding a `status` field to replace the `isOn` boolean.
This PR changes the client to set both `isOn` and `status`.

I also have a branch with a 'script' for setting the `status` field based on the current value of `isOn`, once this change is live - https://github.com/guardian/support-admin-console/tree/tf-migrate-status

We then need to change SDC to use the `status` field, before removing `isOn` completely


KaliedaRik (approved):
A welcome improvement!

PR #352, opened 2022-06-16
Add campaign name field to tests

MaelGNM:
## What does this change?

This adds a `campaignName` optional string field to banner, epic and header tests.

### DIFF
--- a/public/src/models/banner.ts
+++ b/public/src/models/banner.ts
@@ -52,4 +52,5 @@ export interface BannerTest extends Test {
   articlesViewedSettings?: ArticlesViewedSettings;
   controlProportionSettings?: ControlProportionSettings;
   deviceType?: DeviceType;
+  campaignName?: string;
 }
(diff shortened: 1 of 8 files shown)


public/src/models/banner.ts
```diff
@@ -52,4 +52,5 @@ export interface BannerTest extends Test {
   articlesViewedSettings?: ArticlesViewedSettings;
   controlProportionSettings?: ControlProportionSettings;
   deviceType?: DeviceType;
+  campaignName?: string;
```
tomrf1, line 55: The field should be on the common `Test` type - https://github.com/guardian/support-admin-console/blob/main/public/src/components/channelManagement/helpers/shared.ts#L18
Also - unlike with scala, in TS you don't _have_ to define the field on each subtype as well, though there's nothing wrong with leaving it in.

tomrf1 (approved):
👍

PR #353, opened 2022-06-16
Set campaign field

tomrf1:
When creating/copying a test, the selected `campaignName` is now passed up to the `TestsForm.tsx` and sent to the server.

I've also added some additional validation to prevent users from accidentally adding multiple instances of `__` to a test name. There should only be one, as this is used by the AB tests dashboard to group together tests with a common prefix.

### DIFF
--- a/public/src/components/channelManagement/createTestDialog.tsx
+++ b/public/src/components/channelManagement/createTestDialog.tsx
@@ -70,7 +70,7 @@ interface CreateTestDialogProps {
   sourceNickname?: string | void;
   mode: Mode;
   testNamePrefix?: string; // set if all tests must have the same prefix
-  createTest: (name: string, nickname: string) => void;
+  createTest: (name: string, nickname: string, campaignName?: string) => void;
 }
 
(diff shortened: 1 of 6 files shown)


public/src/components/channelManagement/createTestDialog.tsx
```diff
@@ -178,7 +190,13 @@ const CreateTestDialog: React.FC<CreateTestDialogProps> = ({
               value: VALID_CHARACTERS_REGEX,
               message: INVALID_CHARACTERS_ERROR_HELPER_TEXT,
             },
-            validate: createDuplicateValidator(existingNames, buildPrefix()),
+            validate: name => {
+              const withPrefix = addPrefix(name);
+              return (
+                doubleUnderscoresValidator(withPrefix) ??
+                createDuplicateValidator(existingNames)(withPrefix)
```
tomrf1, line 197: these 2 validators are chained together

KaliedaRik (approved):
We discussed in the Team Chat. Happy to approve.

PR #355, opened 2022-06-20
Endpoint to list tests in a campaign

tomrf1:
1. adds a new index on (`campaignName`, `name`)
2. adds a GET endpoint for listing all tests in a campaign, ordered by channel
3. adds Decoder/Encoder for the `ChannelTest` trait. This is to validate the tests we get back from dynamodb. Arguably this is a bit excessive - we should be able to trust the data in dynamodb.

### DIFF
--- a/app/models/BannerTests.scala
+++ b/app/models/BannerTests.scala
@@ -1,17 +1,14 @@
 package models
 
-import enumeratum.{CirceEnum, Enum, EnumEntry}
 import io.circe.generic.extras.Configuration
 import io.circe.generic.extras.auto._
+import io.circe.generic.extras.semiauto._
 import io.circe.{Decoder, Encoder}
 
-import scala.collection.immutable.IndexedSeq
 
 
-sealed trait BannerTemplate extends EnumEntry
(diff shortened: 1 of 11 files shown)


app/models/BannerTests.scala
```diff
@@ -66,6 +67,6 @@ case class BannerTest(
 
 object BannerTest {
   implicit val customConfig: Configuration = Configuration.default.withDefaults
-  implicit val bannerTestDecoder = Decoder[BannerTest]
-  implicit val bannerTestEncoder = Encoder[BannerTest]
+  implicit val bannerTestDecoder: Decoder[BannerTest] = deriveConfiguredDecoder[BannerTest]
+  implicit val bannerTestEncoder: Encoder[BannerTest] = deriveConfiguredEncoder[BannerTest]
```
tomrf1, line 71: Previously this used circe's 'automatic' derivation, but I've changed it to 'semi-automatic' derivation.
This is because the compiler gets confused around the new `ChannelTest` encoder/decoder if we use automatic. There's quite a lot of magic happening at compile-time.

MaelGNM (approved):
Looks good to me!

PR #356, opened 2022-06-20
Enable test selection from the url

tomrf1:
e.g. `/epic-tests/MY_EPIC_TEST` will immediately open the test named `MY_EPIC_TEST`, if it exists.

On the client, it uses an optional url param with react-router:
https://v5.reactrouter.com/web/example/url-params

On the server, I've had to add an additional endpoint for each channel to enable this. But the server doesn't actually do anything with the `testName` from the path, as the client does all the work.

This will be useful for the campaigns page, where we may want to link to tests.
But it's also useful for linking people to a specific test in e.g. chat.

Edit - following a suggestion from @KaliedaRik I've also added a "copy link" button:
Screen Shot 2022-06-20 at 15 02 14


KaliedaRik (approved):
I am very happy with this improvement - it opens up some really useful functionality!

PR #357, opened 2022-06-21
Campaigns page - initial restructure work

KaliedaRik:
## What does this change?
Build out the new campaigns page

TODO: details

TODO: screenshots

### DIFF
--- a/app/models/Campaigns.scala
+++ b/app/models/Campaigns.scala
@@ -4,7 +4,11 @@ import io.circe.generic.extras.Configuration
 import io.circe.generic.extras.auto._
 import io.circe.{Decoder, Encoder}
 
-case class Campaign(name: String)
+case class Campaign(
+  name: String,
+  nickname: Option[String],
+  description: Option[String]
+)
 
 object Campaigns {
   type Campaigns = List[Campaign]
--- a/public/src/components/channelManagement/campaigns/CampaignArchiveButton.tsx
+++ b/public/src/components/channelManagement/campaigns/CampaignArchiveButton.tsx
@@ -1,0 +1,72 @@
+import React from 'react';
+import useOpenable from '../../../hooks/useOpenable';
+import {
+  Button,
+  Dialog,
+  DialogActions,
+  DialogContent,
+  DialogContentText,
+  DialogTitle,
+  makeStyles,
+  Theme,
+  Typography,
+} from '@material-ui/core';
+import ArchiveIcon from '@material-ui/icons/Archive';
+import { grey } from '@material-ui/core/colors';
+
+const useStyles = makeStyles(({ palette }: Theme) => ({
+  buttonText: {
+    fontSize: '14px',
+    fontWeight: 500,
+    textTransform: 'uppercase',
+    letterSpacing: '1px',
+    color: palette.grey[800],
+  },
+}));
+
+interface CampaignArchiveButtonProps {
+  onCampaignArchive: () => void;
+}
+
+export const CampaignArchiveButton: React.FC<CampaignArchiveButtonProps> = ({
+  onCampaignArchive,
+}: CampaignArchiveButtonProps) => {
+  const classes = useStyles();
+  const [isOpen, open, close] = useOpenable();
+
+  return (
+    <>
+      <Button
+        variant="outlined"
+        startIcon={<ArchiveIcon style={{ color: grey[700] }} />}
+        size="medium"
+        onClick={open}
+      >
+        {/* eslint-disable-next-line react/prop-types */}
+        <Typography className={classes.buttonText}>Archive campaign</Typography>
+      </Button>
+      <Dialog
+        open={isOpen}
+        onClose={close}
+        aria-labelledby="archive-campaign-dialog-title"
+        aria-describedby="archive-campaign-dialog-description"
+      >
+        <DialogTitle id="archive-campaign-dialog-title">Are you sure?</DialogTitle>
+        <DialogContent>
+          <DialogContentText id="archive-campaign-dialog-description">
+            Archiving this campaign will remove it from the RRCP - you can only restore it with an
+            engineer&apos;s help.
+          </DialogContentText>
+        </DialogContent>
+        <DialogActions>
+          <Button color="primary" onClick={close}>
+            Cancel
+          </Button>
+          <Button color="primary" onClick={onCampaignArchive}>
+            Archive campaign
+          </Button>
+        </DialogActions>
+      </Dialog>
+    </>
+  );
+};
--- a/public/src/components/channelManagement/campaigns/CampaignCopyButton.tsx
+++ b/public/src/components/channelManagement/campaigns/CampaignCopyButton.tsx
@@ -1,0 +1,64 @@
+import React from 'react';
+import useOpenable from '../../../hooks/useOpenable';
+import { Button, makeStyles, Theme, Typography } from '@material-ui/core';
+import FileCopyIcon from '@material-ui/icons/FileCopy';
+import { grey } from '@material-ui/core/colors';
+import CreateCampaignDialog from './createCampaignDialog';
+
+const useStyles = makeStyles(({ palette }: Theme) => ({
+  buttonText: {
+    fontSize: '14px',
+    fontWeight: 500,
+    textTransform: 'uppercase',
+    letterSpacing: '1px',
+    color: palette.grey[800],
+  },
+}));
+
+interface CampaignCopyButtonProps {
+  existingNames: string[];
+  existingNicknames: string[];
+  // sourceName: string;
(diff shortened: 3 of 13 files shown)


public/src/components/channelManagement/campaigns/CampaignArchiveButton.tsx
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
+import useOpenable from '../../../hooks/useOpenable';
```
KaliedaRik, line 2: This PR skips the opportunity to display the `archive` button in the sticky top bar. The functionality of such a button is going to be much more involved than archiving a test as we need to make decisions about whether archiving a campaign also automatically archives all the tests associated with that campaign. 

There's also the issue about having campaigns data stored in S3 instead of DynamoDB - we'd need to resurrect the old tests archiving functionality so campaign data can be stored in S3 ... which would be wasted work if we make the decision (fairly soon) to shift campaign data into DynamoDB so it can sit alongside channel test data

public/src/components/channelManagement/campaigns/CampaignCopyButton.tsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import useOpenable from '../../../hooks/useOpenable';
+import { Button, makeStyles, Theme, Typography } from '@material-ui/core';
```
KaliedaRik, line 3: This PR skips the opportunity to display the `copy` button in the sticky top bar. The functionality of such a button is going to be much more involved than copying a test as we need to make decisions about whether copying a campaign also automatically copies all the tests associated with that campaign.

I think solving the copy dilemma is of lower priority compared to the functionality we want to add to the campaign display - especially listing all the tests associated with a campaign (including links to the individual test views)

public/src/components/channelManagement/campaigns/CampaignsForm.tsx
```diff
@@ -0,0 +1,125 @@
...
+    setNewCampaignCreated(true);
+    setSelectedCampaign(campaign);
+  };
+
+  const saveAllCampaigns = (): void => {
+    saveCampaigns();
+    setNewCampaignCreated(false);
+  };
+
+  // const saveCampaign = (campaign: Campaign): void => {
```
KaliedaRik, line 73: Commented out for now. Save and archive (not delete!) functional will come once we make decisions around HOW we want to approach campaign edit/locking, and campaign archiving

public/src/components/channelManagement/campaigns/CampaignsList.tsx
```diff
@@ -0,0 +1,78 @@
...
+}
+
+const CampaignsList = ({
+  campaigns,
+  selectedCampaign,
+  onCampaignSelected,
+}: CampaignsListProps): React.ReactElement => {
+  const classes = useStyles();
+
+  const checkIfCampaignIsSelected = (): 'outlined' => {
```
KaliedaRik, line 52: TODO (but lower priority): add in the visual signalling surrounding whether a campaign has been selected for viewing (and eventually, editing)

public/src/components/channelManagement/campaigns/StickyCampaignBar.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+  return (
+    <header className={classes.container}>
+      <div className={classes.namesContainer}>
+        <Typography variant="h2" className={classes.mainHeader}>
+          {mainHeader}
+        </Typography>
+        <Typography className={classes.secondaryHeader}>{secondaryHeader}</Typography>
+      </div>
+      {/*
```
KaliedaRik, line 77: See comments in other files. For now, I'm suppressing the display of `copy`, `edit` (and `archive`) buttons until after we've made decisions on how to handle these functionalities

app/models/Campaigns.scala
```diff
@@ -4,7 +4,11 @@ import io.circe.generic.extras.Configuration
 import io.circe.generic.extras.auto._
 import io.circe.{Decoder, Encoder}
 
-case class Campaign(name: String)
+case class Campaign(
+  name: String,
+  nickname: Option[String],
```
KaliedaRik, line 9: TODO: check whether `nickname` is optional

public/src/components/channelManagement/campaigns/CampaignsForm.tsx
```diff
@@ -0,0 +1,125 @@
...
+export type Campaigns = Campaign[];
+
+const CampaignsForm: React.FC<InnerProps<Campaigns>> = ({
+  data: campaigns,
+  setData: setCampaigns,
+  saveData: saveCampaigns,
+}: InnerProps<Campaigns>) => {
+  const classes = useStyles();
+
+  const createCampaign = (campaign: Campaign): void => {
```
KaliedaRik, line 62: I can't find a way around the old issue of getting a campaign/test to save itself in the S3 data when it is created. Instead, I've gone with the option of visually highlighting the `save all campaigns` button to make it more obvious to users that they need to save all the campaigns before navigating away from the page. 

Creating a campaign will select the new campaign for display, which should save users a little bit of work/frustration

public/src/components/channelManagement/campaigns/CampaignArchiveButton.tsx
tomrf1: If this functionality doesn't exist can we just remove this code? Is it copied from elsewhere?
Same goes for CampaignCopy etc

public/src/components/channelManagement/campaigns/CampaignsForm.tsx
```diff
@@ -0,0 +1,128 @@
...
+    setNewCampaignCreated(false);
+  };
+
+  const [newCampaignCreated, setNewCampaignCreated] = useState(false);
+
+  const [selectedCampaign, setSelectedCampaign] = useState<Campaign | undefined>();
+
+  const getSelectedCampaign = (): Campaign | undefined => {
+    if (selectedCampaign == null && testName != null) {
+      const requiredCampaign = campaigns.filter(c => c.name === testName);
```
tomrf1, line 81: prefer `campaigns.find`

public/src/components/channelManagement/campaigns/CampaignsForm.tsx
```diff
@@ -0,0 +1,128 @@
...
+      if (requiredCampaign[0] != null) {
+        return requiredCampaign[0];
+      }
+    }
+    return selectedCampaign;
+  };
+
+  const onCampaignSelected = (name: string) => {
+    const requiredCampaign = campaigns.filter(c => c.name === name);
```
tomrf1, line 90: prefer `campaigns.find`

public/src/components/channelManagement/campaigns/CampaignArchiveButton.tsx
KaliedaRik: File deleted - archive functionality will be too different from equivalent test functionality to reuse of this code.

public/src/components/channelManagement/campaigns/CampaignsSidebar.tsx
```diff
@@ -0,0 +1,88 @@
...
+  saveAllCampaigns,
+}: CampaignsSidebarProps): React.ReactElement {
+  const classes = useStyles();
+
+  return (
+    <div className={classes.root}>
+      <div className={classes.buttonsContainer}>
+        <NewCampaignButton
+          existingNames={campaigns.map(c => c.name)}
+          existingNicknames={campaigns.map(c => c.nickname)}
```
tomrf1, line 66: Elsewhere we default to empty string if nickname is undefined (a bit hacky), see https://github.com/guardian/support-admin-console/blob/main/public/src/components/channelManagement/testsForm.tsx#L288
This currently breaks in CODE where nickname isn't defined in the json

public/src/components/channelManagement/campaigns/CampaignsForm.tsx
KaliedaRik: Done

public/src/components/channelManagement/campaigns/StickyCampaignBar.tsx
KaliedaRik: Removed code - we'll add in editing functionality in a future PR

public/src/components/channelManagement/campaigns/CampaignsForm.tsx
KaliedaRik: Removed code - we'll add in editing functionality in a future PR

app/models/Campaigns.scala
KaliedaRik: Made nickname field mandatory

MaelGNM (approved):
Looks good to me

PR #358, opened 2022-06-23
Include null values in dynamodb

tomrf1:
Bug: if a test has an optional field (e.g. `articlesViewedSettings`) set, and a user removes that field in the UI, then the backend does not remove it from dynamodb because null values are dropped during serialization.

The solution is to include null values.
There is the special case of `campaignName`, which is optional, but is defined in an index as a `string` and so cannot be written with `null`. I have changed the backend to write `NOT_IN_CAMPAIGN` for this field if it is not set.


KaliedaRik (approved):
LGTM

PR #360, opened 2022-06-24
Show campaign name on tests

MaelGNM:
## What does this change?

This adds the campaign name to the top bar of the test editor (if the test is in a campaign).

## Images

Test in a campaign
Screenshot test in a campaign

Test not in a campaign
Screenshot test not in a campaign


tomrf1 (approved):
looks good, though I'm wondering if this line will ever become too long. We can always change the design later...

PR #363, opened 2022-06-27
Decouple test status from test editing

tomrf1:
Currently a user has to lock/edit the whole test to toggle status between Draft/Live.
This will become problematic when we add status toggling to the campaigns page.

This PR:
1. moves the toggle into the StickTopBar
2. the toggle becomes disabled while the test is being edited, to avoid any unexpected behaviour
3. replaces the `/test/archive` endpoint in favour of more general `test/status/:rawStatus` endpoint, for changing the status of a list of tests
4. removes the deprecated `isOn` field, which is no longer used by SDC

### Not editing the test:
Screen Shot 2022-06-27 at 15 57 01

### Editing the test:
Screen Shot 2022-06-27 at 16 02 26

### Confirmation dialogs:
Screen Shot 2022-06-27 at 16 02 35
Screen Shot 2022-06-27 at 16 02 45

### DIFF
--- a/public/src/components/channelManagement/epicTests/epicTestEditor.tsx
+++ b/public/src/components/channelManagement/epicTests/epicTestEditor.tsx
@@ -6,7 +6,6 @@ import {
   UserCohort,
   EpicEditorConfig,
   DeviceType,
-  setStatus,
 } from '../helpers/shared';
 import { FormControlLabel, Switch, Typography } from '@material-ui/core';
 import TestVariantsEditor from '../testVariantsEditor';
@@ -23,7 +22,6 @@ import EpicTestMaxViewsEditor from './epicTestMaxViewsEditor';
 import { ARTICLE_COUNT_TEMPLATE, COUNTRY_NAME_TEMPLATE } from '../helpers/validation';
 import TestVariantsSplitEditor from '../testVariantsSplitEditor';
 import { getDefaultVariant } from './utils/defaults';
-import LiveSwitch from '../../shared/liveSwitch';
 import {
   canHaveCustomVariantSplit,
   ControlProportionSettings,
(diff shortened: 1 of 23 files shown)


public/src/components/channelManagement/epicTests/epicTestEditor.tsx
```diff
@@ -207,23 +201,14 @@ export const getEpicTestEditor = (
...
         {epicEditorConfig.allowMultipleVariants && (
           <div className={classes.sectionContainer}>
-            <Typography variant={'h3'} className={classes.sectionHeader}>
-              Variants
-            </Typography>
+            <div className={classes.variantsHeaderContainer}>
+              <Typography variant={'h3'} className={classes.sectionHeader}>
+                Variants
+              </Typography>
+              <EpicTestPreviewButton test={test} />
```
tomrf1, line 210: this change moves the `EpicTestPreviewButton` button in-line with the "Variants" heading, because it used to be in-line with the status toggle

public/src/components/channelManagement/stickyTopBar/stickyTopBar.tsx
```diff
@@ -126,62 +142,72 @@ const StickyTopBar: React.FC<StickyTopBarProps> = ({
...
-              size="medium"
-              startIcon={<SaveIcon className={classes.icon} />}
-              onClick={() => onTestSave(name)}
-            >
-              <Typography className={classes.buttonText}>Save test</Typography>
-            </Button>
-          </>
-        )}
+
+      <div className={classes.buttonsContainer}>
```
tomrf1, line 146: this change moves all the ctas into the `buttonsContainer` div, and adds the `TestLiveSwitch`

public/src/components/channelManagement/testLiveSwitch.tsx
```diff
@@ -0,0 +1,100 @@
+import React from 'react';
```
tomrf1, line 1: this file is based on the existing `LiveSwitch` component, which was shared with the amounts tests tool. They're now sufficiently different that I don't think it's worth sharing a component

PR #365, opened 2022-06-30
updating cdk

KaliedaRik:
## What does this change?
A small update to the CDK definition to allow the EC2 instance permissions so it can read our DynamoDB index


tomrf1 (approved):
👍

PR #367, opened 2022-06-30
Campaigns page - bulk status switch

KaliedaRik:
## What does this change?
This PR adds functionality to the Campaigns page to allow a user to update the status of each test associated with that campaign. This will save users considerable time at the start and end of a campaign.

The functionality is triggered by a button at on the right hand side of the campaign sticky bar. When clicked, a modal displays, showing the current state of each test in the campaign. The user can then adjust these values; when they click on the modal's update button, the test state changes take immediate effect on the Guardian's main website.

### Screenshots:

Screenshot 2022-07-01 at 12 46 50

Screenshot 2022-07-01 at 12 47 39


public/src/components/channelManagement/campaigns/StatusUpdateDialog.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+    if (changes.length > 0) {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const promises: Promise<any>[] = [];
+      changes.forEach(change => {
+        promises.push(updateStatuses(...change));
+      });
+
+      Promise.all(promises).then(() => updatePage());
+    }
```
MaelGNM, line 115: You could group the names and only send one request per channel

public/src/components/channelManagement/campaigns/StatusUpdateDialog.tsx
```diff
@@ -0,0 +1,165 @@
...
+  isOpen,
+  close,
+  tests,
+  updatePage,
+}: StatusUpdateDialogProps) => {
+  const classes = useStyles();
+
+  const oldTestStatuses: TestStatus = {};
+
+  const [testData, setTestData] = useState<TestStatus>(oldTestStatuses);
```
tomrf1, line 68: Here we're initialising a `useState` with a value which is mutated elsewhere, which isn't very idiomatic and I think potentially buggy.
You can implement this without defining `oldTestStatuses`.

First, you can initialise a `useState` with a function (instead of the `useEffect` below):
```
const [testData, setTestData] = useState<TestStatus>(() => {
  const oldTestStatuses: TestStatus = {};
  tests.map(test => {
    const key = getStatusKey(test);
    oldTestStatuses[key] = test.status;
  });
  return oldTestStatuses;
});
```

Later, in the `onSubmit` function where you're iterating over `tests`, you can instead use:
`if (test.status !== testData[key])`

public/src/components/channelManagement/campaigns/StatusUpdateDialog.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+interface StatusUpdateDialogProps {
+  isOpen: boolean;
+  close: () => void;
+  tests: Test[];
+  updatePage: () => void;
+}
+
+interface TestStatus {
+  [index: string]: string;
```
tomrf1, line 53: `[index: string]: Status;`?

public/src/components/channelManagement/campaigns/StatusUpdateDialog.tsx
tomrf1: Agreed - it'll be much faster if you batch them, and the endpoint was designed to take a batch of test names.
It's a bit fiddly, but how about replacing `UpdateStatusesArguments` with something like this to group test names by channel/status:
```
interface StatusUpdates {
  [channel: string]: {
    [status in Status]: string[];
  }
}
```

public/src/components/channelManagement/campaigns/StatusUpdateDialog.tsx
KaliedaRik: I tried the solution, but it doesn't work. The campaign's test data is asynchronous and doesn't resolve until after the component completes loading (as part of the page load).

This code works. Does it address your mutated object worries?

```
  const [testData, setTestData] = useState<TestStatus>({});

  useEffect(() => {
    const res: TestStatus = {};
    tests.map(test => {
      const key = getStatusKey(test);
      res[key] = test.status;
    });
    setTestData(res);
  }, [tests]);

```

public/src/components/channelManagement/campaigns/StatusUpdateDialog.tsx
tomrf1: I'm happy to ship as it is for now, but I'll create a card to do batching later. We should leverage the backend/dynamo's ability to send a batch of updates as a single transaction. We can make the backend's endpoint more helpful here, so the client does less work.

tomrf1 (approved):
great feature!
We'll improve support for batch updates in a follow up - this will require some backend changes

public/src/components/channelManagement/campaigns/StatusUpdateDialog.tsx
KaliedaRik: Hi @tomrf1  - what would your `happy place` be on this? My `happy place` would be a single endpoint which accepts an array of objects containing test update data (similar to the one you proposed above) which the backend could process and send out to Dynamo in a single update call.

tomrf1: yes ideally a single endpoint which takes an array of objects with name/channel/status. Though currently the backend has a controller per channel so it may be a bit of a refactor. Or perhaps we can live with sending a request for each channel. Something to be decided when the card is picked up

PR #368, opened 2022-07-01
Add editable campaign field to test forms

KaliedaRik:
## What does this change?
The current functionality is for users to be able to assign a test to an existing campaign when that test is created. After creation there is no way for a user to assign the test to a different campaign, or remove it from the campaign.

This makes using the new campaigns functionality difficult for users as there may be times when they create some tests before thinking to group them together in a new campaign. It also means existing, long running tests that should be grouped together (for instance, the "evergreen" default tests, or the tests dedicated to the Ukraine war) cannot be assigned to a campaign.

This PR addresses these issues by adding an editable `campaign` selector to test forms. The selector will list all existing campaigns, and the update saved in the normal way.

### Screenshots

Screenshot 2022-07-05 at 13 03 20

Screenshot 2022-07-05 at 13 04 22

### DIFF
--- a/public/src/components/channelManagement/CampaignSelector.tsx
+++ b/public/src/components/channelManagement/CampaignSelector.tsx
@@ -39,6 +39,8 @@ interface CampaignSelectorProps {
   disabled: boolean;
 }
 
+const unassignedCampaignLabel = 'NOT_IN_CAMPAIGN';
+
 const CampaignSelector: React.FC<CampaignSelectorProps> = ({
   onCampaignChange,
   test,
@@ -63,9 +65,6 @@ const CampaignSelector: React.FC<CampaignSelectorProps> = ({
           value={test.campaignName}
           displayEmpty
           renderValue={(campaign): string => {
-            if (campaign === undefined) {
-              return ''; // triggers the displayEmpty behaviour
-            }
             return campaign as string;
           }}
           onChange={(event: React.ChangeEvent<{ value: unknown }>): void => {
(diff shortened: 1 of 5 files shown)


public/src/components/channelManagement/CampaignSelector.tsx
```diff
@@ -0,0 +1,93 @@
...
+          {campaigns.map(campaign => (
+            <MenuItem value={campaign.name} key={`campaign-${campaign.name}`}>
+              {campaign.name}
+            </MenuItem>
+          ))}
+        </Select>
+      </FormControl>
+      <div className={classes.warning}>
+        If you want to use this Test in a new or different campaign, please consider copying it,
+        rather than changing the Test campaign value.
```
tomrf1, line 87: what's the reason for this message? It takes up a lot of room!

public/src/components/channelManagement/bannerTests/bannerTestEditor.tsx
```diff
@@ -157,6 +165,19 @@ const BannerTestEditor: React.FC<ValidatedTestEditorProps<BannerTest>> = ({
   if (test) {
     return (
       <div className={classes.container}>
+        <div className={classes.sectionContainer}>
+          <Typography variant={'h3'} className={classes.sectionHeader}>
+            Campaign
```
tomrf1, line 170: I'm worried about pushing important things down the page - and having an h3 with the dropdown below it takes up a lot of room! What do you think about having a label + dropdown horizontally instead?

public/src/components/channelManagement/CampaignSelector.tsx
MaelGNM: I was concerned about moving tests in a different campaign when the name of the original campaign is as a prefix in the test name. This message is to encourage good practices.
But it's true that the whole campaign part takes quite a lot of space.

KaliedaRik: Message added in line with Mael's comments. Message moved to sit to the right of the selector box.

public/src/components/channelManagement/bannerTests/bannerTestEditor.tsx
KaliedaRik: Campaigns section moved down the page to follow the variants, and variants split editor, sections

PR #369, opened 2022-07-05
Aus brand moment banner

lucymonie:
## What does this change?
This is to support the Australia brand moment banner that's due to kick off on 18 July.

### DIFF
--- a/public/src/components/channelManagement/bannerTests/bannerVariantPreview.tsx
+++ b/public/src/components/channelManagement/bannerTests/bannerVariantPreview.tsx
@@ -152,6 +152,10 @@ const bannerModules = {
     path: 'postElectionAuMoment/PostElectionAuMomentMorrisonBanner.js',
     name: 'PostElectionAuMomentMorrisonBanner',
   },
+  [BannerTemplate.AusBrandMomentBanner]: {
+    path: 'ausBrandMoment/ausBrandMomentBanner.js',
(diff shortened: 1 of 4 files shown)


public/src/components/channelManagement/bannerTests/bannerVariantPreview.tsx
```diff
@@ -152,6 +152,10 @@ const bannerModules = {
     path: 'postElectionAuMoment/PostElectionAuMomentMorrisonBanner.js',
     name: 'PostElectionAuMomentMorrisonBanner',
   },
+  [BannerTemplate.AusBrandMomentBanner]: {
+    path: 'ausBrandMoment/ausBrandMomentBanner.js',
```
tompretty, line 156: This just needs to be `ausBrandMoment/AusBrandMomentBanner.js`

lucymonie: Thanks!

tompretty (approved):
Looks great! Just a small note: If we want the highlighted text at the end of the banner to be editable, we'll have to add

```ts
template === BannerTemplate.AuBrandMomentBanner
```

to the end of [this condition](https://github.com/guardian/support-admin-console/blob/main/public/src/components/channelManagement/bannerTests/bannerTestVariantEditor.tsx#L287).

PR #373, opened 2022-07-13
Add navigation buttons to campaign channel cards

KaliedaRik:
## What does this change?
This is a small UX fix to help users navigate from the campaigns page a little quicker. It adds navigation buttons to each channel card to take the user to that channel's main page.

### Before:
Screenshot 2022-07-13 at 17 04 13

### After:
Screenshot 2022-07-13 at 17 04 56


tomrf1 (approved):
👍

PR #376, opened 2022-07-21
RRCP archived tests - display and copy test data in a modal

KaliedaRik:
Note that most of these *apparent* commits are in fact commits made in a the PR from which I branched this PR

## What does this change?
Following on from adding the ability to display archived Tests in the Campaigns page, this PR adds functionality to allow users to view the data associated with a Test in a modal and, from that modal, copy the data to the clipboard so that it can be copied elsewhere - for example a Google Doc.

The PR also includes some UX changes to the campaign listing:

- Separate the `navigate to Test page` functionality from each test name into its own button in the top right of the Test's card
- Add a `View data` button to each test card
- The `view` modal includes a `copy data` button so users can copy the test data and paste it into a Google doc

### Screenshots
Screenshot 2022-07-22 at 17 42 35

Screenshot 2022-07-22 at 17 42 54

Screenshot 2022-07-22 at 17 43 07

Screenshot 2022-07-22 at 17 43 20


public/src/components/channelManagement/campaigns/TestDataDialog.tsx
```diff
@@ -0,0 +1,537 @@
...
+    locations: {
+      label: 'Locations',
+      type: 'string-array',
+      exclude: [],
+      optional: false,
+    },
+    deviceType: {
+      label: 'Device type',
+      type: 'string',
+      exclude: ['Banner1', 'Banner2'],
```
MaelGNM, line 100: There is a device type on banners, is it not actually used?

public/src/components/channelManagement/campaigns/TestDataDialog.tsx
```diff
@@ -0,0 +1,537 @@
...
+        res += `Variant: ${v.name}
+---------------------------------------------------------------------
+${parseData(channel, variantFields.core, v)}
+Copy and CTAs
+~~~~~~~~~~~~~~~~~~~~~~~~
+${parseData(channel, variantFields.copy, v.content)}`;
+
+        if (v.mobileContent == null) {
+          res += `
+Mobile copy and CTAs
```
MaelGNM, line 433: You could move this two lines out of the `if` and not repeat them

public/src/components/channelManagement/campaigns/TestDataDialog.tsx
```diff
@@ -0,0 +1,537 @@
...
+${parseData(channel, variantFields.core, v)}
+Copy and CTAs
+~~~~~~~~~~~~~~~~~~~~~~~~
+${parseData(channel, variantFields.copy, v.bannerContent)}
+`;
+
+        if (v.mobileBannerContent == null) {
+          res += `
+Mobile copy and CTAs
+~~~~~~~~~~~~~~~~~~~~~~~~
```
MaelGNM, line 471: As above, these could go out of the `if`

public/src/components/channelManagement/campaigns/TestDataDialog.tsx
KaliedaRik: You're right. Good spot!

public/src/components/channelManagement/campaigns/TestDataDialog.tsx
KaliedaRik: I did it this way because Epic variants don't have the capacity to display different messaging or CTAs on mobile devices, unlike Banners and Headers. (This is different to the device targeting issue you spotted above)

public/src/components/channelManagement/campaigns/TestDataDialog.tsx
MaelGNM: Not really useful, marking as resolved

PR #379, opened 2022-07-25
Async credentials update

tomrf1:
This was suggested by AWS as a possible solution to expired credentials.
I've also bumped the AWS version - `asyncCredentialUpdateEnabled` breaks the credentials refresh completely in the older version.


app/services/Aws.scala
```diff
@@ -7,8 +7,11 @@ import software.amazon.awssdk.regions.Region
 object Aws {
   val credentialsProvider = AwsCredentialsProviderChain.builder().credentialsProviders(
     ProfileCredentialsProvider.builder.profileName("membership").build,
-    InstanceProfileCredentialsProvider.builder.asyncCredentialUpdateEnabled(false).build
-  )
+    InstanceProfileCredentialsProvider
+      .builder
+      .asyncCredentialUpdateEnabled(true)
+      .build
+  ).build
```
tomrf1, line 14: Note that we now `build` the `AwsCredentialsProviderChain` once here. Otherwise we get errors from the underlying `ThreadPoolExecutor`

KaliedaRik (approved):
All looks fine!

PR #384, opened 2022-08-19
fix useEffect to rectify issue

KaliedaRik:
## What does this change?
This PR fixes a live preview issue for Banner templates. Currently, user is able to change a test variant's template, but cannot view the banner copy in the live preview until after they have saved the changes, clicked on a different banner test, and then returned to the original test - at this point the live preview will display the newly selected banner template.

Once this PR is merged users will be able to change the banner template while editing the test and, when they click on the live preview button, the correct banner template will appear.

### DIFF
--- a/public/src/hooks/useModule.ts
+++ b/public/src/hooks/useModule.ts
@@ -30,7 +30,7 @@ export const useModule = <T>(path: string, name: string): React.FC<T> | undefine
     window.remoteImport(`${baseUrl}/${path}`).then(bannerModule => {
       setModule(() => withPreviewStyles(bannerModule[name]));
     });
-  }, []);
+  }, [name]);
 
   return Module;
 };


public/src/hooks/useModule.ts
```diff
@@ -30,7 +30,7 @@ export const useModule = <T>(path: string, name: string): React.FC<T> | undefine
     window.remoteImport(`${baseUrl}/${path}`).then(bannerModule => {
       setModule(() => withPreviewStyles(bannerModule[name]));
     });
-  }, []);
+  }, [name]);
```
tjmw, line 33: [eslint-plugin-react-hooks](https://www.npmjs.com/package/eslint-plugin-react-hooks) has a rule `react-hooks/exhaustive-deps` which lets you know about things your effect depends on but that aren't in the dependency list passed to `useEffect`. Maybe worth adding?

tjmw (approved):

PR #385, opened 2022-08-24
Disable save button while saving

tomrf1:
The amounts tool UI gives no feedback after clicking save, and sometimes users spam the save button as a result.

This PR updates the `withS3Data` HOC to pass a `saving` boolean into the wrapped component.
The amounts tool can then disable the save button while `saving === true`.
Screen Shot 2022-08-24 at 08 29 26

In the 2nd commit I updated all other components which use `withS3Data` to also disable the save button.


KaliedaRik (approved):
Nice! No more alarms (fingers crossed)

PR #394, opened 2022-10-07
fix liveblog epic preview

tomrf1:
Fixes the querystring parameter when you open the web preview (it should be `force-liveblog-epic`, not `force-epic`)

### DIFF
--- a/public/src/components/channelManagement/testEditorVariantSummaryWebPreviewButton.tsx
+++ b/public/src/components/channelManagement/testEditorVariantSummaryWebPreviewButton.tsx
@@ -36,10 +36,8 @@ const getPreviewUrl = (
   articleType: ArticleType,
 ): string => {
   const stage = getStage();
-  const queryString = `?dcr&force-${getChannelName(
-    testType,
-    articleType,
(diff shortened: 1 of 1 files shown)


public/src/components/channelManagement/testEditorVariantSummaryWebPreviewButton.tsx
```diff
@@ -9,7 +9,7 @@ export type ArticleType = 'Standard' | 'Liveblog';
 const ArticlePaths: Record<ArticleType, string> = {
   Standard:
     'world/2020/may/08/commemorating-ve-day-during-coronavirus-lockdown-somehow-the-quiet-made-it-louder',
-  Liveblog: 'uk-news/live/2022/sep/19/queen-elizabeth-ii-state-funeral-westminster-abbey-updates',
+  Liveblog: 'world/live/2022/oct/07/nobel-peace-prize-2022-live-winners',
```
tomrf1, line 12: I've changed the liveblog article because in the old one the first block was massive

tjmw (approved):

PR #395, opened 2022-10-14
Delete AU election banners

tomrf1:
SDC PR - https://github.com/guardian/support-dotcom-components/pull/773


KaliedaRik (approved):
Brings SAC in line with SDSC cleanup

PR #396, opened 2022-10-17
Better error message for duplicate name errors

tomrf1:
Currently if a user tries to create a test with a name that already exists, the error message is not user friendly:
err
This can happen if the existing test is archived (and so the client doesn't know about it).

This PR changes the backend to handle the condition check error and return a better message to the client:
Screenshot 2022-10-17 at 09 35 38


KaliedaRik (approved):
Sensible change

PR #399, opened 2022-10-31
Bat

tomrf1:
Screenshot 2022-10-31 at 10 20 38


public/src/components/drawer.tsx
```diff
@@ -112,6 +115,11 @@ export default function NavDrawer(): React.ReactElement {
     setState({ ...state, [anchor]: open });
   };
 
+  const isHalloween = () => {
+    const now = new Date();
+    return now.getMonth() == 9 && now.getDate() == 31;
```
tjmw, line 120: Month being 0 indexed gets me every time!

tjmw (approved):
🎃

public/src/components/drawer.tsx
tomrf1: haha I've seen so many bugs caused by this, it's horrible!

PR #398, opened 2022-10-24
Add new campaigns dynamo table

silvacb:
## What does this change?

Updates the cdk script to create a new dynamo table for campaigns data.

PR #401, opened 2022-11-02
Move campaigns data from S3 to DynamoDB

silvacb:
## What does this change?
Moves campaigns data that was previously stored in S3 to a DynamoDB table  ([PR to create the table](https://github.com/guardian/support-admin-console/pull/398)).

#### Server-side changes:

-  Introduces the `DynamoCampaigns` class, with basic functionality for writing campaigns to the table.
- The existing `CampaingsController` now takes a `createCampaign` function to create a campaign.
- A new endpoint is added to the `routes` file.
- A new `DynamoErrors` file is created as the erros are no longer used only by a single class.

#### Client-side changes:

- `sendCreateCampaignRequest` function added to `CampaignForms`, which creates a new row in the DynamoDB table with the newly added campaign.
- `Save all campaigns` button has been removed from the UI as each campaign is now saved automatically when created.

Paired with @tomrf1 and @KaliedaRik

## Images

| Before      | After   | 
|-------------|------------|
| image | image |


app/controllers/CampaignsController.scala
```diff
@@ -16,13 +18,35 @@ class CampaignsController(
   components: ControllerComponents,
   stage: String,
   runtime: zio.Runtime[ZEnv],
-  dynamo: DynamoChannelTests
+  dynamoChannelTests: DynamoChannelTests,
+  dynamoCampaigns: DynamoCampaigns
 )(implicit ec: ExecutionContext)
   extends S3ObjectController[Campaigns](authAction, components, stage, filename = "campaigns.json", runtime) {
```
tomrf1, line 24: I've just realised we can now remove this line. This controller no longer needs to interact with S3 (by inheriting from the `S3ObjectController`)

app/controllers/CampaignsController.scala
```diff
@@ -16,13 +18,35 @@ class CampaignsController(
...
 
-  def getTests(campaignName: String) = authAction.async { request =>
+  override def get(): Action[AnyContent] = authAction.async { request =>
```
tomrf1, line 26: you can also remove `override` here once it's no longer extending `S3ObjectController`

app/services/DynamoCampaigns.scala
```diff
@@ -0,0 +1,67 @@
+package services
+
+import com.typesafe.scalalogging.StrictLogging
+import models.Campaign
+import services.DynamoChannelTests.{DynamoDuplicateNameError, DynamoError, DynamoGetError, DynamoPutError}
```
tomrf1, line 5: I think we should now move the `DynamoError` models into a separate file, because they're no longer only used by the `DynamoChannelTests` class.
How about creating `app/models/DynamoError.scala`?

app/controllers/CampaignsController.scala
silvacb: It seems that it works if instead of `S3ObjectController` we extend `AbstractController with Circe with LazyLogging` and implement a private `run` function (as in `ChannelTestsController`). Simply deleting the above line will throw errors.

tomrf1: ah yes, that sounds right!

silvacb: Also, if I go ahead with these changes, the `controllers.CampaignsController.set` is broken, so I assume we'd need to add a `set` function as well to the controller (would this be similar to `setStatus` in `ChannelTestsController`?).

tomrf1: The endpoint using `controllers.CampaignsController.set` can also be removed. It was for POSTing the full array of campaigns to S3. That endpoint is no longer used now that you've removed the call to `saveSettings` in the client code

app/services/DynamoCampaigns.scala
silvacb: Good suggestion, thanks. I've decided to go with DynamoErrors as it contains several error classes.

tomrf1 (approved):
Code looks great!
We need to migrate the old campaigns data from S3 before merging. It's perhaps small enough that this can be done manually

PR #402, opened 2022-11-14
Remove 'supporter ticker' option

tomrf1:
This type of ticker hasn't been used in a long time, and the system that calculated it would need to be migrated to use BigQuery if we ever need it again.
We can simplify the SDC ticker logic if we remove support for this.

[SDC PR](https://github.com/guardian/support-dotcom-components/pull/792)

Before:
Screenshot 2022-11-14 at 13 55 53

After:
Screenshot 2022-11-14 at 13 55 29

### DIFF
--- a/public/src/models/epic.ts
+++ b/public/src/models/epic.ts
@@ -7,20 +7,14 @@ import {
   SecondaryCta,
   Status,
   Test,
+  TickerCountType,
+  TickerEndType,
   UserCohort,
   Variant,
 } from '../components/channelManagement/helpers/shared';
 import { Region } from '../utils/models';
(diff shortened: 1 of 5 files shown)


public/src/models/epic.ts
```diff
@@ -7,20 +7,14 @@ import {
...
   Test,
+  TickerCountType,
+  TickerEndType,
   UserCohort,
   Variant,
 } from '../components/channelManagement/helpers/shared';
 import { Region } from '../utils/models';
 import { ControlProportionSettings } from '../components/channelManagement/helpers/controlProportionSettings';
 
-export enum TickerEndType {
```
tomrf1, line 16: these types are already defined in `shared.ts`

PR #403, opened 2022-11-16
Banner ticker settings

tomrf1:
Sometimes banner templates can support tickers.
SDC already supports `tickerSettings` in the `BannerVariant` model - https://github.com/guardian/support-dotcom-components/blob/main/packages/shared/src/types/abTests/banner.ts#L33

This PR makes it possible to configure the epic from the banner tool. Currently no template supports tickers, but the upcoming US + AU banners will.

Screenshot 2022-11-16 at 08 19 03

### DIFF
--- a/app/models/EpicTest.scala
+++ b/app/models/EpicTest.scala
@@ -8,34 +8,6 @@ import io.circe.{Decoder, Encoder}
 
 import scala.collection.immutable.IndexedSeq
 
-sealed trait TickerEndType extends EnumEntry
-object TickerEndType extends Enum[TickerEndType] with CirceEnum[TickerEndType] {
-  override val values: IndexedSeq[TickerEndType] = findValues
-
-  case object unlimited extends TickerEndType
-  case object hardstop extends TickerEndType
-}
-
-sealed trait TickerCountType extends EnumEntry
-object TickerCountType extends Enum[TickerCountType] with CirceEnum[TickerCountType] {
-  override val values: IndexedSeq[TickerCountType] = findValues
-
-  case object money extends TickerCountType
-}
-
-case class TickerCopy(
-  countLabel: String,
-  goalReachedPrimary: String,
-  goalReachedSecondary: String
-)
-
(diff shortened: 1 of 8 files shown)


app/models/EpicTest.scala
```diff
@@ -8,34 +8,6 @@ import io.circe.{Decoder, Encoder}
 
 import scala.collection.immutable.IndexedSeq
 
-sealed trait TickerEndType extends EnumEntry
```
tomrf1, line 11: Moved to Ticker.scala

public/src/models/epic.ts
```diff
@@ -7,26 +7,13 @@ import {
...
-  TickerCountType,
-  TickerEndType,
+  TickerSettings,
   UserCohort,
   Variant,
 } from '../components/channelManagement/helpers/shared';
 import { Region } from '../utils/models';
 import { ControlProportionSettings } from '../components/channelManagement/helpers/controlProportionSettings';
 
-interface TickerCopy {
```
tomrf1, line 18: these models are also defined in shared.ts

public/src/components/channelManagement/bannerTests/bannerTestVariantEditor.tsx
```diff
@@ -356,6 +357,8 @@ const BannerTestVariantEditor: React.FC<BannerTestVariantEditorProps> = ({
   const classes = useStyles();
   const setValidationStatusForField = useValidation(onValidationChange);
 
+  const allowVariantTicker = false; // TODO - set based on template
```
tjmw, line 360: So until this line is changed, this PR won't have a visible affect on any RRCP pages?

tjmw (approved):

public/src/components/channelManagement/bannerTests/bannerTestVariantEditor.tsx
tomrf1: that's right

PR #408, opened 2022-11-24
Super mode dashboard

tomrf1:
A new dashboard.

1. Table view of all articles that are currently in super mode (a row per article/region). This data comes from the super-mode Dynamodb table.

Screenshot 2022-11-30 at 09 37 24

2. Bar charts with recent impressions/conversions data for a selected article. This data comes from Athena (which is where super mode gets its data from)

Screenshot 2022-11-30 at 13 49 10

On the client it uses:
- `@mui/x-data-grid` for the table
- `recharts` for the charts


conf/application.conf
```diff
@@ -5,3 +5,15 @@ googleAuth.domain = "guardian.co.uk"
...
 
 play.http.parser.maxMemoryBuffer=4MB
+
+athena {
+  default {
+    driver="com.simba.athena.jdbc.Driver"
+    url="jdbc:awsathena://AwsRegion=eu-west-1"
+    readOnly="true"
+    S3OutputLocation="s3://gu-support-analytics/support-admin-console/"
+    AwsCredentialsProviderClass="com.simba.athena.amazonaws.auth.profile.ProfileCredentialsProvider"
```
tomrf1, line 15: TODO - this only works locally. But the `scalikejdbc-athena` library doesn't support different configurations - it can only load config from `conf/application.conf`.
Either:
1. submit a PR to make the conf path configurable
2. use a different library

app/controllers/CapiController.scala
```diff
@@ -20,4 +20,10 @@ class CapiController(authAction: AuthAction[AnyContent], capiService: CapiServic
       .getSections(request.rawQueryString)
       .map(response => Ok(response))
   }
+
+  def getContent(path: String) = authAction.async { request =>
+    capiService
+      .getContent(path)
+      .map(response => Ok(response))
+  }
```
tomrf1, line 28: to get the article headline + publication timestamp

app/services/Athena.scala
```diff
@@ -0,0 +1,177 @@
+package services
```
tomrf1, line 1: the downside to athena is that executing queries is very complicated!

cdk/lib/admin-console.ts
```diff
@@ -160,15 +176,18 @@ export class AdminConsole extends GuStack {
...
+          ? {
+              http5xxAlarm: {
+                tolerated5xxPercentage: 0,
+                numberOfMinutesAboveThresholdBeforeAlarm: 1,
+                alarmName: `5XX error returned by ${app} ${this.stage}`,
+              },
+              unhealthyInstancesAlarm: true,
+              snsTopicName: 'marketing-dev',
+            }
+          : { noMonitoring: true },
```
tomrf1, line 190: disable alarms in CODE

conf/application.conf
tomrf1: solved by using the aws athena client instead

app/services/Athena.scala
KaliedaRik: Naive question: how "live" is the extracted data? Or, to put it another way, how often does the query run?

app/services/CapiService.scala
```diff
@@ -7,13 +7,15 @@ import scala.concurrent.{ExecutionContext, Future}
...
       .map(response => response.body)
 
-  def getTags(querystring: String): Future[String] = request("tags", querystring)
+  def getTags(querystring: String): Future[String] = request("tags", Some(querystring))
 
-  def getSections(querystring: String): Future[String] = request("sections", querystring)
+  def getSections(querystring: String): Future[String] = request("sections", Some(querystring))
+
+  def getContent(path: String): Future[String] = request(path, None)
 }
```
KaliedaRik, line 21: Can I assume that the changes to `getTags` and `getSections` don't impact on the performance of the existing functionality to fetch and display the tag/section selectors in the Epic forms?

app/services/Athena.scala
tomrf1: I think the data used by athena comes through within 5 mins (there's some batching before it gets written to s3).
SAC then uses athena to query that data when a user double clicks a row in this page

package.json
```diff
@@ -27,22 +27,27 @@
...
     "@remirror/react": "^1.0.21",
     "@types/lodash": "^4.14.176",
     "@types/react-beautiful-dnd": "^13.0.0",
+    "date-fns": "^2.29.3",
     "immutability-helper": "^3.0.1",
     "lodash": "^4.17.21",
     "react": "^17.0.2",
     "react-beautiful-dnd": "^13.0.0",
-    "react-dom": "^16.9.0",
+    "react-dom": "^17.0.2",
```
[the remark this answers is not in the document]
tomrf1: it's currently out of sync with the react version, which is wrong and caused problems when I added the new dependency

app/services/Athena.scala
KaliedaRik: Is it worth considering a cache for this data - say 10mins? - to prevent excessively running the query when a user goes mad and clicks on several articles in a matter of seconds?

app/services/CapiService.scala
tomrf1: just checked in CODE - it's still working!

public/src/components/drawer.tsx
```diff
@@ -191,6 +194,12 @@ export default function NavDrawer(): React.ReactElement {
             <ListItemText primary="Channel Switches" />
           </ListItem>
         </Link>
+        <Link key="Super Mode" to="/super-mode" className={classes.link}>
+          <ListItem className={classes.listItem} button key="Super Mode">
+            <ListItemText primary="Epic Super Mode Dashboard" />
+            <span className={classes.super}>🦸</span>
```
KaliedaRik, line 200: Nice touch!

app/services/Athena.scala
tomrf1: the query is different for each article/time. I'm not too worried about what our users can do here - athena is cheap! On the hack day when I was testing this repeatedly I spent $0.20

public/src/components/channelManagement/superMode/articleDetails.tsx
```diff
@@ -0,0 +1,69 @@
...
+        <div className={classes.webTitle}>{webTitle}</div>
+        <div className={classes.timestampContainer}>
+          <span className={classes.timestampLabel}>Published:</span>{' '}
+          {formatDate(new Date(webPublicationDate))}
+        </div>
+        <div className={classes.timestampContainer}>
+          <span className={classes.timestampLabel}>Went super in {region}:</span>{' '}
+          {formatDate(new Date(startTimestamp))}
+        </div>
+      </div>
```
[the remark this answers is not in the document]
tomrf1: yes it is common! That sounds like a useful feature. Currently they can filter by article url in the table view, but they have to know what to look for!

public/src/components/channelManagement/superMode/articleDataChart.tsx
```diff
@@ -0,0 +1,148 @@
...
+  const classes = useStyles();
+
+  return (
+    <Dialog open={!!row} onClose={onClose} maxWidth={'lg'}>
+      {row && (
+        <>
+          <div className={classes.dialogHeader}>
+            <DialogTitle>
+              <ArticleDetails
+                url={row.url.replace('https://www.theguardian.com/', '')}
```
KaliedaRik, line 132: Would it be worth repeating this trimming for the main table - to make it look a bit tidier?

If we go to a V2, could there be some value to break the path into parts for display in separate columns in the main table - eg so user could filter articles based on `football`, `us-news`, `commentisfree`, etc?

PR #409, opened 2022-11-28
Default promos page RRCP

silvacb:
## What does this change?
Makes necessary backend and UI changes in order to enable the marketing team to update default promos file.

#### Backend
- Adds `DefaultPromosController.scala` and `DefaultPromos.scala`

#### UI
- Creates a new page in the RRCP for managing default promos.

## Screenshots

| Admin Console Before     | Admin Console After      |  New Default Promos Page   |
|-------------|-------------|------------|
image | image | image|

### DIFF
--- a/app/models/DefaultPromos.scala
+++ b/app/models/DefaultPromos.scala
@@ -4,9 +4,9 @@ import io.circe.generic.auto._
 import io.circe.{Decoder, Encoder}
 
 case class DefaultPromos(
-                          guardianWeekly: Seq[String],
-                          paper: Seq[String]
-                               )
+  guardianWeekly: Seq[String],
+  paper: Seq[String]
+)
 
 object DefaultPromos {
   implicit val decoder = Decoder[DefaultPromos]
--- a/app/models/DefaultPromos.scala
+++ b/app/models/DefaultPromos.scala
@@ -1,0 +1,14 @@
+package models
+
+import io.circe.generic.auto._
+import io.circe.{Decoder, Encoder}
+
+case class DefaultPromos(
+                          guardianWeekly: Seq[String],
+                          paper: Seq[String]
+                               )
+
+object DefaultPromos {
+  implicit val decoder = Decoder[DefaultPromos]
+  implicit val encoder = Encoder[DefaultPromos]
+}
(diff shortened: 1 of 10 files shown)


app/models/DefaultPromos.scala
```diff
@@ -0,0 +1,14 @@
+package models
+
+import io.circe.generic.auto._
+import io.circe.{Decoder, Encoder}
+
+case class DefaultPromos(
+                          guardianWeekly: Seq[String],
+                          paper: Seq[String]
+                               )
```
tjmw, line 9: The formatting looks slightly out here. It doesn't look like scalafmt is configured for this project but perhaps we could fix manually?

public/src/components/defaultPromos.tsx
```diff
@@ -0,0 +1,94 @@
...
+}));
+
+const DefaultPromos: React.FC<InnerProps<DefaultPromos>> = ({
+  data,
+  setData,
+  saveData,
+  saving,
+}: InnerProps<DefaultPromos>) => {
+  const [gwPromosString, setGwPromosString] = useState<string>(data.guardianWeekly.join(', '));
+  const [paperPromosString, setpaperPromosString] = useState<string>(data.paper.join(', '));
```
tomrf1, line 36: Are these `useState`s actually needed?
Because you're calling `setData` in the `onChange` handlers I think this would have the same effect -
```
<TextField
  value={data.guardianWeekly.join(', ')}
  ...
/>
```

tjmw: We started out with exactly that, but found interacting with the text field felt quite awkward (as it was constantly reformatted as you type, for example superfluous spaces being immediately removed).

tomrf1: actually I'm wrong, this wouldn't work!

tomrf1 (approved):
great work!

app/models/DefaultPromos.scala
silvacb: I've fixed this manually, should be good now.

public/src/components/defaultPromos.tsx
tomrf1: ah, your reply didn't appear until I sent the last message! This makes sense.
I guess the `setData` call could happen onBlur, but it probably doesn't make much difference

PR #410, opened 2022-11-28
Add search field above campaigns list

KaliedaRik:
## What does this change?
Adds a search input field above the list of campaigns on the Campaign page.

This will help users more quickly find the campaign they are interested in

## Screenshots

**Before**
Screenshot 2022-11-28 at 14 17 50

**After**
Screenshot 2022-11-28 at 17 00 08
Screenshot 2022-11-28 at 17 00 24

### DIFF
--- a/public/src/components/channelManagement/campaigns/CampaignsList.tsx
+++ b/public/src/components/channelManagement/campaigns/CampaignsList.tsx
@@ -84,7 +84,7 @@ const CampaignsList = ({
       }
       return 0;
     });
-    return [...campaignArray];
+    return campaignArray;
   };
 
   const sortedCampaigns = sortCampaigns(filterCampaigns(campaigns));
--- a/public/src/components/channelManagement/campaigns/CampaignsList.tsx
+++ b/public/src/components/channelManagement/campaigns/CampaignsList.tsx
@@ -58,29 +58,36 @@ const CampaignsList = ({
     return 'outlined';
   };
 
-  const sortedCampaigns = campaigns.filter(c => {
-    if (!campaignSearch) {
-      return true;
-    } else if (c.nickname && c.nickname.indexOf(campaignSearch) >= 0) {
-      return true;
-    } else if (c.name.indexOf(campaignSearch) >= 0) {
-      return true;
-    }
-    return false;
-  });
+  const filterCampaigns = (campaignArray: Campaigns) => {
+    return campaignArray.filter(c => {
+      if (!campaignSearch) {
+        return true;
+      } else if (c.nickname && c.nickname.indexOf(campaignSearch) >= 0) {
+        return true;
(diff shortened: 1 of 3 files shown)


public/src/components/channelManagement/campaigns/CampaignsList.tsx
```diff
@@ -56,8 +58,16 @@ const CampaignsList = ({
     return 'outlined';
   };
 
-  // When we get creation timestamps, we are thinking of allowing users to sort the list either alphabetically or by most recently created. For now, only the alphabetically sorted version is possible
-  const sortedCampaigns = [...campaigns];
+  const sortedCampaigns = campaigns.filter(c => {
```
tomrf1, line 61: The `sortedCampaigns` name is quite confusing now that we're initialising it with a filter!
Perhaps the filtering could be moved to a function and then you can do something like:
`const sortedAndFilteredCampaigns = filterCampaigns().sort((a, b)...`

KaliedaRik: I'll give it a go

KaliedaRik: The Array `sort` function mutates the array so I've separated that out to its own function alongside the new filter function. Code now self-documents what is happening here through function/variable names

tomrf1: 👍

public/src/components/channelManagement/campaigns/CampaignsList.tsx
```diff
@@ -56,21 +58,36 @@ const CampaignsList = ({
...
-  });
+      if (A < B) {
+        return -1;
+      }
+      if (B < A) {
+        return 1;
+      }
+      return 0;
+    });
+    return [...campaignArray];
```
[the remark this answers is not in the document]
KaliedaRik: Fixed

public/src/components/channelManagement/campaigns/CampaignsList.tsx
```diff
@@ -56,21 +58,36 @@ const CampaignsList = ({
     return 'outlined';
   };
 
-  // When we get creation timestamps, we are thinking of allowing users to sort the list either alphabetically or by most recently created. For now, only the alphabetically sorted version is possible
-  const sortedCampaigns = [...campaigns];
+  const filterCampaigns = (campaignArray: Campaigns) => {
+    return campaignArray.filter(c => {
+      if (!campaignSearch) {
```
tomrf1, line 63: not terribly important but if you move this `if (!campaignSearch)` check outside of the filter then the browser will not have to do pointless work every time `campaignSearch` is empty

KaliedaRik: It's a good idea!

PR #411, opened 2022-11-29
Fix RTE buttons

tomrf1:
Bug: if the last action is to click a button (e.g. currency symbol), then it does not get picked up.
Calling `.focus()` fixes this


MaelGNM (approved):

KaliedaRik (approved):
Much tidier

PR #413, opened 2022-11-30
Change banner body copy recommended length to 500

silvacb:
## What does this change?

Marketing have asked that the recommended body copy length value for banners with a secondary CTA be changed from 390 to 500.

#### Before
Screenshot 2022-12-01 at 15 23 13

#### After
Screenshot 2022-12-01 at 15 24 00


KaliedaRik (approved):
LGTM

PR #416, opened 2022-12-08
profile image validation

tomrf1:
this change leverages `react-hook-form` to require both `mainUrl` and `altText` to be defined when either is defined.
If neither is defined then we set `headshot` as undefined


KaliedaRik (approved):
Excellent improvement

PR #419, opened 2022-12-12
Use real ticker data in preview

tomrf1:
instead of some hardcoded amounts, it fetches the real data based on the configured campaign.
A new custom hook is shared by epic/banner previews


tjmw (approved):
Nice! 👍

PR #422, opened 2022-12-15
Add editable notes field to the campaigns page

KaliedaRik:
## What does this change?
Adds an editable notes field to campaigns in the campaigns page

### Screenshots
TODO

### DIFF
--- a/app/controllers/CampaignsController.scala
+++ b/app/controllers/CampaignsController.scala
@@ -54,6 +54,16 @@ class CampaignsController(
     }
   }
 
+  def updateCampaign: Action[Campaign] = authAction.async(circe.json[Campaign]) { request =>
+    run {
+      val campaign = request.body
+      logger.info(s"${request.user.email} is creating campaign '${campaign.name}'")
+      dynamoCampaigns
+        .updateCampaign(campaign)
+        .map(_ => Ok("updated"))
+    }
+  }
+
   def getTests(campaignName: String): Action[AnyContent] = authAction.async { request =>
     run {
       dynamoChannelTests.getAllTestsInCampaign(campaignName)
--- a/app/models/Campaigns.scala
+++ b/app/models/Campaigns.scala
@@ -7,7 +7,8 @@ import io.circe.{Decoder, Encoder}
 case class Campaign(
   name: String,
   nickname: String,
-  description: Option[String]
+  description: Option[String],
+  notes: Option[List[String]]
 )
 
 object Campaigns {
--- a/app/services/DynamoCampaigns.scala
+++ b/app/services/DynamoCampaigns.scala
@@ -64,4 +64,13 @@ class DynamoCampaigns(stage: String, client: DynamoDbClient) extends StrictLoggi
     put(request)
   }
 
+  def updateCampaign(campaign: Campaign): ZIO[ZEnv, DynamoError, Unit] = {
+    val item = jsonToDynamo(campaign.asJson).m()
+    val request = PutItemRequest
+      .builder
+      .tableName(tableName)
+      .item(item)
+      .build()
+    put(request)
+  }
 }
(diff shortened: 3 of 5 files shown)


public/src/components/channelManagement/campaigns/CampaignsEditor.tsx
```diff
@@ -147,12 +184,65 @@ function CampaignsEditor({ campaign }: CampaignsEditorProps): React.ReactElement
...
+                label="Description"
+                margin="normal"
+                variant="outlined"
+                disabled={!editMode}
+                fullWidth
+              />
+
+              <FormControlLabel
+                control={
+                  <Switch
```
KaliedaRik, line 207: There's a really annoying bug with the switch where, on first click, it changes the `isActive` status of the campaign in the sidebar listing, but the appearance of the switch itself doesn't change. Clicking again on the switch will update its appearance to bring it in line with the campaigns new status. 

Suggestions on how to fix very welcome as this is a poor UX.

public/src/components/channelManagement/campaigns/CampaignsForm.tsx
```diff
@@ -100,21 +118,25 @@ const CampaignsForm: React.FC = () => {
     }
   };
 
-  const currentCampaign = getSelectedCampaign();
+  const hasCampaignBeenSelected = checkForSelectedOrRequestedCampaign();
```
tomrf1, line 121: A few issues here -
1. `hasCampaignBeenSelected` is redundant, as we already have `selectedCampaign`.
2. The use of the `useParams` hook in the `checkForSelectedOrRequestedCampaign` function doesn't feel very idiomatic and worries me. Instead, I think `useParams` should be declared at the top with the `useState`s. The `checkForSelectedOrRequestedCampaign` function can then be removed, and the logic to `setSelectedCampaign` can be moved into the `useEffect` at the top (after the campaigns have been retrieved).
3. `testName` should be `campaignName`

Happy to pair on this!

KaliedaRik: Fixed as suggested

tomrf1: Nice! If you update the parameter name [here](https://github.com/guardian/support-admin-console/blob/main/public/src/main.tsx#L230) then there's no need for `testName` in this file

public/src/components/channelManagement/campaigns/CampaignsEditor.tsx
KaliedaRik: Fixed

public/src/components/channelManagement/campaigns/CampaignsForm.tsx
KaliedaRik: Done!

public/src/components/channelManagement/campaigns/CampaignsEditor.tsx
```diff
@@ -136,6 +179,19 @@ function CampaignsEditor({ campaign }: CampaignsEditorProps): React.ReactElement
     }
   };
 
+  const updateDescription = ({ description }: FormData): void => {
```
tomrf1, line 182: having separate handlers for each field is actually causing it to overwrite earlier changes, because it's fighting against react-hook-form here. Do we even need to use react-hook-form in this component?

KaliedaRik: It took some doing (MUI Switch is not a nice component!) ... but fixed. All three fields now use the the same `onSubmit` function.

KaliedaRik: For future reference - we're pretty tied to using RHF if we're using our RTE component. Also of interest, we're using Remirror v1.0.x but the repo has progressed a lot over the past year and is now on v2.0.x ... it might be worth our while to investigate whether we should upgrade sooner rather than later?

tomrf1: > we're pretty tied to using RHF if we're using our RTE component.
Why? It doesn't use RHF

KaliedaRik: I couldn't get it to work without RHF collecting the data in the parent component

tomrf1 (approved):
Great!
In a follow-up let's have a go at:
1. removing react-hook-form from the CampaignsEditor
2. only POST to the server when the Save button is clicked

PR #425, opened 2022-12-20
Enable highlightedText field for GW banner

tomrf1:
Currently RRCP hides the `highlightedText` field  if the GuardianWeekly banner template is selected.
But - the GW banner in SDC _does_ render the `highlightedText` field.
If a user does not clear the default `highlightedText` content before selecting the GuardianWeekly template then they cannot see or remove it in the tool.

Perhaps the GW banner is not supposed to support `highlightedText`, but that's a question for another day.
This PR makes RRCP consistent with the component.


KaliedaRik (approved):
Fixes the immediate issue

PR #426, opened 2022-12-21
Set asciiGraphWidth

tomrf1:
snyk is failing to pick up our overrides
see https://stackoverflow.com/questions/37412934/how-to-make-sbt-not-truncate-its-output/37478783#37478783


KaliedaRik (approved):
Let's hope this works!

PR #432, opened 2023-02-13
Ubuntu Jammy

tomrf1:
Ubuntu Bionic is approaching end of life. This PR updates the metering api to use an AMI based on Jammy.

Tested in CODE

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -8,4 +8,4 @@ addSbtPlugin("com.typesafe.sbt" % "sbt-digest" % "1.1.3")
 
 addSbtPlugin("com.localytics" % "sbt-dynamodb" % "2.0.2")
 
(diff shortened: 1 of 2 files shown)


project/plugins.sbt
```diff
@@ -8,4 +8,4 @@ addSbtPlugin("com.typesafe.sbt" % "sbt-digest" % "1.1.3")
 
 addSbtPlugin("com.localytics" % "sbt-dynamodb" % "2.0.2")
 
-libraryDependencies += "org.vafer" % "jdeb" % "1.5" artifacts (Artifact("jdeb", "jar", "jar"))
+libraryDependencies += "org.vafer" % "jdeb" % "1.10" artifacts (Artifact("jdeb", "jar", "jar"))
```
tomrf1, line 11: older versions of jdeb are now incompatible because of an issue with config paths in the package

PR #439, opened 2023-03-21
Two-tier geographical targeting for amounts tests

KaliedaRik:
## What does this change?
Marketing have asked us to supply them with the ability to set up a contribution amounts test for a given country, in addition to the tests for existing regions.

+ Note that this request represents a major refactor of the data model used to store amounts tests, which previously restricted amounts test data to just our 7 named regions.

Associated PRs in other repos
+ SAC: https://github.com/guardian/support-admin-console/pull/439
+ SDC: https://github.com/guardian/support-dotcom-components/pull/940
+ Support Frontend: https://github.com/guardian/support-frontend/pull/5027

This PR updates the RRCP `Amounts` page to allow Marketing colleagues to create, update and delete `country` tests in addition to existing `region tests`. It also fixes some existing UX issues with the existing page.

We also remove the RRCP `Contribution types` page as that data will now be recorded as part of the amounts test data model

## Navigating through the PR
This PR touches 45 files, as follows:

**/app and /conf folders** (7 files)
The changes in these Scala files handle typing, encoding and decoding the new amounts-v2 model (stored in S3). It also removes code handling the ContributionTypes model which has now been folded into the amounts-v2 model

**/public/src/components/amounts folder** (24 files)
I need to apologise here, as the number of files changed is a result of me renaming files to make their purpose more obvious. Sadly GitHub has failed to pickup that some new/deleted files were in fact an attempt to rename the file.

Nevertheless the bulk of the PR code resides here, as I adapted the RRCP amounts page to accommodate the management of 2-tier amounts tests - see screenshots below for more details.

**Other /public/src/components folders** (13 files)
Changes in these files are mainly minor edits for bugs arising from the move from a single amounts-test object model to an array of amounts-test objects model, and the removal of the RRCP `Contribution types` page.

**/public/src/utils folder** (1 file)
We define the JS types for the new amounts-v2 model here

## How to test
Testing is in 3 parts:
+ Functionality testing of updated RRCP amounts page on local machine, and in CODE
+ End-to-end testing in CODE to make sure that a country-specific amounts test displays correctly in Epics and Banners for a user in a country with an associated amounts test, that details of the correct amounts test get coded into the support site URL, and that clicking on the CTA will take the user to the support site which displays the same amounts data and choice as the user saw/made in the Epic/Banner. [Link to testing document](https://docs.google.com/document/d/1P-MS1Do_5Xqro6E6KKiw_7gID8c4U3L_CE_nAn-j2VI/edit).

#### Amounts page before changes
+ The name of the test (used for tracking) is not clear
+ The delete buttons - both for the live test, and for each variant in the test - are unclear; clicking on the delete buttons gives no warnings that the action is permanent and irrevocable
+ No visual indication on whether a test is running in a given region, except by clicking on the test and checking its `isLive` selector
+ A live test name can only be set once
+ The default contribution type, and parameters to determine which contributions types to display on the checkout page, are set elsewhere in RRCP
+ Only a single save button for all tests, with no feedback that the save has actioned.

Screenshot 2023-05-10 at 16 01 39

#### New Amounts page - initial load
+ The page no longer loads with one of the amounts tests pre-loaded
+ All amounts tests are displayed in the left hand column, with region tests shown above country tests
+ every test displays its `isLive` status by colour (live tests have red borders)
+ new `create test` button at top of the country tests list

Screenshot 2023-05-10 at 16 02 18

+ Country tests are created using an interactive selector which lists all countries for which no test currently exists

Screenshot 2023-05-10 at 16 48 40

**FOR ALL TESTS**
+ The test tracking names are displayed at the top of the form; the Live A/B test name is editable
+ The `save test` and `add variant` buttons appear twice, near the top of the form, and at the bottom
+ The `display contribution type` checkboxes are displayed at the top of each variant.
+ The `default contribution type` buttons selector is displayed at the top of each variant.
+ Each variant panel, except the `control` variant, has a clearly marked `delete variant` button at the top of the panel
+ Clicking the `delete variant` button triggers a confirm action modal to display
+ The `save test` button will be enabled after the first change is made to the test's data

Screenshot 2023-05-15 at 15 30 30

**ADDITIONALLY FOR COUNTRY-SPECIFIC TESTS**
+ The test's names are defined automatically when the test is first created; the Live test name is editable
+ A `delete test` button appears at the top and bottom of the test's form; clicking the button will launch a warning modal asking for the action to be confirmed

Screenshot 2023-05-15 at 15 30 56


app/models/Amounts.scala
```diff
@@ -1,48 +1,51 @@
...
+  amountsCardData: AmountsCardData,
 )
 
 case class AmountsTest(
-    name: String,
-    isLive: Boolean,
-    variants: List[AmountsTestVariant],
-    seed: Int,
+  testName: String,
+  liveTestName: Option[String],
```
tomrf1, line 29: what is `liveTestName` for?

app/models/EpicTest.scala
```diff
@@ -33,7 +33,7 @@ case class EpicVariant(
   separateArticleCount: Option[SeparateArticleCount],
   showChoiceCards: Option[Boolean],
   bylineWithImage: Option[BylineWithImage],
-  defaultChoiceCardFrequency: Option[ContributionType],
+  defaultChoiceCardFrequency: Option[String],
```
tomrf1, line 36: this can still have a type, `ContributionType`

public/src/components/amounts/AmountsForm.tsx
```diff
@@ -0,0 +1,174 @@
...
+  saveData: saveConfiguredAmounts,
+}: InnerProps<AmountsTests>) => {
+  const classes = useStyles();
+
+  const [selectedTest, setSelectedTest] = useState<AmountsTest | undefined>();
+
+  const onTargetSelected = (target: Territory) => {
+    const currentTest = configuredAmounts.filter(test => test.target === target);
+    if (currentTest.length) {
+      setSelectedTest(currentTest[0]);
```
tomrf1, line 62: you can use `find` instead of `filter` if you only expect 0 or 1 values

app/models/Amounts.scala
```diff
@@ -1,48 +1,51 @@
...
+  isLive: Boolean,
+  target: String,
```
tomrf1, line 31: It seems a shame to define this as a string - would it make things nicer for SDC if this type encoded whether it's a region or a country? In the old scoping doc I suggested something like `{ region: Region } | { countryCode: string }` for the TS

app/models/Amounts.scala
KaliedaRik: Discussed in chat. Unless this is a show-stopper, I'd like to keep both `testName` and `liveTestName` fields

app/models/EpicTest.scala
KaliedaRik: I deleted the old `app/models/ContributionTypes.scala` file containing the `ContributionType` definition as the contribution types page in RRCP has been deleted (as agreed). 

What I haven't done is remove this optional field from the EpicVariant definition, as all existing Epic variants still include this data even though the ability to set which contribution type (one-off, monthly, annual) should display as the default is (has long been) ignored by amounts cards in the Epic - which instead will use the amounts test variant equivalent data to set the default tab.

I have removed the ability for users to indicate which contribution amount to display when creating an epic variant as it was both redundant, and confusing UX.

public/src/components/amounts/AmountsForm.tsx
KaliedaRik: Updated as suggested

app/models/Amounts.scala
KaliedaRik: I don't understand this comment as this is a Scala file and the comment talks about TS definitions.

In TS, the target field is typed as `Territory` which is defined in the `src/utils/models.ts` file as follows:

Screenshot 2023-07-24 at 13 27 55

The `app/models/Region.scala` file defines a `Region` object but I don't know how to add code to that file to create a type that would be either a Region, or a 2-letter string:

Screenshot 2023-07-24 at 13 33 45

tomrf1: My comment is more general - whatever we do, the scala + TS models would have to align.
This field can hold two different types of thing, but currently the model doesn't convey which it is - it just ends up as a string in dynamodb. I'm wondering if it's better for the apps which read this data if they can check the type. A model like `{ region: Region } | { countryCode: string }` would allow this.

app/models/EpicTest.scala
KaliedaRik: Definitely good for removal in a future PR

app/models/Amounts.scala
tomrf1: If we were to use the above TS model, I think the scala would look something like this:

```
sealed trait AmountsTestTargeting
case class RegionTargeting(region: Region) extends AmountsTestTargeting
case class CountryTargeting(countryCode: String) extends AmountsTestTargeting
```
And for decoding it:
```
object AmountsTestTargeting {
  import io.circe.generic.auto._
  import cats.syntax.functor._  // for the widen syntax

  implicit val amountsTestTargetingDecoder: Decoder[AmountsTestTargeting] =
    Decoder[RegionTargeting].widen or Decoder[CountryTargeting].widen
}
```

tjmw: I agree with the above comments. If we can it'd be really nice to have the type system convey what type of geographical entity we're talking about as well as enforcing, whether it's a region or country code, that a valid value is passed (e.g. right now the string "foo" would keep the type system happy, even though it's neither a country code or a region).

public/src/components/amounts/AmountsVariantEditorRowInput.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+export const AmountsVariantEditorRowInput: React.FC<AmountsVariantEditorRowInputProps> = ({
+  amounts,
+  addAmount,
+}: AmountsVariantEditorRowInputProps) => {
+  const [currentValue, setCurrentValue] = useState<number | undefined>();
+  const [currentError, setCurrentError] = useState('');
+
+  const onSubmit = () => {
+    if (!currentError && currentValue != null && currentValue) {
```
tjmw, line 27: Does the `&& currentValue` here make the `currentValue != null` redundant?

public/src/components/channelManagement/campaigns/TestCard.tsx
```diff
@@ -189,7 +189,7 @@ function TestCard({ test, keyId, linkPath }: TestCardProps): React.ReactElement
   };
 
   const getLocations = (test: Test) => {
-    const locations: string[] = test.locations || [];
+    const locations: (string | number)[] = test.locations || [];
```
tjmw, line 192: Oh! When might this be a `number`?

public/src/components/amounts/AmountsForm.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+      const updatedTests = [...configuredAmounts, newTest];
+      setConfiguredAmounts(updatedTests);
+      setSelectedTest(newTest);
+    }
+  };
+
+  const updateLocalTest = (updated: AmountsTest) => {
+    const updatedTests = configuredAmounts.filter(t => t.testName !== updated.testName);
+    updatedTests.push(updated);
```
tjmw, line 115: Does the ordering here matter at all? Is it okay that the updated test always moves to the end of the array?

public/src/components/amounts/AmountsTestsList.tsx
```diff
@@ -0,0 +1,188 @@
...
+  onTargetSelected: (target: Territory) => void;
+  create: (selected: CountryOptions) => void;
+}
+
+interface AmountsTestButtonProps {
+  target: Territory;
+}
+
+const regionLabels = Object.keys(Regions);
+const countryLabels = Object.keys(Countries);
```
tjmw, line 93: What does "labels" mean here? The keys of these objects are the code I think (as opposed to something designed for humans to read).

public/src/components/amounts/AmountsTestEditor.tsx
```diff
@@ -0,0 +1,288 @@
...
+        key={`variant_key_${variant.variantName}`}
+        variant={variant}
+        updateVariant={updateVariant}
+        deleteVariant={deleteVariant}
+      />
+    );
+  };
+
+  const checkIfTestIsCountryTier = () => {
+    return target != null && !regionLabels.includes(target as string) ? true : false;
```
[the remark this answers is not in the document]
KaliedaRik: Probably? I can update and test. Though I do have a personal preference for defined (as in: obvious) rather than implied return values

public/src/components/amounts/AmountsForm.tsx
KaliedaRik: No, ordering is not an issue here. It's just an array of tests

public/src/components/channelManagement/campaigns/TestCard.tsx
KaliedaRik: The change fixes a typescript error. Given that the campaigns page functionality is currently broken due to us having to separate out archived tests into their own DynamoDB table, and I don't think Marketing are using the Campaigns functionality as we expected them to, thus needs review, I think we can live with this in the short-term?

public/src/components/amounts/AmountsVariantEditorRowInput.tsx
KaliedaRik: Fixed

public/src/components/amounts/AmountsTestEditor.tsx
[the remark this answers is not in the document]
KaliedaRik: Fixed

app/models/Amounts.scala
tjmw: Any thoughts on this one?

public/src/components/channelManagement/campaigns/TestCard.tsx
tjmw: Sure thing, if you're happy that's all good! (I'm still not sure when `test.locations` will include a `number`, but if it's a short term thing I'm sure we can live with it).

public/src/components/amounts/AmountsForm.tsx
tjmw: 👍  As mentioned elsewhere, if order isn't important then perhaps an object indexed by some unique identifier (`testName`?) would be a better choice over an array? It'd make lookups more straightforward too. Perhaps one for a future PR if that's out of scope here.

public/src/components/amounts/AmountsForm.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+  const saveLocalTestToS3 = () => {
+    saveConfiguredAmounts();
+  };
+
+  const getAllTestNames = () => {
+    const namesArray: string[] = [];
+    configuredAmounts.forEach(t => {
+      if (t.testName === selectedTest?.testName) {
+        namesArray.push(t.testName);
```
tjmw, line 133: This line is duplicated in the `if` and `else` blocks. Is that intentional? If it is perhaps it should be moved outside of the if/else?

tjmw: Also seems like a possible nice use case for `flatMap`?

```js
const getAllTestNames = () =>
  configuredAmounts.flapMap(t => {
    if (t.liveTestName) {
      return [t.testName, t.liveTestName];
    } else {
      return [t.testName];
    }
  });
```

public/src/components/amounts/AmountsTestsList.tsx
```diff
@@ -0,0 +1,188 @@
...
+  };
+
+  const getCountryTestCandidates = () => {
+    const existingCountryTests = getExistingCountryTests();
+    const potentialTests = countryLabels.filter(l => !existingCountryTests.includes(l));
+    return potentialTests.sort();
+  };
+
+  const getButtonStyling = (target: Territory) => {
+    const myTest = tests.filter(t => target === t.target)[0];
```
tjmw, line 132: Would `tests.find` be better here?

PR #446, opened 2023-04-24
Remove _2022 from ticker names

undergroundquizscene:
Rather than add a new ticker each year, it seems more sensible to reuse the existing ticker names. Therefore, removing the year from the name makes sense. This will need to be co-ordinated with the similar PRs on support-dotcom-components and contributions-ticker-calculator.

### DIFF
--- a/public/src/components/channelManagement/tickerEditor.tsx
+++ b/public/src/components/channelManagement/tickerEditor.tsx
@@ -43,7 +43,7 @@ const DEFAULT_TICKER_SETTINGS: TickerSettings = {
     goalReachedSecondary: 'but you can still support us',
   },
   currencySymbol: '$',
-  name: TickerName.US_2022,
+  name: TickerName.US,
 };
 
 interface TickerEditorProps {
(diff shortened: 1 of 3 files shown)


public/src/components/channelManagement/tickerEditor.tsx
```diff
@@ -149,13 +149,25 @@ const TickerEditor: React.FC<TickerEditorProps> = ({
                 <FormControlLabel
                   value={TickerName.US_2022}
                   control={<Radio />}
-                  label="US_2022"
+                  label="US_2022 (do not use)"
```
tomrf1, line 152: do we need to include the old ones here?

undergroundquizscene: I’m not certain, but I have stopped looking at this for now. Will investigate when I revisit.

undergroundquizscene: The tickers are not currently in use, so I’ve removed the old names now

undergroundquizscene:
Co-ordinated with https://github.com/guardian/support-dotcom-components/pull/886

tomrf1 (approved):
👍 thanks for doing this!

undergroundquizscene:
And https://github.com/guardian/contributions-ticker-calculator/pull/53, of course!

undergroundquizscene:
I should note that there are some archived tests (43 as far as I can tell) that still reference the old names: if unarchiving those tests we’ll need to update the names before they’re visible in the RRCP (which will currently drop tests referring to unknown tickers).

PR #447, opened 2023-04-25
Node 18

tomrf1:
The node update caused `npm install` to fail because of some incompatible peer dependencies that were previously not flagged as a problem. Setting overrides 'fixes' this for now.
It was also necessary to upgrade webpack


package.json
```diff
@@ -52,6 +53,11 @@
   "resolutions": {
     "csstype": "^2.6.18"
   },
+  "overrides": {
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
```
tomrf1, line 58: these react overrides are for the benefit of `react-beautiful-dnd` (a drag-and-drop library). This is no longer actively maintained, so we should consider removing/replacing it

package.json
```diff
@@ -52,6 +53,11 @@
...
+    "prettier": "^1.19.1"
```
tomrf1, line 59: This isn't necessary if we update our `prettier` dependency, but newer versions want to reformat the entire codebase so I didn't want to do this now

package.json
```diff
@@ -4,9 +4,10 @@
     "@types/material-ui": "^0.21.7",
     "@types/react-dom": "^16.9.0",
     "@types/react-router-dom": "^5.0.1",
+    "@types/react": "^17.0.58",
```
tjmw, line 7: Should this be `"~17.0.58"`? I think these types libraries guarantee compatibility within the minor version?

package.json
tjmw: Earlier on when we were looking at this I was translating `dnd` => dungeons and dragons 😄

tjmw (approved):
Nice!

PR #456, opened 2023-06-14
Use channel-status-index

tomrf1:
We've started to see an issue in the epic tool where some tests are not being returned by Dynamodb.
tl;dr - this is because we now have so many archived epic tests that the sdk wants to paginate.

### Existing behaviour:
When we query for all Live + Draft tests for a channel, the query looks like this:
```
.keyConditionExpression("channel = :channel")
...
.filterExpression("#status <> :archived")
```
Here we're using the primary index, which is on `(channel, name)`. So the result of the query is _all_ epic tests, and the filter expression is then applied by Dynamodb to filter out tests with `status` of `Archived`. So while the query on `channel` is efficient, Dynamodb has to then scan all epic tests to do the `status` filter.
Once the size of the data to be scanned exceeds a limit, the Dynamodb sdk [requires us to paginate](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Scan.html#Scan.Pagination), which is why we're not getting all results.

### New behaviour:
This PR adds a new secondary index on `(channel, status)`, which we can now query. This means Dynamodb doesn't have to perform a scan on the query results.

### Result:
While testing this locally I logged:
- count - the number of items returned by Dynamodb
- scannedCount - the number of items that had to be scanned by Dynamodb after querying
- lastEvaluatedKey - a value returned by Dynamodb if it's having to paginate

Before:
```
count: 20
scannedCount: 267
lastEvaluatedKey: {name=AttributeValue(S=PP_V3__ABORTION_US), channel=AttributeValue(S=Epic)}
```

After:
```
count: 20
scannedCount: 20
lastEvaluatedKey: {}
```

TODO - change SDC and apple-news to use the new secondary index when fetching channel tests from Dynamodb

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,13 @@ Refresh automatically:
 npm run watch
 ```
 
+### Running scala tests
+The scala backend tests use dynamodb-local. This doesn't support Apple Silicon (M1).
+
+If while running `sbt test` you get the error `cannot load library: java.lang.UnsatisfiedLinkError: no sqlite4java-osx-aarch64`, use this work around:
+1. download the correct `.dylib` from e.g. https://repo1.maven.org/maven2/io/github/ganadist/sqlite4java/libsqlite4java-osx-arm64/1.0.392/libsqlite4java-osx-arm64-1.0.392.dylib
+2. copy it into the `dynamodb-local/DynamoDBLocal_lib/` directory in this project
+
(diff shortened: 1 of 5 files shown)


tomrf1:
Closing this, as we've decided to migrate to a new archived tests table instead

PR #457, opened 2023-06-14
Create archived tests table

tomrf1:
We need to move archived tests into a new table.
This PR creates a new dynamodb table and gives the server read/write permissions

### DIFF
--- a/cdk/lib/admin-console.ts
+++ b/cdk/lib/admin-console.ts
@@ -84,6 +84,31 @@ export class AdminConsole extends GuStack {
     return table;
   }
 
+  buildArchivedTestsTable(): Table {
+    const id = `ArchivedChannelTestsDynamoTable`;
+
+    const table = new Table(this, id, {
+      tableName: `support-admin-console-archived-channel-tests-${this.stage}`,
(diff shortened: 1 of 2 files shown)


cdk/lib/admin-console.ts
```diff
@@ -84,6 +84,31 @@ export class AdminConsole extends GuStack {
...
+        type: AttributeType.STRING,
+      },
+      sortKey: {
+        name: 'name',
+        type: AttributeType.STRING,
+      },
+    });
+
+    // Give it a better name
+    const defaultChild = table.node.defaultChild as unknown as CfnElement;
```
tjmw, line 106: Out of interest, what doesn't TS like about this? What type does it think `defaultChild` has without the casting?

tjmw (approved):

cdk/lib/admin-console.ts
tomrf1: It's an `IConstruct`, which `CfnElement` also extends

tjmw: Got it, thanks. But `IConstruct ` doesn't have `.overrideLogicalId`?

PR #459, opened 2023-06-14
Archived tests migration 

tomrf1:
Following on from [the creation of an archived tests table](https://github.com/guardian/support-admin-console/pull/457), this PR adds a temporary endpoint for migrating archived tests over.

I've added it as an endpoint because it's reusing logic on the server, and so this is simpler than building a standalone program.

The endpoint is e.g. `/migrate-archived-tests/Epic`, and it:
1. gets all archived tests for that channel from the old table
2. copies them to the new archived tests table
3. does not delete them from the old table. This can be done manually from the dynamodb console once we're happy

### DIFF
--- a/app/models/Channel.scala
+++ b/app/models/Channel.scala
@@ -1,0 +1,20 @@
+package models
+
+import io.circe.generic.extras.Configuration
+import io.circe.generic.extras.semiauto.{deriveEnumerationDecoder, deriveEnumerationEncoder}
+
+sealed trait Channel
+
+object Channel {
+  case object Epic extends Channel
+  case object EpicAMP extends Channel
+  case object EpicAppleNews extends Channel
+  case object EpicLiveblog extends Channel
+  case object Banner1 extends Channel
+  case object Banner2 extends Channel
+  case object Header extends Channel
+
+  implicit val customConfig: Configuration = Configuration.default.withDefaults
+  implicit val statusEncoder = deriveEnumerationEncoder[Channel]
+  implicit val statusDecoder = deriveEnumerationDecoder[Channel]
+}
--- a/app/services/DynamoArchivedChannelTests.scala
+++ b/app/services/DynamoArchivedChannelTests.scala
@@ -1,0 +1,64 @@
+package services
+
+import com.typesafe.scalalogging.StrictLogging
+import models.DynamoErrors._
+import models._
+import software.amazon.awssdk.services.dynamodb.DynamoDbClient
+import software.amazon.awssdk.services.dynamodb.model._
+import zio.blocking.effectBlocking
+import zio.duration.durationInt
+import zio.stream.ZStream
+import zio.{ZEnv, ZIO}
+
+import scala.jdk.CollectionConverters._
+
+
+class DynamoArchivedChannelTests(stage: String, client: DynamoDbClient) extends StrictLogging {
+
+  private val tableName = s"support-admin-console-archived-channel-tests-$stage"
+
+  // Sends a batch of write requests, and returns any unprocessed items
+  private def putAll(writeRequests: List[WriteRequest]): ZIO[ZEnv, DynamoPutError, List[WriteRequest]] =
+    effectBlocking {
+      val batchWriteRequest =
+        BatchWriteItemRequest
+          .builder
+          .requestItems(Map(tableName -> writeRequests.asJava).asJava)
+          .returnConsumedCapacity(ReturnConsumedCapacity.TOTAL)
+          .build()
+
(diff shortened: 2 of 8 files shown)


app/services/DynamoArchivedChannelTests.scala
```diff
@@ -0,0 +1,64 @@
...
+import software.amazon.awssdk.services.dynamodb.model._
+import zio.blocking.effectBlocking
+import zio.duration.durationInt
+import zio.stream.ZStream
+import zio.{ZEnv, ZIO}
+
+import scala.jdk.CollectionConverters._
+
+
+class DynamoArchivedChannelTests(stage: String, client: DynamoDbClient) extends StrictLogging {
```
tomrf1, line 16: The `putAll` and `putAllBatched` are copied from the existing `DynamoChannelTests.scala`. This feature is temporary so it's not worth a refactor

app/models/Channel.scala
```diff
@@ -0,0 +1,20 @@
+package models
+
+import io.circe.generic.extras.Configuration
+import io.circe.generic.extras.semiauto.{deriveEnumerationDecoder, deriveEnumerationEncoder}
+
+sealed trait Channel
```
tomrf1, line 6: I moved this from `ChannelTests.scala` as I was getting weird circe issues when importing it from there, and the simplest solution is to split the file up

app/services/DynamoArchivedChannelTests.scala
```diff
@@ -0,0 +1,64 @@
...
+        BatchWriteItemRequest
+          .builder
+          .requestItems(Map(tableName -> writeRequests.asJava).asJava)
+          .returnConsumedCapacity(ReturnConsumedCapacity.TOTAL)
+          .build()
+
+      val result = client.batchWriteItem(batchWriteRequest)
+      logger.info(s"BatchWriteItemResponse: $result")
+
+      result.unprocessedItems().asScala
```
tjmw, line 33: Just out of interest - why might there be `unprocessedItems` here if `putAllBatched` (which calls this) is splitting into batches of 25 alread?

app/services/DynamoChannelTests.scala
```diff
@@ -77,6 +77,25 @@ class DynamoChannelTests(stage: String, client: DynamoDbClient) extends StrictLo
...
+          .tableName(tableName)
+          .keyConditionExpression("channel = :channel")
+          .expressionAttributeValues(Map(
+            ":channel" -> AttributeValue.builder.s(channel.toString).build,
+            ":archived" -> AttributeValue.builder.s("Archived").build
+          ).asJava)
+          .expressionAttributeNames(Map(
+            "#status" -> "status"
+          ).asJava)
+          .filterExpression("#status = :archived")
```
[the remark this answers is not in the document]
tjmw: What's the interaction between this line and the `expressionAttributeValues` call above?

```scala
          .expressionAttributeValues(Map(
            ":channel" -> AttributeValue.builder.s(channel.toString).build,
            ":archived" -> AttributeValue.builder.s("Archived").build
          ).asJava)
```

tjmw (approved):
Looks good! Nice solution to build a new endpoint to handle this.

I left a couple of questions just to help me understand things better but definitely no blockers!

app/services/DynamoArchivedChannelTests.scala
tomrf1: Good question - it's in case of errors. From the docs:
>>  If any requested operations fail because the table's provisioned throughput is exceeded or an internal processing failure occurs, the failed operations are returned in the UnprocessedItems response parameter.
https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html

app/services/DynamoChannelTests.scala
tjmw, line 94: What's the interaction between this line and the `expressionAttributeValues` call above?

```scala
          .expressionAttributeValues(Map(
            ":channel" -> AttributeValue.builder.s(channel.toString).build,
            ":archived" -> AttributeValue.builder.s("Archived").build
          ).asJava)
```

tomrf1: It's not the clearest api!

`expressionAttributeValues` is how you define the values you'd like to reference in the query expressions. So here we define `:archived` (with value of `"Archived"`), which we can then compare against in the filter.

`expressionAttributeNames` creates an alias for the `status` field. This is only necessary because `status` is a reserved word in dynamodb, so we have to alias it to `#status`!

`filterExpression` is the filter to be applied after the query. This expression compares the aliased `status` field against the value `"Archived"`

PR #462, opened 2023-06-20
Enable primary cta for Apple News

tomrf1:
Currently users cannot configure the primary cta.
However, it was quietly setting the primary cta with the default values ("Continue").

apple-news currently uses a hardcoded value, but will soon be using the configured value from RRCP


KaliedaRik (approved):
LGTM

PR #463, opened 2023-06-20
Add switch addition/removal functionality to the RRCP

LAKSHMIRPILLAI:
## What does this change?

Currently, to add a switch to or remove a switch from the [RRCP](https://support.gutools.co.uk/), one must edit the underlying JSON file [switches_v2.json](https://s3.console.aws.amazon.com/s3/object/support-admin-console?prefix=PROD%2Fswitches_v2.json&region=eu-west-1#). This task is to enable switch addition/removal directly in the RRCP.

Since support-frontend explicitly models the switches in [Switches.scala](https://github.com/guardian/support-frontend/blob/main/support-frontend/app/admin/settings/Switches.scala), it would probably be worthwhile asking the user for confirmation that they’ve removed a switch from that file before allowing them to delete it from the RRCP (otherwise support-frontend will break).

Also there is a small change in the Switches page layout in RRCP.

### Before Change

image

### After Change

image

### Add New Switch

Steps
1.Enter new switch details into Switch name and Description textbooks
2.The unsaved changes would be shown as an alert-"Added Testing New Switch in Feature switches"
3.Click on save to confirm the changes

Screenshot 2023-06-21 at 09 59 51

### Toggle Switch State 
1.Toggle the switch state 
2.The unsaved changes would be shown as an alert-"Turned On: Testing New Switch in Feature switches / Turned Off: Testing New Switch in Feature switches"
3.Click on save to confirm the changes

Screenshot 2023-06-21 at 10 00 16

### Remove Switch

Steps
1.Click the delete button next to the Switch to be removed 
2.User confirmation dialog opens up to check the remove action 
3.The unsaved changes would be shown as an alert-"Removed Testing New Switch in Feature switches" 
4.Click on save to confirm the changes,Otherwise refresh the page to undo the changes.

Screenshot 2023-06-21 at 10 12 58

Screenshot 2023-06-21 at 10 13 04

Removed the below part from the Switches Page
image

### DIFF
--- a/public/src/components/switchboard.tsx
+++ b/public/src/components/switchboard.tsx
@@ -106,10 +106,6 @@ const useStyles = makeStyles(({ palette, spacing }: Theme) => ({
     display: 'flex',
     justifyContent: 'space-evenly',
   },
-  inputSpacing: {
-    // padding: '10px 12px',
-    // float: 'left',
-  },
   switchLayout: {
     display: 'flex',
     justifyContent: 'space-between',
@@ -137,8 +133,7 @@ const Switchboard: React.FC<InnerProps<SupportFrontendSwitches>> = ({
     return (
       pendingChanges.length > 0 && (
         <Alert severity="warning">
-          Switch settings have been changed. Changes need to be saved before they take
-          effect! Refresh the page to undo the changes.
+          Switch settings have been changed. Changes need to be saved before they take effect! Refresh the page to undo the changes.
           <List>
             {pendingChanges.map((change, index) => (
               <ListItem key={index}>
--- a/public/src/components/switchboard.tsx
+++ b/public/src/components/switchboard.tsx
@@ -100,10 +100,7 @@ const useStyles = makeStyles(({ palette, spacing }: Theme) => ({
     marginBottom: spacing(2),
   },
   input: {
-    display: 'inline-block',
     width: '33%',
-    padding: '0.2em',
-    margin: '0.1em 0.2em',
   },
   inputGroup: {
     marginTop: spacing(2),
@@ -111,8 +108,8 @@ const useStyles = makeStyles(({ palette, spacing }: Theme) => ({
     justifyContent: 'space-evenly',
   },
   inputSpacing: {
-    padding: '10px 12px',
-    float: 'left',
+    // padding: '10px 12px',
+    // float: 'left',
   },
   switchLayout: {
     display: 'flex',
@@ -135,21 +132,13 @@ const Switchboard: React.FC<InnerProps<SupportFrontendSwitches>> = ({
 }: InnerProps<SupportFrontendSwitches>) => {
   const classes = useStyles();
 
-  const [needToSaveDataWarning, setNeedToSaveDataWarning] = useState(false);
-  const [newUnsavedData, setNewUnsavedData] = useState(false);
   const [pendingChanges, setPendingChanges] = useState<string[]>([]);
-
-  type FormData = {
-    switchId: string;
-    description: string;
-  };
 
   const displayNeedToSaveDataWarning = (): JSX.Element | false => {
     return (
-      needToSaveDataWarning &&
-      newUnsavedData && (
+      (pendingChanges.length > 0)  && (
         <Alert severity="warning">
-          Switch settings have been changed. Changes need to be saved before they take effect!
+          Switch settings have been changed. Changes need to be saved before they take effect!Refresh the page to undo the changes.
           <List>
             {pendingChanges.map((change, index) => (
               <ListItem key={index}>
@@ -163,7 +152,7 @@ const Switchboard: React.FC<InnerProps<SupportFrontendSwitches>> = ({
   };
 
   const setPendingChange = (changeName: string, description: string, groupId: string): void => {
-    const unsavedChange = changeName + ' ' + description + ' in ' + groupId;
+    const unsavedChange = `${changeName} “${description}” in “${groupId}”`;
     // Add the changeName to the list of pendingChanges
     setPendingChanges(prevChanges => [...prevChanges, unsavedChange]);
   };
@@ -179,13 +168,11 @@ const Switchboard: React.FC<InnerProps<SupportFrontendSwitches>> = ({
     updatedState[groupId].switches[switchId].state = isChecked ? SwitchState.On : SwitchState.Off;
     const currentSwitchState = updatedState[groupId].switches[switchId].state;
     setData(updatedState);
-    setNewUnsavedData(true);
     setPendingChange(
       'Turned ' + currentSwitchState + ':',
       switchData.description,
       groupData.description,
     );
-    setNeedToSaveDataWarning(true);
   };
 
   const createSwitchesFromGroupData = (
@@ -207,7 +194,7 @@ const Switchboard: React.FC<InnerProps<SupportFrontendSwitches>> = ({
         />
         <IconButton
           onClick={(): void => actionRemoveSwitchData(group, switchId, switchData.description)}
-          aria-label="removeSwitch"
+          aria-label="Remove Switch"
         >
           <DeleteIcon />
         </IconButton>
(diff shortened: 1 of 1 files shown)


public/src/components/switchboard.tsx
```diff
@@ -82,6 +99,28 @@ const useStyles = makeStyles(({ palette, spacing }: Theme) => ({
   textParagraph: {
     marginBottom: spacing(2),
   },
+  input: {
+    display: 'inline-block',
```
undergroundquizscene, line 103: I don’t really know what the `inline-block` display mode does: what’s the reason for using it here?

public/src/components/switchboard.tsx
```diff
@@ -97,61 +136,164 @@ const Switchboard: React.FC<InnerProps<SupportFrontendSwitches>> = ({
   const classes = useStyles();
 
   const [needToSaveDataWarning, setNeedToSaveDataWarning] = useState(false);
+  const [newUnsavedData, setNewUnsavedData] = useState(false);
```
undergroundquizscene, line 139: `setPendingChanges` and `setNeedToSaveDataWarning` are called everywhere `setNewUnsavedData` is called. Could we get rid of these two `useStates`, and check if there are any pending changes instead? (I.e. on line 150 below we could replace `needToSaveDataWarning && newUnsavedData &&` with `(pendingChanges.length > 0) &&`.)

public/src/components/switchboard.tsx
```diff
@@ -97,61 +136,164 @@ const Switchboard: React.FC<InnerProps<SupportFrontendSwitches>> = ({
...
+  const [pendingChanges, setPendingChanges] = useState<string[]>([]);
+
+  type FormData = {
+    switchId: string;
+    description: string;
+  };
```
undergroundquizscene, line 145: Could we move this type down to where it’s used?

public/src/components/switchboard.tsx
```diff
@@ -97,61 +136,164 @@ const Switchboard: React.FC<InnerProps<SupportFrontendSwitches>> = ({
...
+          label={switchData.description}
+          checked={switchData.state === 'On'}
+          control={<SwitchUI />}
+          onChange={(): void => updateSwitchSetting(switchId, switchData, group, !isChecked)}
+          value={switchId}
+          key={switchId}
+        />
+        <IconButton
+          onClick={(): void => actionRemoveSwitchData(group, switchId, switchData.description)}
+          aria-label="removeSwitch"
```
undergroundquizscene, line 210: From my quick reading of [the MDN docs for aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label), it sounds like this should probably be a description a screen reader can read out, rather than an ID label. I.e. I think this would be better with spaces:

```suggestion
          aria-label="Remove switch"
```

public/src/components/switchboard.tsx
```diff
@@ -97,61 +136,164 @@ const Switchboard: React.FC<InnerProps<SupportFrontendSwitches>> = ({
...
+      setPendingChange('Added', description, groupData.description);
+      setNeedToSaveDataWarning(true);
+    };
 
     return (
       <FormControl className={classes.formControl} key={groupId}>
-        <FormLabel>{groupData.description}</FormLabel>
+        <FormLabel>
+          {' '}
+          <strong>{groupData.description} </strong>
```
undergroundquizscene, line 239: Are these spaces meant to be there?

```suggestion
          <strong>{groupData.description}</strong>
```

public/src/components/switchboard.tsx
```diff
@@ -196,24 +365,7 @@ const Switchboard: React.FC<InnerProps<SupportFrontendSwitches>> = ({
 
       {displayNeedToSaveDataWarning()}
       <SaveButton />
-
       <Divider variant="fullWidth" className={classes.divider} />
```
undergroundquizscene, line 368: We could probably remove the Divider as well, since there’s now nothing below it

public/src/components/switchboard.tsx
```diff
@@ -97,61 +136,164 @@ const Switchboard: React.FC<InnerProps<SupportFrontendSwitches>> = ({
...
+        <br />
         {Object.entries(groupData.switches)
           .sort(sortByDescription)
           .map(([switchId, switchData]) =>
-            createSwitchesFromGroupData(switchId, switchData, groupId),
+            createSwitchesFromGroupData(switchId, switchData, group),
           )}
+        <div className={classes.inputGroup}>
+          <TextField
+            className={classes.input}
```
undergroundquizscene, line 249: At the moment, these text fields don’t have a screen-reader-accessible description (tested by tabbing through the page with VoiceOver turned on). From [the material-ui/core docs](https://mui.com/material-ui/react-text-field/#accessibility), it looks like this is because we haven’t given the text fields IDs. To make unique IDs we can use the switch group ID:

```suggestion
            className={classes.input}
            id={groupId + '-add-switch-switch-name'}
```

public/src/components/switchboard.tsx
```diff
@@ -97,61 +136,164 @@ const Switchboard: React.FC<InnerProps<SupportFrontendSwitches>> = ({
...
+            name="switchId"
+            label="Switch name"
+            margin="normal"
+            variant="outlined"
+            autoFocus
+            fullWidth
+          />
+          <span className={classes.inputSpacing} />
+          <TextField
+            className={classes.input}
```
undergroundquizscene, line 267:
```suggestion
            className={classes.input}
            id={groupId + '-add-switch-switch-description'}
```

public/src/components/switchboard.tsx
```diff
@@ -97,61 +136,164 @@ const Switchboard: React.FC<InnerProps<SupportFrontendSwitches>> = ({
...
+            label="Description"
+            margin="normal"
+            variant="outlined"
+            autoFocus
+            fullWidth
+          />
+          <span className={classes.inputSpacing} />
+          <div className={classes.buttons}>
+            <Button
+              variant="contained"
```
undergroundquizscene, line 288: The add button also doesn’t have any label, which we can add with aria-label (since we don’t have the text anywhere in the UI already):

```suggestion
              aria-label="Add switch"
              variant="contained"
```

public/src/components/switchboard.tsx
```diff
@@ -82,6 +99,28 @@ const useStyles = makeStyles(({ palette, spacing }: Theme) => ({
...
+  },
+  inputGroup: {
+    marginTop: spacing(2),
+    display: 'flex',
+    justifyContent: 'space-evenly',
+  },
+  inputSpacing: {
+    padding: '10px 12px',
+    float: 'left',
+  },
```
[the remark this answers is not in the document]
LAKSHMIRPILLAI: These were the initial change made before the layout change using justifyContent.Have taken them out

public/src/components/switchboard.tsx
```diff
@@ -82,6 +98,25 @@ const useStyles = makeStyles(({ palette, spacing }: Theme) => ({
...
+  inputGroup: {
+    marginTop: spacing(2),
+    display: 'flex',
+    justifyContent: 'space-evenly',
+  },
+  inputSpacing: {
+    // padding: '10px 12px',
+    // float: 'left',
+  },
```
undergroundquizscene, line 112: Could remove this class now

public/src/components/switchboard.tsx
```diff
@@ -54,10 +65,15 @@ const useStyles = makeStyles(({ palette, spacing }: Theme) => ({
     paddingLeft: spacing(2),
     paddingRight: spacing(2),
     border: `1px solid ${palette.grey['300']}`,
+    width: '45%',
```
undergroundquizscene, line 68: At small breakpoints (below about 900px) this doesn’t look great anymore. It might be worth us adding a media query at whatever our closest breakpoint to that width is, to make the width 100% below that point.

undergroundquizscene (approved):
Two minor suggestions – up to you whether you want to follow them or go ahead and merge :)

PR #465, opened 2023-07-06
Fix amounts in preview

tomrf1:
It was broken in the epic preview because the model in SDC has changed, and we've never had it in the banner.
Uses hardcoded amounts.

Screenshot 2023-07-06 at 08 20 23
Screenshot 2023-07-06 at 08 34 17


KaliedaRik (approved):
Many thanks for spotting, and fixing, this!

PR #469, opened 2023-07-12
Upgrade cdk

tomrf1:
In response to this dependabot alert about `parse-url`: https://github.com/guardian/support-admin-console/security/dependabot/47

Upgrading `@guardian/cdk` fixes this by bringing in `parse-url@^8.1.0`.
The cdk snapshot is now quite different.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.ts.snap linguist-generated=true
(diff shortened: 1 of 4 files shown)


.gitattributes
```diff
@@ -0,0 +1 @@
+*.ts.snap linguist-generated=true
```
tomrf1, line 1: tells github to collapse the `snap` diff by default

tjmw (approved):
I don't completely understand all the diffs in the snapshot, but assuming it deploys/works fine on CODE I won't worry about it too much!

PR #472, opened 2023-08-03
Basic Design UI: Part 2 (Creating and locking designs)

tjmw:
## What does this change?

More baby steps with the design UI.

This adds the ability to create a banner design. The only field specifiable in the UI currently is the name (which must be unique).

When a new banner design is created it's opened in an editable state and is not yet saved to the server. From here you can either save or discard. I've added an `imageUrl` field with a default value, it's not yet configurable in the UI. I had to do this as the update statement we create for dynamo assumes there will be at least one other field to set.

You can also click a banner design to edit it, though there are no editable fields yet, this puts it into a locked state until you click save or discard.

Lots of this code has been duplicated from the existing tests code. In future there may be refactoring to consolidate some of the shared logic.

Screen Recording 2023-08-03 at 13 13 56 mov

## How to test

Run locally (or on CODE).

## Accessibility

### DIFF
--- a/app/controllers/banner/BannerDesignsController.scala
+++ b/app/controllers/banner/BannerDesignsController.scala
@@ -33,6 +33,7 @@ class BannerDesignsController(
 
   case class BannerDesignsResponse(
       bannerDesigns: List[BannerDesign],
+      userEmail: String,
   )
 
   val lockFileName = "banner-designs"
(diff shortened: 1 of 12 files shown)


KaliedaRik (approved):
Good to see the basic functionality working!

PR #474, opened 2023-08-15
Banner Design UI: Part 3 (First configurable field - imageUrl)

tjmw:
## What does this change?

This PR adds a first configurable field to a banner design (beyond the existing name field). It is a required field and we provide a default when creating a new design. It is editable for existing designs too. The UI will show an inline error message if the user clears the field, and will show an alert if the user attempts to save a design with an empty image URL field. In this PR validation is based on presence only (i.e. it isn't validated for being a URL, we could add this in future).

Screen Recording 2023-08-15 at 16 06 56 mov

## Accessibility


KaliedaRik (approved):
LGTM

PR #476, opened 2023-08-18
Banner Design UI: Part 5 (Banner variants have a template or a design)

tjmw:
## What does this change?

This PR updates the banner variant model to have either a template (as it does now) or a banner design. When creating/editing a banner variant you will be able to choose between selecting a variant or selecting a design:

Screen Recording 2023-08-25 at 15 31 55 mov

This toggling is hidden behind a feature flag in CODE and PROD, so it isn't actually possible to give a banner variant a design in these environments. This is a good thing because currently SDC wouldn't know what to do in the case where a banner variant has a design, so would break.

The banner variant model has been extended in a backwards compatible way which I think means this change is safe to deploy. The TS variant model has been updated to go from having this:

```ts
template: BannerTemplate
```

to having this:

```ts
template: BannerTemplate | BannerDesignName
```

where `BannerDesignName` looks like this:

```ts
export interface BannerDesignName {
  designName: string;
}
```

So for all existing cases, where a `BannerTemplate` is specified, things will continue to work as they are.

On the server when creating/updating variants there isn't any validation currently that the banner design specified actually exists. I'm intending to add this in a future PR.

Note: previewing doesn't currently work for variants with a design (and won't work until SDC has been updated).

## Accessibility

### DIFF
--- a/public/src/components/channelManagement/bannerTests/bannerTemplateSelector.tsx
+++ b/public/src/components/channelManagement/bannerTests/bannerTemplateSelector.tsx
@@ -1,13 +1,9 @@
 import React from 'react';
 import { MenuItem, Select } from '@material-ui/core';
-import { BannerTemplate } from '../../../models/banner';
-
-function isBannerTemplate(s: string): s is BannerTemplate {
-  return Object.values(BannerTemplate).includes(s as BannerTemplate);
-}
+import { BannerTemplate, BannerUI, isBannerTemplate } from '../../../models/banner';
 
 interface BannerTemplateSelectorProps {
-  template: BannerTemplate;
+  template: BannerUI;
   onTemplateChange: (updatedTemplate: BannerTemplate) => void;
   editMode: boolean;
 }
--- a/public/src/components/channelManagement/bannerTests/bannerUiSelector.tsx
+++ b/public/src/components/channelManagement/bannerTests/bannerUiSelector.tsx
@@ -16,6 +16,7 @@ import {
   uiIsDesign,
 } from '../../../models/banner';
 import { BannerDesign } from '../../../models/BannerDesign';
+import { shouldShowBannerDesignsFeature } from '../../../utils/features';
 
 interface BannerUiSelectorProps {
   ui: BannerUi;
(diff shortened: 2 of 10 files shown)


public/src/components/channelManagement/bannerTests/bannerTemplateSelector.tsx
```diff
@@ -1,74 +0,0 @@
-import React from 'react';
```
tjmw, line 1: This has been renamed to `BannerUiSelector` bit GH is showing as a deletion and creation.

public/src/components/channelManagement/bannerTests/bannerVariantPreview.tsx
```diff
@@ -226,10 +231,18 @@ const BannerVariantPreview: React.FC<BannerVariantPreviewProps> = ({
...
-    `banners/${bannerModules[variant.template].path}`,
-    bannerModules[variant.template].name,
-  );
+  const moduleConfig = isBannerTemplate(variant.template)
+    ? {
+        path: `banners/${bannerModules[variant.template].path}`,
+        name: bannerModules[variant.template].name,
+      }
+    : {
+        // TODO: fixme!
```
tjmw, line 240: Previewing won't work until SDC has been updated to handle banner designs.

public/src/components/channelManagement/bannerTests/bannerUiSelector.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+interface BannerUiSelectorProps {
+  ui: BannerUi;
+  onUiChange: (updatedUi: BannerUi) => void;
+  editMode: boolean;
+  designs: BannerDesign[];
+}
+
+const templatesWithLabels = [
+  { template: BannerTemplate.AusAnniversaryBanner, label: 'Aus 10 yr moment' },
```
tomrf1, line 29: hopefully this work will eventually lead to this type of config being removed!

tjmw: Yes! (I've not added this btw, it's showing up as new because of the file rename).

public/src/components/channelManagement/bannerTests/bannerUiSelector.tsx
```diff
@@ -0,0 +1,189 @@
...
+    const uiType = event.target.value as UiType;
+    setUiType(uiType);
+
+    onUiChange(
+      uiType === 'Design' ? { designName: 'EXAMPLE' } : BannerTemplate.ContributionsBanner,
+    );
+  };
+
+  useEffect(() => {
+    // This isn't part of the design, so when changes are discarded we want to reflect the unedited configuration
```
tomrf1, line 145: not sure I understand the meaning of "This isn't part of the design" - can you explain?

tomrf1: 👍 makes sense

public/src/components/channelManagement/bannerTests/bannerUiSelector.tsx
```diff
@@ -0,0 +1,189 @@
...
+  designs,
+}: BannerUiSelectorProps) => {
+  const [uiType, setUiType] = useState<UiType>(uiIsDesign(ui) ? 'Design' : 'Template');
+
+  const onUiTypeChange = (event: React.ChangeEvent<{ value: unknown }>): void => {
+    const uiType = event.target.value as UiType;
+    setUiType(uiType);
+
+    onUiChange(
+      uiType === 'Design' ? { designName: 'EXAMPLE' } : BannerTemplate.ContributionsBanner,
```
tomrf1, line 140: Will this become something like `uiType === 'Design' ? designs[0] : ...`?
Shall we do this now?

tomrf1: Or do we not want to initialise the design when the type changes?

tjmw: Great spot! This is an oversight (originally I hardcoded a list of designs, one of which was called `EXAMPLE` before wiring up fetching the real ones from Dynamo). It wasn't _obviously_ broken in the UI though because currently there really is one called `EXAMPLE` in the DEV dynamo table 😃

Screenshot 2023-08-30 at 15 02 10

I'll fix this now.

tomrf1 (approved):
👍

PR #482, opened 2023-08-24
Improve withS3Data

tomrf1:
`withS3Data` is a small helper for maintaining state from S3, used by several pages.
The new amounts page has the new requirement to update _and_ send to S3 in one go. This isn't currently possible (because behind the scenes it's using the `useState` hook).

This PR renames the existing functions and adds a new one, so we now have:
```
update: (data: T) => void;
sendToS3: () => void;
updateAndSendToS3: (data: T) => void;
```


KaliedaRik (approved):
Much better names!

PR #485, opened 2023-09-04
Banner targeting by page context

tomrf1:
SDC change: https://github.com/guardian/support-dotcom-components/pull/953

I've moved/renamed `epicTestTargetContentEditor.tsx`, and changed the props to take the new `PageContextTargeting` model


app/models/BannerTests.scala
```diff
@@ -75,6 +75,7 @@ case class BannerTest(
     minArticlesBeforeShowingBanner: Int,
     userCohort: UserCohort,
     locations: List[Region] = Nil,
+    contextTargeting: Option[ContextTargeting] = Some(ContextTargeting(Nil,Nil,Nil,Nil)),
```
tomrf1, line 78: existing banner tests will not have this field, so set a default with no targeting

app/models/PageContextTargeting.scala
```diff
@@ -0,0 +1,8 @@
+package models
+
+case class PageContextTargeting(
```
tomrf1, line 3: I have not updated the epic model to use this type as it already has the fields (`tagIds` etc) at the top level, so it'd require a data migration.
But I thought it was nice to put these fields together in a type that can be passed around in SAC + SDC

public/src/components/channelManagement/testEditorContextTargeting.tsx
```diff
@@ -0,0 +1,76 @@
+import React from 'react';
```
tomrf1, line 1: Most of this is copied from epicTestTargetContentEditor.tsx.
It now uses the `PageContextTargeting` model

app/models/BannerTests.scala
```diff
@@ -75,6 +75,7 @@ case class BannerTest(
     minArticlesBeforeShowingBanner: Int,
     userCohort: UserCohort,
     locations: List[Region] = Nil,
+    contextTargeting: Option[PageContextTargeting] = Some(PageContextTargeting(Nil,Nil,Nil,Nil)),
```
tjmw, line 78: Oh! I'm curious as to why the default here isn't `None` given that it's an `Option`?

tjmw (approved):

app/models/BannerTests.scala
tomrf1: Good question. I've just changed this to not be an Option.
We can use the default value to handle any records in dynamodb that do not have the field.
And the client can expect this field to always exist

PR #496, opened 2023-09-27
Prevent changing banner design to draft if in use

tomrf1:
If a design is already in use by a test variant then we must not change its status.

- Refactors the status change endpoint to only take a single design name (in the path). This simplifies things on the backend, and we're unlikely to need the ability to batch change statuses.
- The status endpoint first fetches all banner tests and checks if the design is already in use. If it is, a 400 is returned and the user sees the list of test names

Screenshot 2023-09-27 at 09 03 18

### DIFF
--- a/app/controllers/banner/BannerDesignsController.scala
+++ b/app/controllers/banner/BannerDesignsController.scala
@@ -1,22 +1,18 @@
 package controllers.banner
 
 import com.gu.googleauth.AuthAction
-import models.DynamoErrors.{DynamoDuplicateNameError, DynamoNoLockError}
-import models.{BannerDesign, LockStatus}
+import models.DynamoErrors.{DynamoDuplicateNameError, DynamoError, DynamoNoLockError}
+import models.{BannerDesign, BannerTest, Channel, LockStatus}
 import play.api.libs.circe.Circe
-import play.api.mvc.{
-  AbstractController,
-  AnyContent,
-  ControllerComponents,
-  Result
-}
-import services.{DynamoBannerDesigns, S3Json, VersionedS3Data}
+import play.api.mvc.{AbstractController, AnyContent, ControllerComponents, Result}
+import services.{DynamoBannerDesigns, DynamoChannelTests, S3Json, VersionedS3Data}
 import services.S3Client.{S3ClientError, S3ObjectSettings}
 import utils.Circe.noNulls
 import zio.{IO, ZEnv, ZIO}
 import com.typesafe.scalalogging.LazyLogging
 import io.circe.syntax.EncoderOps
 import io.circe.generic.auto._
+import models.BannerUI.BannerDesignName
 
 import scala.concurrent.{ExecutionContext, Future}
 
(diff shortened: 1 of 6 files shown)


app/services/DynamoBannerDesigns.scala
```diff
@@ -232,30 +232,25 @@ class DynamoBannerDesigns(stage: String, client: DynamoDbClient)
     putAllBatched(deleteRequests)
   }
 
-  def updateStatuses(
```
tomrf1, line 235: updated this function to only take a single designName, which means no need for a batched transaction

tomrf1: which has the nice side effect of being much faster!

tjmw: 👍  My rational for going straight to bulk setting of statuses was that I could imagine wanting to archive multiple at once down the line (assuming we'd add `archived` to the list of possible statuses) but I was probably overthinking it. Better to keep it simple until we need that functionality.

app/controllers/banner/BannerDesignsController.scala
```diff
@@ -165,18 +162,43 @@ class BannerDesignsController(
       case _       => None
     }
 
-  def setStatus(rawStatus: String) =
-    authAction.async(circe.json[List[String]]) { request =>
+  private def getAllBannerTests(): ZIO[ZEnv, DynamoError, List[BannerTest]] = {
```
tjmw, line 165: I wonder if a better home for this function would be the `DynamoChannelTests` class rather than this controller?

conf/routes
```diff
@@ -200,10 +200,10 @@ GET   /frontend/banner-designs                         controllers.banner.Banner
 GET   /frontend/banner-designs/:designName             controllers.banner.BannerDesignsController.get(designName: String)
 POST  /frontend/banner-designs/update                  controllers.banner.BannerDesignsController.update
 POST  /frontend/banner-designs/create                  controllers.banner.BannerDesignsController.create
-POST  /frontend/banner-designs/lock/:testName          controllers.banner.BannerDesignsController.lock(testName: String)
-POST  /frontend/banner-designs/unlock/:testName        controllers.banner.BannerDesignsController.unlock(testName: String)
-POST  /frontend/banner-designs/takecontrol/:testName   controllers.banner.BannerDesignsController.forceLock(testName: String)
-POST  /frontend/banner-designs/status/:rawStatus       controllers.banner.BannerDesignsController.setStatus(rawStatus: String)
+POST  /frontend/banner-designs/lock/:designName        controllers.banner.BannerDesignsController.lock(designName: String)
```
tjmw, line 203: Thank you for spotting and fixing this!

tjmw (approved):
Looks great!

app/controllers/banner/BannerDesignsController.scala
tomrf1: Maybe an argument against that is that everything in `DynamoChannelTests` is generic, and this function specifies types of `ChannelTest`. But I'm being a bit picky here!

PR #500, opened 2023-10-03
ColourInput component

tomrf1:
We're going to have a lot of colour inputs in the banner design tool.
This PR refactors the existing colour inputs into a new component.
The component does validation of the string, but does not handle transformation to/from the `HexColour` model. We did explore this, but it conflicts with react-hook-form, which wants to work with simple types like strings.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "react": "^17.0.2",
     "react-beautiful-dnd": "^13.0.0",
     "react-dom": "^17.0.2",
-    "react-hook-form": "^6.7.2",
+    "react-hook-form": "^6.15.8",
     "react-qr-code": "^2.0.7",
     "react-router-dom": "^5.0.1",
     "recharts": "2.0.8",
--- a/public/src/components/channelManagement/bannerDesigns/BannerDesignForm.tsx
+++ b/public/src/components/channelManagement/bannerDesigns/BannerDesignForm.tsx
@@ -9,12 +9,9 @@ import {
   DEFAULT_BANNER_DESIGN,
 } from './utils/defaults';
 import { useStyles } from '../helpers/testEditorStyles';
-import {
-  hexColourToString,
-  hexColourStringRegex,
-  stringToHexColour,
-} from '../../../utils/bannerDesigns';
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -44,7 +44,7 @@
     "react": "^17.0.2",
     "react-beautiful-dnd": "^13.0.0",
     "react-dom": "^17.0.2",
-    "react-hook-form": "^6.7.2",
+    "react-hook-form": "^6.15.8",
```
tomrf1, line 47: for `useController`

public/src/components/channelManagement/bannerDesigns/BannerDesignForm.tsx
```diff
@@ -106,6 +98,7 @@ const BannerDesignForm: React.FC<Props> = ({
     const isValid = Object.keys(errors).length === 0;
     onValidationChange(isValid);
   }, [
+    // TODO - can we just use errors here?
```
tjmw, line 101: Not sure if you meant to include this TODO? (Fine either way!)

public/src/components/channelManagement/bannerDesigns/ColourInput.tsx
```diff
@@ -0,0 +1,58 @@
...
+      pattern: colourValidation,
+    },
+  });
+
+  return (
+    <TextField
+      name={name}
+      value={value}
+      onBlur={onBlur}
+      onChange={x => {
```
tjmw, line 45: Would `e` or `event` be slightly more descriptive/conventional than `x`?

tjmw (approved):
Looks great! I left a couple of v minor comments (not blockers).

public/src/components/channelManagement/bannerDesigns/ColourInput.tsx
tomrf1: yes! And actually react-form-hook can take the event directly - have updated

public/src/components/channelManagement/bannerDesigns/BannerDesignForm.tsx
tomrf1: removed - will address in another PR

PR #502, opened 2023-10-04
Banner design tool: change validation

tomrf1:
This PR proposes removing react-hook-form from the top-level of the banner design tool.
This is because react-hook-form makes it hard to nest components.

We already have a hook for monitoring the overall validation status of the form, to prevent the user from submitting invalid data. This hook stores validation status per field name. `BannerDesignForm` now makes better use of this, instead of using react-hook-form.

This means we no longer need the `BannerDesignFormData` model, which mirrors the `BannerDesign` model.

react-hook-form is still useful inside some nested components where we want to validate input fields.

The downside is that we now mix our own validation (`useValidation` at the top level) and react-hook-form in certain subcomponents. This is a bit confusing, but is similar to the approach in the channel test tools

### DIFF
--- a/public/src/components/channelManagement/bannerDesigns/BannerDesignForm.tsx
+++ b/public/src/components/channelManagement/bannerDesigns/BannerDesignForm.tsx
@@ -1,28 +1,16 @@
-import React, { useEffect } from 'react';
-import { TextField, Typography } from '@material-ui/core';
-import { EMPTY_ERROR_HELPER_TEXT } from '../helpers/validation';
-import { useForm } from 'react-hook-form';
-import { BannerDesign } from '../../../models/bannerDesign';
-import {
-  BannerDesignFormData,
-  BannerDesignFormData as FormData,
-  DEFAULT_BANNER_DESIGN,
-} from './utils/defaults';
+import React from 'react';
+import { Typography } from '@material-ui/core';
+import { BannerDesign, BannerDesignImage, BasicColours } from '../../../models/bannerDesign';
 import { useStyles } from '../helpers/testEditorStyles';
-import { hexColourToString, stringToHexColour } from '../../../utils/bannerDesigns';
 import { makeStyles, Theme } from '@material-ui/core/styles';
-import { ColourInput } from './ColourInput';
+import { ImageUrlsEditor } from './ImageUrlsEditor';
+import { BasicColoursEditor } from './BasicColoursEditor';
 
 type Props = {
   design: BannerDesign;
   setValidationStatus: (scope: string, isValid: boolean) => void;
   isDisabled: boolean;
   onChange: (design: BannerDesign) => void;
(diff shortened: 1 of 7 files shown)


public/src/components/channelManagement/bannerDesigns/BannerDesignForm.tsx
```diff
@@ -42,160 +30,50 @@ const BannerDesignForm: React.FC<Props> = ({
   const classes = useStyles();
   const localClasses = useLocalStyles();
 
-  const onValidationChange = (isValid: boolean): void => {
-    const validationScope = design.name;
-    setValidationStatus(validationScope, isValid);
+  const onValidationChange = (fieldName: string, isValid: boolean): void => {
+    setValidationStatus(fieldName, isValid);
```
tomrf1, line 34: we can now record validation status per field

public/src/components/channelManagement/bannerDesigns/ColourInput.tsx
```diff
@@ -1,53 +1,64 @@
...
+import { HexColour } from '../../../models/bannerDesign';
 
 const colourValidation = {
   value: hexColourStringRegex,
   message: 'Colours must be a valid 6 character hex code e.g. FF0000',
 };
 
 interface Props {
-  control: Control;
+  colour: HexColour;
```
tomrf1, line 19: it's not perfect, but I think it's nice to be able to work with the real model here

tjmw (approved):
Looks good, definitely an improvement I think 👍

public/src/components/channelManagement/bannerDesigns/BannerDesignForm.tsx
```diff
@@ -42,160 +30,50 @@ const BannerDesignForm: React.FC<Props> = ({
...
-            onBlur={handleSubmit(onSubmit)}
-            name="image.altText"
-            label="Banner Image Description (alt text)"
-            margin="normal"
-            variant="outlined"
-            disabled={isDisabled}
-            fullWidth
-          />
-        </div>
+        <ImageUrlsEditor
```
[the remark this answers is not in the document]
tomrf1: 👍 good point

public/src/components/channelManagement/bannerDesigns/ColourInput.tsx
```diff
@@ -1,53 +1,64 @@
...
-  const {
-    field: { ref, onChange, value },
-  } = useController({
-    name,
-    control,
-    rules: {
-      required: EMPTY_ERROR_HELPER_TEXT,
-      pattern: colourValidation,
-    },
+  const defaultValues = { colour: hexColourToString(colour) };
```
tomrf1, line 35: @tjmw the colour is transformed to a string here, for use by react-hook-form

PR #503, opened 2023-10-05
More banner colours

tomrf1:
Corresponding SDC PR: https://github.com/guardian/support-dotcom-components/pull/977

Adds the ability to edit lots more colours, including ctas.

Screenshot 2023-10-05 at 16 50 08
Screenshot 2023-10-05 at 16 50 21
Screenshot 2023-10-05 at 16 50 29

Some things to consider later:

1. Make each section collapsible, as the form is now very long and there's a lot of info
2. Improve the `ColourInput` component to have a fixed width, and display the chosen colour next to the input
3. Display the different roundel designs next to the radios

### DIFF
--- a/app/models/BannerDesign.scala
+++ b/app/models/BannerDesign.scala
@@ -18,35 +18,67 @@ object BannerDesignStatus {
   implicit val statusDecoder = deriveEnumerationDecoder[BannerDesignStatus]
 }
 
+sealed trait GuardianRoundelDesign
+object GuardianRoundelDesign {
+  case object default extends GuardianRoundelDesign
+  case object brand extends GuardianRoundelDesign
+  case object inverse extends GuardianRoundelDesign
+
+  implicit val customConfig: Configuration = Configuration.default.withDefaults
+  implicit val encoder = deriveEnumerationEncoder[GuardianRoundelDesign]
+  implicit val decoder = deriveEnumerationDecoder[GuardianRoundelDesign]
+}
+
 case class BannerDesignImage(
-    mobileUrl: String,
-    tabletDesktopUrl: String,
-    wideUrl: String,
-    altText: String
+  mobileUrl: String,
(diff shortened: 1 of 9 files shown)


app/models/BannerDesign.scala
```diff
@@ -18,35 +18,67 @@ object BannerDesignStatus {
...
+  implicit val customConfig: Configuration = Configuration.default.withDefaults
+  implicit val encoder = deriveEnumerationEncoder[GuardianRoundelDesign]
+  implicit val decoder = deriveEnumerationDecoder[GuardianRoundelDesign]
+}
+
 case class BannerDesignImage(
-    mobileUrl: String,
-    tabletDesktopUrl: String,
-    wideUrl: String,
-    altText: String
```
tjmw, line 25: Oh are you using scalafmt? If so I must not have it configured to run on save for this project, sorry about that.

public/src/components/channelManagement/bannerDesigns/utils/defaults.ts
```diff
@@ -1,7 +1,9 @@
 import { BannerDesign } from '../../../../models/bannerDesign';
 import { stringToHexColour } from '../../../../utils/bannerDesigns';
 
-export const DEFAULT_BANNER_DESIGN = {
+export const createDefaultBannerDesign = (name: string): BannerDesign => ({
```
tjmw, line 4: Ah great that now the colour fields in the form are `HexColour`s this can all be simplified/duplication removed 👍

tjmw (approved):
👍  Great! Really nice in the UI to have organised the default and hover states into columns so the counterparts are next to each other.

app/models/BannerDesign.scala
tomrf1: I'm not - I was just making indentation consistent (2 spaces). Perhaps we should enable it for this project though

PR #505, opened 2023-10-06
Display banner design usage

tomrf1:
When you select a design it fetches a list of tests that currently use that design, and links to them.

Screenshot 2023-10-06 at 15 12 32
Screenshot 2023-10-06 at 15 12 37


tjmw (approved):
👍

PR #515, opened 2023-10-13
Fix colour input error position

tomrf1:
Before:
Screenshot 2023-10-13 at 12 28 07
After:
Screenshot 2023-10-13 at 13 21 30


tjmw (approved):
Looks much better!

PR #536, opened 2023-12-05
Add header image form to designable banner page

KaliedaRik:
## What does this change?
Extends the designable banner page to allow users to self-serve a header image

## How to test

- In the CODE designable banner page, live view a design which makes use of this new component
- In the CODE banner channels, create and preview banner tests which uses the design: one with header copy, another without header copy

### DIFF
--- a/app/models/BannerDesign.scala
+++ b/app/models/BannerDesign.scala
@@ -27,6 +27,13 @@ object GuardianRoundelDesign {
   implicit val decoder = deriveEnumerationDecoder[GuardianRoundelDesign]
 }
 
+case class HeaderImage(
+  mobileUrl: String,
+  tabletDesktopUrl: String,
+  wideUrl: String,
+  altText: String
+)
+
 sealed trait BannerDesignVisual
 object BannerDesignVisual {
   case class Image(
@@ -99,6 +106,7 @@ case class BannerDesign(
   name: String,
   status: BannerDesignStatus,
   visual: Option[BannerDesignVisual],
+  headerImage: Option[HeaderImage],
   colours: BannerDesignColours,
   lockStatus: Option[LockStatus],
 )
--- a/public/src/components/channelManagement/bannerDesigns/BannerDesignForm.tsx
+++ b/public/src/components/channelManagement/bannerDesigns/BannerDesignForm.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { Accordion, AccordionDetails, AccordionSummary } from '@material-ui/core';
 import {
   BannerDesign,
-  BannerDesignImage,
+  BannerDesignHeaderImage,
   BannerDesignVisual,
   BasicColours,
   CtaDesign,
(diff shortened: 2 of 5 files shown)


public/src/components/channelManagement/bannerDesigns/HeaderImageEditor.tsx
```diff
@@ -0,0 +1,104 @@
+import { TextField } from '@material-ui/core';
+import { EMPTY_ERROR_HELPER_TEXT } from '../helpers/validation';
+import React, { useEffect } from 'react';
+import { useForm } from 'react-hook-form';
+import { BannerDesignImage } from '../../../models/bannerDesign';
+
+const imageUrlValidation = {
+  value: /^https:\/\/i\.guim\.co\.uk\//,
+  message: 'Images must be valid URLs hosted on https://i.guim.co.uk/',
+};
```
tjmw, line 10: Not a big deal, but perhaps this common validation could be shared with the other image fields?

public/src/components/channelManagement/bannerDesigns/HeaderImageEditor.tsx
```diff
@@ -0,0 +1,104 @@
...
+}: Props) => {
+  const { register, handleSubmit, errors, reset } = useForm<BannerDesignImage>({
+    mode: 'onChange',
+    defaultValues: headerImage,
+  });
+
+  useEffect(() => {
+    const isValid = Object.keys(errors).length === 0;
+    onValidationChange('headerImage', isValid);
+    console.log(errors);
```
tjmw, line 33: Should this console.log be removed?

app/models/BannerDesign.scala
```diff
@@ -27,6 +27,13 @@ object GuardianRoundelDesign {
   implicit val decoder = deriveEnumerationDecoder[GuardianRoundelDesign]
 }
 
+case class HeaderImage(
```
tomrf1, line 30: did you consider reusing `BannerDesignVisual.Image`?

tjmw: I think I may have dissuaded Rik from that because that type has a `kind` which isn't used/needed here. What do you think?

tomrf1: Ok fair enough

public/src/components/channelManagement/bannerDesigns/BannerDesignForm.tsx
```diff
@@ -161,7 +170,7 @@ const BannerDesignForm: React.FC<Props> = ({
 
       <Accordion className={classes.accordion}>
         <AccordionSummary className={classes.sectionHeader} expandIcon={<ExpandMoreIcon />}>
-          Image or Choice Cards
+          Main Image or Choice Cards
```
tomrf1, line 173: ❤️

tomrf1 (approved):
great!

PR #538, opened 2023-12-08
remove minArticlesBeforeShowingBanner

tomrf1:
The "Display rules" targeting option in the RRCP banner tool lets users define after how many pageviews a banner test should be displayed. But the local storage counter that is used for this, `alreadyVisited`, counts all pageviews from a browser's first visit to theguardian.com.
This feature is not doing what Marketing want, so we're removing it.

We're replacing it [with a DCR change](https://github.com/guardian/dotcom-rendering/pull/9819) to prevent different banner views on consecutive pageviews.

[SDC change](https://github.com/guardian/support-dotcom-components/pull/1024)


tjmw (approved):
Looks good.

There will be tests in Dynamo which have this option set in the serialized JSON - I presume that's not an issue as it'll just be ignored? (And probably removed if the test is modified & saved in future?)

PR #539, opened 2023-12-18
Banner design archive

tomrf1:
Adds the ability to archive individual banner designs.
The button is only available if the design status is `Draft`.
Stores archived tests in a separate Dynamodb table (as with channel tests). The table is keyed on `name` and `date`, because we're likely to get duplicate names as time goes on

Screenshot 2023-12-18 at 10 44 18
Screenshot 2023-12-18 at 10 44 25

### DIFF
--- a/app/services/DynamoBannerDesigns.scala
+++ b/app/services/DynamoBannerDesigns.scala
@@ -95,6 +95,16 @@ class DynamoBannerDesigns(stage: String, client: DynamoDbClient)
       case other                                => DynamoPutError(other)
     }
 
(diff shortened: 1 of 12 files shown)


app/services/DynamoBannerDesigns.scala
```diff
@@ -216,20 +226,13 @@ class DynamoBannerDesigns(stage: String, client: DynamoDbClient)
     update(request)
   }
 
-  def deleteBannerDesigns(designNames: List[String],
```
tomrf1, line 219: this wasn't used

KaliedaRik (approved):
Nice work!

PR #547, opened 2024-01-23
Add article count copy editor field

andrewHEguardian:
## What does this change?

Adds the "copy" property to Epic test models [(already exists here)](https://github.com/guardian/support-dotcom-components/blob/0594dce2d06307b0b01e25a304e26255ce7f99c7/packages/shared/src/types/abTests/epic.ts#L30) and creates an editor field for it. 

## How to test

Edit a variant of an epic test - update the "Article count copy" under "Separate article count". Preview and see this copy. Save, reload and see the copy is still there.

## Images

image

## Accessibility

### DIFF
--- a/public/src/components/channelManagement/variantEditorSeparateArticleCountEditor.tsx
+++ b/public/src/components/channelManagement/variantEditorSeparateArticleCountEditor.tsx
@@ -2,6 +2,12 @@ import React from 'react';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
 import Checkbox from '@material-ui/core/Checkbox';
 import { SeparateArticleCount } from '../../models/epic';
+import { TextField } from '@material-ui/core';
+import { useForm } from 'react-hook-form';
+
+interface FormData {
+  copy: string;
+}
 
 interface VariantEditorSeparateArticleCountEditorProps {
   separateArticleCount?: SeparateArticleCount;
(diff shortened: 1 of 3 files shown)


public/src/components/channelManagement/variantEditorSeparateArticleCountEditor.tsx
```diff
@@ -2,6 +2,12 @@ import React from 'react';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
 import Checkbox from '@material-ui/core/Checkbox';
 import { SeparateArticleCount } from '../../models/epic';
+import { TextField } from '@material-ui/core';
+import { useForm } from 'react-hook-form';
+
+interface FormData {
+  copy: string;
```
tomrf1, line 9: If you clear the contents of this field does the tool submit an empty string (rather than null) to the database? I'm wondering what then happens in the epic

tomrf1: That makes sense. Perhaps it's worth being defensive and transforming `''` to `undefined` in this component?

public/src/components/channelManagement/variantEditorSeparateArticleCountEditor.tsx
```diff
@@ -18,6 +24,20 @@ const VariantEditorSeparateArticleCountEditor: React.FC<VariantEditorSeparateArt
     updateSeparateArticleCount(Boolean(separateArticleCount) ? undefined : { type: 'above' });
```
andrewHEguardian, line 24: the way this worked before means that the copy will be lost if "Above" is set to false, which I guess is fine?

tomrf1 (approved):
great!

PR #548, opened 2024-01-26
Remove nulls when serializing to dynamodb

tomrf1:
Currently when fields in the scala model are `None`, they're serialized to Dynamodb as `Null`.
This adds complexity to support-dotcom-components, where the typescript and zod validation have to allow for both `undefined` _or_ `null`.
We can simplify this by just never writing nulls to dynamodb.

This will require an initial migration of all existing data in dynamodb to remove nulls


andrewHEguardian (approved):
seems worth a try! nice 👍

andrewHEguardian:
this was done on the SDC side
https://github.com/guardian/support-dotcom-components/pull/1053/files#diff-a7cefd0d19353800ccc1541c0399a65b4af34918966b909b12f20ba3aa49a35e

PR #549, opened 2024-01-30
Ensure test exists before locking

tomrf1:
We've had a couple of instances of the following bug:

1. user opens the banner tests page
2. user opens the banner tests page in another tab and archives a test
3. user then tries to edit the same test from the original page (which still shows that test), which triggers a request to lock the test
4. the backend creates a new entry in dynamodb for the archived test, because dynamo UpdateItemRequests will create a row if it's not there!

This PR fixes that by checking the item exists in dynamodb before performing the update

Screenshot 2024-01-30 at 09 10 56

### DIFF
--- a/app/services/DynamoChannelTests.scala
+++ b/app/services/DynamoChannelTests.scala
@@ -256,12 +256,16 @@ class DynamoChannelTests(stage: String, client: DynamoDbClient) extends DynamoSe
         .key(buildKey(channel, testName))
         .updateExpression("set lockStatus = :lockStatus")
         .expressionAttributeValues(Map(
-          ":lockStatus" -> jsonToDynamo(lockStatus.asJson)
+          ":lockStatus" -> jsonToDynamo(lockStatus.asJson),
         ).asJava)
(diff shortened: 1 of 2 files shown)


app/services/DynamoChannelTests.scala
```diff
@@ -256,12 +256,16 @@ class DynamoChannelTests(stage: String, client: DynamoDbClient) extends DynamoSe
         .key(buildKey(channel, testName))
         .updateExpression("set lockStatus = :lockStatus")
         .expressionAttributeValues(Map(
-          ":lockStatus" -> jsonToDynamo(lockStatus.asJson)
+          ":lockStatus" -> jsonToDynamo(lockStatus.asJson),
         ).asJava)
+        .expressionAttributeNames(Map("#name" -> "name").asJava)
```
tomrf1, line 261: we have to alias `name` to `#name` here because `name` is a reserved word in dynamodb!

charleycampbell (approved):
Looks good to me! Nice fix 👍

PR #551, opened 2024-02-07
New banner image url component for desktop 

charleycampbell:
## What does this change?
TL;DR Banner image URLs have been updated so the tablet url field is only for Tablet (1:1) and desktop url field is for Desktop and up breakpoints (5:3)

Currently when we pull an image from the grid to use in our banners the same url is used for tablet and desktop for 1:1 image ratio. As part of the work to align RRCP we have discovered that the new image ratio requested for desktop to be 5:3, so a slight reworking is required. 

Two new URL fields have been introduced `tabletURL` which will replace `tabletDesktopURL` and `desktopURL` which will replace `wideURL`.

For backward compatibility these new fields are optional. It will look for the previous URL of the image being used if no URL is provided. Once we are happy with the change we will removed deprecated fields and the functionality which handles checking for the URL.

## How to test
Tested locally and on CODE. On RRCP check to see that new fields are visible and populated with the same image URLs as their matching field;
- tabletDesktopURL  = tabletURL
- wideUrl = desktopUrl

Click the edit button on the banner design and then save. This will then populate the DynamoDB table with all of the fields being populated. Check the banner looks correct!
## Images
An example of a banner on the RRCP CODE environment:
image

## Accessibility


KentGrm (approved):

tomrf1 (approved):
👍

PR #555, opened 2024-02-21
Upgrade emotion

tomrf1:
tested in CODE by using the banner live preview feature (which depends on emotion being present)


andrewHEguardian (approved):
👍 👩‍🎤

PR #556, opened 2024-02-21
Upgrade to MUI (material UI) 5

tomrf1:
This is quite a major upgrade, mostly changing imports.

The UI is mostly the same. I've made a few tweaks to keep things looking ok.

The biggest change is that the `main` colour now applies in more places. In PROD this is blue, in CODE it's purple.
Previously it only applied to the top bar and some buttons. It now applies to a few more buttons, e.g. in the left sidebar:
before:
Screenshot 2024-02-21 at 15 19 02
after:
Screenshot 2024-02-21 at 15 19 05

Note - the recommendation is to [migrate to emotion](https://mui.com/material-ui/migration/migrating-from-jss/) when upgrading to MUI 5. I've created a health card to do this, as it's probably quite a big change.

### DIFF
--- a/public/src/components/amounts/AmountsVariantEditorRowInput.tsx
+++ b/public/src/components/amounts/AmountsVariantEditorRowInput.tsx
@@ -1,5 +1,6 @@
 import React, { useState } from 'react';
-import { Button, makeStyles, TextField, Theme } from '@material-ui/core';
+import { Button, TextField, Theme } from '@mui/material';
+import { makeStyles } from '@mui/styles';
 
 const useStyles = makeStyles(({ spacing }: Theme) => ({
   textField: {
--- a/public/src/components/amounts/AmountsVariantEditorRowInput.tsx
+++ b/public/src/components/amounts/AmountsVariantEditorRowInput.tsx
@@ -55,6 +55,7 @@ export const AmountsVariantEditorRowInput: React.FC<AmountsVariantEditorRowInput
   return (
     <div>
       <TextField
+        variant={'standard'}
         className={classes.textField}
         error={!!currentError}
         helperText={currentError}
--- a/public/src/components/channelManagement/epicTests/tagsEditor.tsx
+++ b/public/src/components/channelManagement/epicTests/tagsEditor.tsx
@@ -1,6 +1,6 @@
-import { Autocomplete } from '@material-ui/lab';
+import { Autocomplete } from '@mui/lab';
 import React, { useEffect, useCallback } from 'react';
-import { TextField } from '@material-ui/core';
+import { TextField } from '@mui/material';
 import throttle from 'lodash/throttle';
 
 interface Tag {
@@ -100,11 +100,11 @@ export const TagsEditor: React.FC<TagEditorProps> = ({
       renderInput={(params): JSX.Element => (
         <TextField {...params} variant="outlined" label={label} />
       )}
-      renderOption={(option): JSX.Element => {
+      renderOption={(props, option): JSX.Element => {
         return <div>{option.name ? `${option.name} (${option.id})` : option.id}</div>;
       }}
       onChange={(event, values: Tag[], reason): void => {
-        if (reason === 'select-option' || reason === 'remove-option') {
+        if (reason === 'selectOption' || reason === 'removeOption') {
           onUpdate(values.map(value => value.id));
           setInputValue('');
         }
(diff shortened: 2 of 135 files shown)


public/src/components/amounts/AmountsVariantEditorRowInput.tsx
```diff
@@ -54,6 +55,7 @@ export const AmountsVariantEditorRowInput: React.FC<AmountsVariantEditorRowInput
   return (
     <div>
       <TextField
+        variant={'standard'}
```
tomrf1, line 58: keeps the UI as it was before (an underlined field, rather than a field with a box around it)

public/src/components/channelManagement/testListSidebarFilterSelector.tsx
```diff
@@ -1,7 +1,14 @@
 import React from 'react';
-import { FormControl, InputLabel, Select, MenuItem } from '@material-ui/core';
+import { FormControl, InputLabel, Select, MenuItem, SelectChangeEvent } from '@mui/material';
 
 import { regions, regionIds, RegionsAndAll } from '../../utils/models';
+import { makeStyles } from '@mui/styles';
+
+const useStyles = makeStyles(() => ({
+  container: {
+    marginTop: '8px',
```
tomrf1, line 9: adding margin above the region filter as it's become a bit cramped

public/src/components/channelManagement/variantEditorSecondaryCtaEditor.tsx
```diff
@@ -56,8 +57,15 @@ const VariantEditorSecondaryCtaEditor: React.FC<VariantEditorSecondaryCtaEditorP
...
-          <Select value={cta?.type || 'None'} onChange={handleChange} disabled={isDisabled}>
+          <InputLabel id="secondaryCtaTypeLabel">{label}</InputLabel>
+          <Select
+            id={'secondaryCtaType'}
+            labelId={'secondaryCtaTypeLabel'}
+            label={label}
+            value={cta?.type || 'None'}
+            onChange={handleChange}
+            disabled={isDisabled}
+          >
```
tomrf1, line 68: The UI is a bit different here and this change ensures the label looks ok -
Screenshot 2024-02-21 at 15 13 32

public/src/main.tsx
```diff
@@ -18,25 +18,30 @@ import {
...
 import QrCodePage from './components/utilities/QrCodePage';
 import AppsMeteringSwitches from './components/appsMeteringSwitches';
 import { SuperModeDashboard } from './components/channelManagement/superMode/superModeDashboard';
 import BannerDesigns from './components/channelManagement/bannerDesigns/';
 import DefaultPromos from './components/defaultPromos';
+import { StyledEngineProvider } from '@mui/material';
+
+declare module '@mui/styles' {
+  type DefaultTheme = Theme;
+}
```
tomrf1, line 44: See https://mui.com/material-ui/migration/v5-style-changes/#%E2%9C%85-add-module-augmentation-for-defaulttheme-typescript

public/src/main.tsx
```diff
@@ -144,111 +149,109 @@ const AppRouter = () => {
   );
 
   return (
-    <ThemeProvider theme={getTheme()}>
```
tomrf1, line 147: `ThemeProvider` is now outside of this component, because we're not allowed to `useStyles` (above) before

public/src/main.tsx
```diff
@@ -144,111 +151,109 @@ const AppRouter = () => {
...
   );
 };
 
 const container = document.getElementById('root');
 if (container) {
   const root = createRoot(container);
-  root.render(<AppRouter />);
+  root.render(
+    <ThemeProvider theme={getTheme()}>
+      <StyledEngineProvider injectFirst>
```
tomrf1, line 252: this prevents our styles from being overridden:
https://mui.com/material-ui/migration/v5-style-changes/#%E2%9C%85-adjust-css-injection-order

public/src/components/channelManagement/epicTests/tagsEditor.tsx
```diff
@@ -100,11 +100,11 @@ export const TagsEditor: React.FC<TagEditorProps> = ({
       renderInput={(params): JSX.Element => (
         <TextField {...params} variant="outlined" label={label} />
       )}
-      renderOption={(option): JSX.Element => {
+      renderOption={(props, option): JSX.Element => {
         return <div>{option.name ? `${option.name} (${option.id})` : option.id}</div>;
       }}
       onChange={(event, values: Tag[], reason): void => {
-        if (reason === 'select-option' || reason === 'remove-option') {
+        if (reason === 'selectOption' || reason === 'removeOption') {
```
andrewHEguardian, line 107: what changed here?

andrewHEguardian (approved):
Wow! You deserve a 🍫

public/src/components/channelManagement/epicTests/tagsEditor.tsx
tomrf1: The values provided by MUI changed slightly. Luckily the typescript caught it because the types are well defined!
See https://mui.com/material-ui/api/autocomplete/#autocomplete-prop-onChange

PR #558, opened 2024-02-26
Fix tag + section pickers

tomrf1:
The ability to select items from the dropdown broke with the recent MUI upgrade.
This is because the API has changed slightly.
The `renderOption` prop now expects a `<li>` element with the props - https://mui.com/material-ui/api/autocomplete/#autocomplete-prop-renderOption

### Before:
Screenshot 2024-02-26 at 09 42 40

### After:
Screenshot 2024-02-26 at 09 42 54


charleycampbell (approved):
Nice 👍 thanks for adding in the link to the documentation too!

PR #567, opened 2024-04-10
feat: add supporter plus default promo code

jamesgorrie:
[Part 1/3](https://github.com/guardian/support-frontend/issues/5897)

## What does this change?
Adds the Supporter Plus option for a default `promoCode`.

## How to test
- Add `promoCode`
- Check s3

## Images
Screenshot 2024-04-10 at 12 02 21
Screenshot 2024-04-10 at 12 02 36

PR #568, opened 2024-04-10
Make SupporterPlus default promo a required field

jamesgorrie:
[Part 2](https://github.com/guardian/support-frontend/issues/5897)

~Depends on [Part 1](https://github.com/guardian/support-admin-console/pull/567)~

As we know this data was introduced [via here](https://github.com/guardian/support-admin-console/pull/567/files#r1559256856) now exists, we can make it required.


paul-daniel-dempsey (approved):
Yep, not optional now, great!

PR #570, opened 2024-04-17
Link builder: set utm_source

tomrf1:
The tracking should also include the utm_source field.
This is based on the chosen `utm_medium` value, so is not explicitly chosen by the user.

I was unable to find a good way to update two separate fields from the same component with react-hook-form, so ended up combining them into a `sourceAndMedium` field, separated by a double-underscore.

While doing this I noticed that the form doesn't pick up subsequent updates to the medium. I fixed this by passing a callback into `MediumSelector` to reset the link, meaning the user has to click the submit button again.


paulbrown1982 (approved):
👍

PR #571, opened 2024-04-17
Add consent targeting to RRCP

charleycampbell:
## What does this change?
Adds user consent as a targeting options to the RRCP. This means we can target users depending on whether they have accepted or rejected cookies.

## How to test
Create (or edit) a banner test and select which consent option you would like, save and reload. The option should be saved.

## Images
image

## Accessibility

### DIFF
--- a/public/src/components/channelManagement/epicTests/epicTestEditor.tsx
+++ b/public/src/components/channelManagement/epicTests/epicTestEditor.tsx
@@ -339,6 +339,7 @@ export const getEpicTestEditor = (
               onSignedInStatusChange={onSignedInStatusChange}
               selectedConsentStatus={test.consentStatus}
               onConsentStatusChange={onConsentChange}
+              showConsentStatusSelector={false}
               platform={epicEditorConfig.platform}
             />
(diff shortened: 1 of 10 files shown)


public/src/components/channelManagement/epicTests/epicTestEditor.tsx
```diff
@@ -8,6 +8,7 @@ import {
   DeviceType,
   SignedInStatus,
   PageContextTargeting,
+  ConsentStatus,
```
andrewHEguardian, line 11: we can edit it in the Epic editor, does it need to be added to the epic model?

charleycampbell: Yeah really good catch here! I've added it to the model for epics and headers but since the scope of this ticket is just for banners we have hidden the toggles from epics and headers but the functionality is there when we do decide to expand

andrewHEguardian (approved):
Nice! 🚀

tomrf1 (approved):
Great!

PR #596, opened 2024-05-16
Change RRCP article count component to support article count targeting by tag

LAKSHMIRPILLAI:
## What does this change?

This PR is to enhance RRCP self serve capability by introducing a new feature to select tags for article count. These tags will enable targeting based on article count by using a specific tag/group of tags for content tagged articles .
## Images

Screenshot 2024-05-16 at 16 40 35

## After Update 

image

It updates the corresponding DynamoDB table correctly

image


public/src/components/channelManagement/MutliSelectTagEditor.tsx
```diff
@@ -0,0 +1,113 @@
...
+  },
+}));
+
+interface Option {
+  label: string;
+  value: string;
+}
+
+const options: Option[] = [
+  { label: 'environment/climate-change', value: 'environment/climate-change' },
```
andrewHEguardian, line 26: where did the list of different tag options come from? is it in a Trello ticket?

public/src/components/channelManagement/MutliSelectTagEditor.tsx
```diff
@@ -0,0 +1,113 @@
...
+  { label: 'world/europe-news', value: 'world/europe-news' },
+  { label: 'world/russia', value: 'world/russia' },
+  { label: 'books/books', value: 'books/books' },
+  { label: 'culture/culture', value: 'culture/culture' },
+  { label: 'world/coronavirus-outbreak', value: 'world/coronavirus-outbreak' },
+  { label: 'world/race', value: 'world/race' },
+  { label: 'inequality/inequality', value: 'inequality/inequality' },
+  { label: 'technology/technology', value: 'technology/technology' },
+  { label: 'business/business', value: 'business/business' },
+  { label: 'tone/recipes', value: 'tone/recipes' },
```
tomrf1, line 43: not at all important, but you could avoid defining each string twice with:
```
const options: Option[] = [
  'environment/climate-change',
  'environment/climate-crisis',
  ...
].map(id => ({
  label: id,
  value: id,
}));
```

public/src/components/channelManagement/MutliSelectTagEditor.tsx
```diff
@@ -0,0 +1,113 @@
...
+  tags,
+  onUpdate,
+}: MultiselectAutocompleteProps) => {
+  const classes = useStyles();
+
+  const [inputValue, setInputValue] = React.useState<string>('');
+
+  return (
+    <div className={classes.container}>
+      <span style={{ fontSize: '1rem', fontWeight: 'normal' }}>Tag for content tagged article</span>
```
andrewHEguardian, line 63:
```suggestion
      <span style={{ fontSize: '1rem', fontWeight: 'normal' }}>Tags for content tagged article</span>
```
Plural "Tags" since it can be multiple maybe

tomrf1, line 63: How about: "Filter article count by tag"?

public/src/components/channelManagement/MutliSelectTagEditor.tsx
LAKSHMIRPILLAI: The tags we track are mentioned [here](https://github.com/guardian/support-dotcom-components/blob/main/packages/shared/src/lib/history.ts#L64) in SDC

andrewHEguardian: thanks!

andrewHEguardian (approved):
👏

app/models/ArticlesViewedSettings.scala
```diff
@@ -9,5 +9,6 @@ case class MaxViews(
 case class ArticlesViewedSettings(
  minViews: Option[Int],
  maxViews: Option[Int],
- periodInWeeks: Int
+ periodInWeeks: Int,
+ tags: List[String]=Nil
```
tomrf1, line 13: this name doesn't match the SDC model - https://github.com/guardian/support-dotcom-components/pull/1134/files#diff-1bf2b8e1c7e257fc1c135431146989151ac2299f647de0de7b70bdc6ea2d347fR81

LAKSHMIRPILLAI: great spot!Thank you

tomrf1 (approved):
great work!

PR #601, opened 2024-06-06
Require minViews

tomrf1:
In SDC the `minViews` field [is required](https://github.com/guardian/support-dotcom-components/blob/main/packages/shared/src/types/abTests/shared.ts#L78) if the `articlesViewedSettings` field is defined.

But this is not the case in the RRCP, which lets you set it to undefined.
This PR makes it mandatory. If Marketing want to target users below a `maxViews` count, they can set `minViews` to 0.

Screenshot 2024-06-06 at 09 18 20


andrewHEguardian (approved):
👍

PR #606, opened 2024-06-19
Epic impressions

tomrf1:
Adds an "Analytics" button in the epic tool.
Screenshot 2024-06-19 at 12 57 19

It opens a model which makes a request to the backend for impressions data for the current epic test. The backend then queries athena.
A chart shows impressions over time per variant (uses the recharts library).
Screenshot 2024-06-19 at 12 25 39

Currently we only have impressions data for the epic in athena.

### DIFF
--- a/app/services/Athena.scala
+++ b/app/services/Athena.scala
@@ -133,8 +133,8 @@ class Athena() extends StrictLogging {
       s"""
          |SELECT date_hour, ab.variant, COUNT(*) AS views
          |FROM acquisition.epic_views_prod, UNNEST(abtests) t(ab)
-         |WHERE date_hour >= timestamp '$from' AND date_hour < timestamp '$to'
-         |AND ab.name = '$testName'
+         |WHERE date_hour >= TIMESTAMP '$from' AND date_hour < TIMESTAMP '$to'
+         |AND ab.name = ?
          |GROUP BY 1,2 ORDER BY 1,2
     """
 
@@ -143,6 +143,7 @@ class Athena() extends StrictLogging {
       .queryString(
         query.stripMargin
       )
+      .executionParameters(testName)
       .queryExecutionContext(queryExecutionContext)
       .resultConfiguration(resultConfiguration)
       .build()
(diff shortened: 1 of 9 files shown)


app/services/Athena.scala
```diff
@@ -144,7 +175,7 @@ class Athena() extends StrictLogging {
       .retryWhileEquals(AthenaResultPending)
 
   // Get resultSet of the completed query and parse each row
-  private def getQueryResult(queryExecutionId: QueryExecutionId): ZIO[ZEnv, AthenaError, List[ArticleEpicData]] = {
+  private def getQueryResult(queryExecutionId: QueryExecutionId): ZIO[ZEnv, AthenaError, List[Map[String, String]]] = {
```
tomrf1, line 178: small refactor to make the `getQueryResult` more generic

app/services/Athena.scala
```diff
@@ -124,6 +126,35 @@ class Athena() extends StrictLogging {
...
 
+  private def startVariantViewsQuery(from: String, to: String, channel: String, testName: String): ZIO[ZEnv, AthenaError, QueryExecutionId] = {
+    logger.info(s"Querying athena")
+
+    val query =
+      s"""
+         |SELECT date_hour, ab.variant, COUNT(*) AS views
+         |FROM acquisition.epic_views_prod, UNNEST(abtests) t(ab)
+         |WHERE date_hour >= timestamp '$from' AND date_hour < timestamp '$to'
+         |AND ab.name = '$testName'
```
andrewHEguardian, line 137: can you do sql injection in Athena? 😀  if I name my AB test `DROP TABLES` what would happen?

tomrf1: good question! It can't change the data, but perhaps it can mess with the athena "tables".
It looks like it's possible - https://docs.aws.amazon.com/athena/latest/ug/querying-with-prepared-statements.html
but I'm not sure yet how to do it with the java sdk

andrewHEguardian (approved):
looks fantastic 👏

app/services/Athena.scala
tomrf1: It's working for the test name now, thanks for suggesting!
See https://github.com/guardian/support-admin-console/pull/606/commits/e96f4e84fcbdff66d530fadbfb065d064220e6af

I can't get it to work with timestamp casting though

PR #608, opened 2024-06-20
Switch on BANDIT_FEATURE_SWITCH

LAKSHMIRPILLAI:
## What does this change?

This is to switch the BANDIT_FEATURE_SWITCH to true , so that Experiment Methodology becomes available to select the Multi-Multi-Armed Bandit option. which in turn sets isBanditTest boolean in DB 

## Images
image

##NOTE

This change should be merged only when we are ready to create the MAB epic tests and need to be switched off (changed back to false)once we are done with the tests.


andrewHEguardian (approved):
👍

tomrf1 (approved):

PR #610, opened 2024-06-21
Add tooltip to chart

tomrf1:
so users can hover on datapoints to see the details -
Screenshot 2024-06-21 at 09 32 02
Also shortens the timestamps - it's always hourly


charleycampbell (approved):
Nice idea!

PR #612, opened 2024-06-24
Fix athena errors

tomrf1:
We're seeing this error and it's triggering an alarm:
`Failed to parse row from athena: None.`
This happens because the first row in the result is always the column names -
`Map(date_hour -> date_hour, variant -> variant, views -> views)`

So we can just drop the first row

### DIFF
--- a/app/services/Athena.scala
+++ b/app/services/Athena.scala
@@ -203,7 +203,7 @@ class Athena() extends StrictLogging {
           .flatMap(row => {
             val data = ArticleEpicData(row)
             if (data.isEmpty) {
-              logger.error(s"Failed to parse row from athena: $row")
+              logger.info(s"Failed to parse row from athena: $row")
             }
             data
           })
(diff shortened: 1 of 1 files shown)


app/services/Athena.scala
```diff
@@ -203,7 +203,7 @@ class Athena() extends StrictLogging {
           .flatMap(row => {
             val data = ArticleEpicData(row)
             if (data.isEmpty) {
-              logger.error(s"Failed to parse row from athena: $row")
+              logger.info(s"Failed to parse row from athena: $row")
```
andrewHEguardian, line 206: why log at all if it's an expected outcome?

tomrf1: Good point.
And actually I just tested this branch in CODE and the improved logging suggests it's because _all_ queries return a row with the column names!
`Map(date_hour -> date_hour, variant -> variant, views -> views)`
So I'm going to find a smarter solution

andrewHEguardian (approved):
cool 👍 😀

PR #613, opened 2024-06-26
Remove legacy banners

tomrf1:
This existing banner tool UI requires the user to choose between a "template" or a "design".
Screenshot 2024-06-26 at 10 18 34

We now only use the designs (created from the RRCP design tool).
The templates are legacy banners and can be removed.

This PR simplifies things by removing support for the legacy templates from the model and the UI -
Screenshot 2024-06-26 at 10 09 44

### Model changes

Before, the `BannerUi` model supported either a "template" or a "design":
```
export enum BannerTemplate {
  ContributionsBanner = 'ContributionsBanner',
  ...
}

export interface BannerDesignName {
  designName: string;
}

export type BannerUi = BannerTemplate | BannerDesignName;
```

Now, `BannerUi` only supports a design name:
```
export interface BannerUi {
  designName: string;
}
```

The model in SDC still supports both, but we can change that later.


public/src/components/channelManagement/bannerTests/bannerDesignSelector.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+interface BannerDesignSelectorProps {
+  designName: string;
+  onUiChange: (updatedUi: BannerUi) => void;
+  editMode: boolean;
+  designs: BannerDesign[];
+  onValidationChange: (isValid: boolean) => void;
+}
+
+const BannerDesignSelector: React.FC<BannerDesignSelectorProps> = ({
```
tomrf1, line 14: replaces `bannerUiSelector.tsx` with a much simpler UI

charleycampbell (approved):
This is a great piece of RRCP health work! I am just updating the guide with the various bits which have been added recently so will update the doc to reflect this change too

PR #619, opened 2024-07-18
Block unknown http methods

tomrf1:
Bots regularly hit the loadbalancer probing for vulnerabilities. Most result in a 4XX from the server.
But Play handles unknown http methods by returning a 501. This is a problem because it triggers our 5XX alarm. We don't want to change the threshold for that alarm as in general we _do_ want to know when the server returns errors.

E.g. from the logs:
`Illegal request, responding with status '501 Not Implemented': Unsupported HTTP method: IOLFJZ`

This PR adds a loadbalancer target rule to only route requests with supported http methods to the server. Anything else will receive a 400.

With invalid methods we now get a 400, not 501:
Screenshot 2024-07-19 at 13 01 15

Valid methods still go through to the server:
Screenshot 2024-07-19 at 13 02 54

### DIFF
--- a/cdk/lib/__snapshots__/admin-console.test.ts.snap
+++ b/cdk/lib/__snapshots__/admin-console.test.ts.snap
@@ -144,6 +144,38 @@ exports[`The AdminConsole stack matches the snapshot 1`] = `
       },
       "Type": "AWS::SSM::Parameter",
     },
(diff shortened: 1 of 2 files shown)


charleycampbell (approved):
hurrah!

PR #622, opened 2024-08-15
Use DCR storybook for banner live preview

tomrf1:
We've just [moved the banner components over to DCR](https://github.com/guardian/dotcom-rendering/pull/12130).

Currently the banner tool's live preview feature uses the SDC version of the banner component. Instead it should use the relevant DCR storybook, in an iframe. This is how the epic preview works.

I've refactored some existing code from the epic preview into a new file called `dcrStorybook.ts`, for reuse.


public/src/components/channelManagement/bannerTests/bannerVariantPreview.tsx
```diff
@@ -114,16 +114,9 @@ const buildProps = (
...
-  DesignableBanner: {
-    path: 'designableBanner/DesignableBanner.js',
-    name: 'DesignableBanner',
-  },
-};
-
 const useStyles = makeStyles(({ palette, spacing }: Theme) => ({
   drawer: {
-    height: 'auto',
+    height: '400px',
```
tomrf1, line 119: the downside to using an iframe is that we have to take a guess at the height. It will be scrollable within the iframe if the banner is taller than this

charleycampbell (approved):
Nice 👍

PR #636, opened 2024-10-11
Banner test deploySchedule

tomrf1:
Associated SDC change: https://github.com/guardian/support-dotcom-components/pull/1235

Marketing want the ability to override the channel-wide deploy schedule.
For example: channel 2 deploys weekly, but they'd like to show an app install banner (for mobile users) every 24 hours.

The solution implemented here is to add a `deploySchedule` to the `BannerTest` model, to allow Marketing users to override the channel-wide deploy schedule. When set, it allows them to set the number of days between deploys per user. This uses the user's `lastClosedAt` timestamp to decide when to next deploy the banner.

Screenshot 2024-10-11 at 15 29 50

### DIFF
--- a/app/models/BannerTests.scala
+++ b/app/models/BannerTests.scala
@@ -26,6 +26,8 @@ case class BannerVariant(
     tickerSettings: Option[TickerSettings] = None,
 )
 
+case class BannerTestDeploySchedule(daysBetween: Int)
+
 case class BannerTest(
     name: String,
     channel: Option[Channel],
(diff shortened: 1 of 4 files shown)


shtukas (approved):

sookburt:
Just based on the screenshot above, is it clear what the number means to the users?

tomrf1:
> Just based on the screenshot above, is it clear what the number means to the users?

good spot - that was the wrong screenshot! I've just updated it

charleycampbell (approved):

PR #640, opened 2024-10-24
Test methodologies editor

tomrf1:
This PR updates the methodologies selector to enable users to compare them, e.g. AB test vs epsilon-greedy.
See also the associated SDC PR: https://github.com/guardian/support-dotcom-components/pull/1251

If more than one methodology is configured then the the audience is split evenly between the methodologies. Each methodology is tracked separately with a different test name. This is done using the new `testName` field on the `Methodology` model.

I've also added the constraint that you cannot change the methodologies once a test is live. This is because it could change the test names, which would break tracking as well as the bandit calculations!

### Single methodology:
Screenshot 2024-11-22 at 15 53 31

### Two methodologies:
Screenshot 2024-11-22 at 15 53 44


app/models/BannerTests.scala
```diff
@@ -44,9 +44,9 @@ case class BannerTest(
     deviceType: Option[DeviceType] = None,
     campaignName: Option[String] = Some("NOT_IN_CAMPAIGN"),
     signedInStatus: Option[SignedInStatus] = Some(SignedInStatus.All),
-    isBanditTest: Option[Boolean] = None,
     consentStatus: Option[ConsentStatus] = Some(ConsentStatus.All),
     deploySchedule: Option[BannerTestDeploySchedule] = None,
+    methodologies: List[Methodology] = Nil
```
tomrf1, line 49: although the `methodologies` is mandatory, it's defaulted to `Nil` (an empty list), meaning validation does not fail when fetching existing data from dynamodb

public/src/components/channelManagement/TestMethodologyEditor.tsx
```diff
@@ -0,0 +1,202 @@
...
+    '& > button': {
+      height: '100%',
+    },
+  },
+  testName: {
+    marginRight: spacing(2),
+    fontWeight: 500,
+  },
+  error: {
+    color: 'red',
```
charleycampbell, line 53: I don't I've noticed we can use the actual colour name before!

public/src/components/channelManagement/TestMethodologyEditor.tsx
```diff
@@ -0,0 +1,202 @@
...
+  },
+}));
+
+const defaultEpsilonGreedyBandit: Methodology = {
+  name: 'EpsilonGreedyBandit',
+  epsilon: 0.1,
```
charleycampbell, line 59: Do we need any documentation to help support setting the epsilon value? Or do marketing already have that knowledge?

public/src/components/channelManagement/TestMethodologyEditor.tsx
```diff
@@ -0,0 +1,202 @@
...
+            Epsilon-greedy bandit
+          </MenuItem>
+        </Select>
+      </div>
+      <div>
+        {methodology.name === 'EpsilonGreedyBandit' && (
+          <div>
+            <TextField
+              type={'number'}
+              InputProps={{ inputProps: { min: 0, max: 1, step: 0.1 } }}
```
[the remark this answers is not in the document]
tomrf1: Good question. It would pick the "best" variant 100% of the time, which is not useful! I'll set min to 0.1

public/src/components/channelManagement/TestMethodologyEditor.tsx
tomrf1: 👍 we can document it in the RRCP user guide

app/models/BannerTests.scala
tomrf1: In general I try to avoid `Option[List[T]]`, because you have the following possible states:
1. `None`
2. empty list
3. non-empty list
And 1 and 2 sort of mean the same thing!
If we make the field optional then the client has to handle that possibility as well.

charleycampbell: ahhh good point!

public/src/components/channelManagement/TestMethodologyEditor.tsx
```diff
@@ -0,0 +1,203 @@
...
+              ...methodologies.slice(0, idx),
+              ...methodologies.slice(idx + 1),
+            ];
+            onChange(updatedMethodologies);
+          }}
+        />
+      ))}
+      <Button
+        onClick={onAddClick}
+        disabled={isDisabled || methodologies.length >= 4}
```
andrewHEguardian, line 195: Is 4 significant anywhere else? Seems like a sensible maximum though

andrewHEguardian (approved):
Looks great! 😀

PR #641, opened 2024-10-30
Test methodologies editor (simpler)

tomrf1:
This is a simpler version of https://github.com/guardian/support-admin-console/pull/640
because there are some complexities around the test names that I want more time to work through.
For now, this change lets users choose a single methodology for a test. The model still supports more than 1 methodology.
The last commit is the difference between the two branches.

Full explanation in the associated SDC PR: https://github.com/guardian/support-dotcom-components/pull/1236

Adds a new `methodologies` field, with a new editor.
Screenshot 2024-10-30 at 15 29 31
Screenshot 2024-10-30 at 15 29 35


charleycampbell (approved):
Nice 👍

PR #642, opened 2024-11-06
Add heading size editor to banner design tool

sookburt:
## What does this change?

This adds the option to have a small headline size in a banner design.  It is intended to be used for mobile banners with no copy but it is difficult to add such a restriction since the design is created separately from (and before) the copy.

Changes to the model are reflected in the following PRs which need to be merged before this (in the following order): 
- support-dotcom-components [PR1245](https://github.com/guardian/support-dotcom-components/pull/1245)
- dotcom-rendering to bump the version of SDC [PR12834](https://github.com/guardian/dotcom-rendering/pull/12834)
- dotcom-rendering to use the new model [PR12835](https://github.com/guardian/dotcom-rendering/pull/12835)

Tiny banners will have smaller headline font sizes which means less wrapping of longer headlines.

There are some risks of the small headline size being used on banners with copy - as this headline size is nearly the same size as the copy, it will be less obvious and may have a knock on effect on revenue.  MRR will, however, be able to review the banner before making it live which should help to reduce the risk.

## Images

Here is the new section in the banner design tool: 

Screenshot 2024-11-07 at 16 51 48

## Accessibility

This is using the same mechanism as other, similar, items in the page.

### DIFF
--- a/app/models/BannerDesign.scala
+++ b/app/models/BannerDesign.scala
@@ -23,6 +23,15 @@ case class HeaderImage(
   altText: String
 )
 
+// case class FontSize // is this needed?
+case class FontSize (
+  size: String
+)
+case class Font (
+  heading: FontSize
+)
+
+
 sealed trait BannerDesignVisual
 object BannerDesignVisual {
   case class Image(
@@ -124,4 +133,5 @@ case class BannerDesign(
   headerImage: Option[HeaderImage],
   colours: BannerDesignColours,
   lockStatus: Option[LockStatus],
+  fonts: Option[Font],
 )
--- a/app/models/BannerDesign.scala
+++ b/app/models/BannerDesign.scala
@@ -23,7 +23,6 @@ case class HeaderImage(
   altText: String
 )
 
-// case class FontSize // is this needed?
 case class FontSize (
   size: String
 )
--- a/app/models/BannerDesign.scala
+++ b/app/models/BannerDesign.scala
@@ -23,13 +23,17 @@ case class HeaderImage(
   altText: String
 )
 
-case class FontSize (
-  size: String
-)
-case class Font (
-  heading: FontSize
-)
+sealed trait FontSize
+object FontSize {
+  case object small extends FontSize
+  case object medium extends FontSize
+  case object large extends FontSize
 
+  implicit val decoder: Decoder[FontSize] = deriveEnumerationDecoder[FontSize]
+  implicit val encoder: Encoder[FontSize] = deriveEnumerationEncoder[FontSize]
+}
+case class Font(size: FontSize)
+case class Fonts(heading: Font)
 
 sealed trait BannerDesignVisual
 object BannerDesignVisual {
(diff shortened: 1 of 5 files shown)


app/models/BannerDesign.scala
```diff
@@ -124,4 +132,5 @@ case class BannerDesign(
   headerImage: Option[HeaderImage],
   colours: BannerDesignColours,
   lockStatus: Option[LockStatus],
+  fonts: Option[Font],
```
tomrf1, line 135: A better way to match the typescript model would be:
```
sealed trait FontSize
object FontSize {
  case object small extends FontSize
  case object medium extends FontSize
  case object large extends FontSize

  implicit val decoder: Decoder[FontSize] = deriveEnumerationDecoder[FontSize]
  implicit val encoder: Encoder[FontSize] = deriveEnumerationEncoder[FontSize]
}
case class Font(size: FontSize)
case class Fonts(heading: Font)
```

And then:
`fonts: Option[Fonts]`

public/src/components/channelManagement/bannerDesigns/BannerDesignForm.tsx
```diff
@@ -117,6 +118,17 @@ const BannerDesignForm: React.FC<Props> = ({
     });
   };
 
+  const onHeadlineSizeChange = (headerSize: 'small' | 'medium' | 'large' = 'medium'): void => {
```
tomrf1, line 121: we can use the `FontSize` type here. 
Also, I don't think the default value is necessary since the `HeadlineSizeEditor` always provides a value

public/src/components/channelManagement/bannerDesigns/HeadlineSizeEditor.tsx
```diff
@@ -0,0 +1,56 @@
+import { FormControl, FormControlLabel, RadioGroup, Radio } from '@mui/material';
+import React from 'react';
+import { FontSize } from '../../../models/bannerDesign';
+
+interface Props {
+  headerSize?: FontSize;
+  isDisabled: boolean;
+  onValidationChange: (fieldName: string, isValid: boolean) => void;
```
tomrf1, line 8: I don't think we need validation in this component as it uses a radio group

public/src/components/channelManagement/bannerDesigns/BannerDesignForm.tsx
sookburt: Thanks - good spot

public/src/components/channelManagement/bannerDesigns/HeadlineSizeEditor.tsx
```diff
@@ -0,0 +1,56 @@
...
+          defaultValue={'medium'}
+          value={headerSize === 'small' ? 'small' : 'medium'}
+          onChange={onRadioGroupChange}
+        >
+          <FormControlLabel
+            name="small"
+            value="small"
+            key="small"
+            control={<Radio />}
+            label="Small heading size - use for tiny banners without copy"
```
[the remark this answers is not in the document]
sookburt: Yep! Thanks!

public/src/components/channelManagement/bannerDesigns/HeadlineSizeEditor.tsx
sookburt: Good point - thanks!

tomrf1 (approved):
👍 great!

PR #646, opened 2024-11-19
Dependency upgrades to include remirror

sookburt:
## What does this change?

Remirror required upgrading owing to a high vulnerability on it's transitive dependency on Marked. Owing to all the interdependencies, it's easier to upgrade everything then test thoroughly and then setup dependabot to keep us up-to-date on a more regular basis.

## How to test

We will be mobbing to test against main. 
It would be useful to later add some UI tests that dependabot can run with each upgrade to help us stay current.

The rich text editor still works and nothing is broken.


sookburt:
Decided on a more gentle approach as it is a bit safer.  [PR to upgrade remirror to v2 here](https://github.com/guardian/support-admin-console/pull/658)

PR #650, opened 2024-11-29
Only show signed-in status selector where available

tomrf1:
we're currently showing this feature on the AMP + Apple News epic tools, where it's not available.


charleycampbell (approved):
Lovely tidy up!

PR #652, opened 2024-12-04
Support more than one google group for auth

tomrf1:
We want to allow other groups (24/7 Support, Central Production) to access the RRCP. This is not currently possible - users have to be in the support.admin.console google group.

### Before the change:
The server fetches a list of google groups from config. When logging in, a user must be a member of _all_ the configured groups. This list includes a special group for enforcing 2fa, since users must also be in this group.

### After the change:
When making any authorised request, a user must be in _at least one_ of the configured groups.
Additionally, we also still need to check that the user is in the special 2fa group. So I have moved this into a separate config item, so that we can check membership of this group separately from the other list of groups.

### The change
Previously the group checking was done in `Login.scala` at the point that a user logs in, with the line:
`processOauth2Callback(requiredGoogleGroups, googleGroupChecker)`
The google auth library would then use the `googleGroupChecker` to do the check - but [this method](https://github.com/guardian/play-googleauth/blob/d77d3bb8d9842a84229df70ed1f8613c7113f3d4/play-v29/src/main/scala/com/gu/googleauth/actions.scala#L219) requires a user to be in all groups, which is not what we want.

Instead, we now do the group checking as part of the `authAction`, which is created in `AppComponents.scala`. This method of group checking[ only requires one group to match](https://github.com/guardian/play-googleauth/blob/d77d3bb8d9842a84229df70ed1f8613c7113f3d4/play-v29/src/main/scala/com/gu/googleauth/actions.scala#L243).

It's a bit confusing that the play google-auth library behaves in two different ways like this. But I can see examples of both methods in our repos.

### Testing
Deployed to CODE, and temporarily removed a user from the support.admin.console google group to confirm their access was lost.

### DIFF
--- a/app/controllers/Login.scala
+++ b/app/controllers/Login.scala
@@ -1,6 +1,6 @@
 package controllers
 
-import com.gu.googleauth.{GoogleAuthConfig, GoogleGroupChecker, LoginSupport}
+import com.gu.googleauth.{AuthAction, Filters, GoogleAuthConfig, GoogleGroupChecker, LoginSupport}
 import play.api.libs.ws.WSClient
 import play.api.mvc._
 
@@ -34,7 +34,8 @@ class Login(
    * Looks up user's identity via Google
    */
   def oauth2Callback: Action[AnyContent] = Action.async { implicit request =>
-    processOauth2Callback(requiredGoogleGroups, googleGroupChecker)
+//    processOauth2Callback(requiredGoogleGroups, googleGroupChecker)
+    processOauth2Callback()
   }
 
   def logout: Action[AnyContent] = Action { implicit request =>
--- a/app/controllers/Login.scala
+++ b/app/controllers/Login.scala
@@ -1,6 +1,6 @@
 package controllers
 
-import com.gu.googleauth.{AuthAction, Filters, GoogleAuthConfig, GoogleGroupChecker, LoginSupport}
+import com.gu.googleauth.{GoogleAuthConfig, LoginSupport}
 import play.api.libs.ws.WSClient
 import play.api.mvc._
 
@@ -10,7 +10,6 @@ class Login(
   val authConfig: GoogleAuthConfig,
   val wsClient: WSClient,
   requiredGoogleGroups: Set[String],
-  googleGroupChecker: GoogleGroupChecker,
   components: ControllerComponents
 )(implicit executionContext: ExecutionContext)
   extends AbstractController(components) with LoginSupport {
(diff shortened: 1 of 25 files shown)


app/wiring/AppComponents.scala
```diff
@@ -15,9 +15,9 @@ import services.{Athena, Aws, CapiService, DynamoArchivedBannerDesigns, DynamoAr
 import software.amazon.awssdk.services.dynamodb.DynamoDbClient
 import software.amazon.awssdk.services.s3.model.GetObjectRequest
 
-class AppComponents(context: Context, stage: String) extends BuiltInComponentsFromContext(context) with AhcWSComponents with NoHttpFiltersComponents with AssetsComponents {
+class AppComponents(context: Context, stage: String) extends BuiltInComponentsFromContext(context) with AhcWSComponents with NoHttpFiltersComponents with AssetsComponents with Filters {
```
tomrf1, line 18: adding the `Filters` trait here gives us access to the `requireGroup` method

app/wiring/AppComponents.scala
```diff
@@ -50,7 +50,9 @@ class AppComponents(context: Context, stage: String) extends BuiltInComponentsFr
 
   private val requiredGoogleGroups: Set[String] = configuration.get[String]("googleAuth.requiredGroups").split(',').toSet
 
-  private val authAction = new AuthAction[AnyContent](authConfig, controllers.routes.Login.loginAction, controllerComponents.parsers.default)(executionContext)
```
tomrf1, line 53: here we adapt `authAction` to also do the group checking

app/controllers/Login.scala
```diff
@@ -34,7 +33,7 @@ class Login(
    * Looks up user's identity via Google
    */
   def oauth2Callback: Action[AnyContent] = Action.async { implicit request =>
-    processOauth2Callback(requiredGoogleGroups, googleGroupChecker)
+    processOauth2Callback()
```
tomrf1, line 36: this means we no longer check group membership during the initial google auth step, and we instead do it on every request

rupertbates (approved):
Looks reasonable to me 👍

app/wiring/AppComponents.scala
```diff
@@ -48,9 +48,15 @@ class AppComponents(context: Context, stage: String) extends BuiltInComponentsFr
     )
   }
 
+  private val twoFactorAuthEnforceGoogleGroup = configuration.get[String]("googleAuth.2faEnforceGroup")
   private val requiredGoogleGroups: Set[String] = configuration.get[String]("googleAuth.requiredGroups").split(',').toSet
```
johnduffell, line 52: probably should have renamed this to `allowedGroups` as they are not all required any more

tomrf1: That is a very good point - it would make it a lot clearer. I'm going to do this in a follow up PR

tomrf1: addressed here https://github.com/guardian/support-admin-console/pull/655

PR #658, opened 2024-12-09
Upgrade Remirror to v2

sookburt:
## What does this change?

Upgrades remirror, remirror/pm, remirror/react and Typescript to the next major version and removes the high vulnerability on Marked.

Also removes some minor vulnerabilities through the use of npm audit fix.

Tested locally in a banner and seems to work fine.

## Screenshot of testing

Screenshot 2024-12-12 at 15 21 29

### DIFF
--- a/public/src/components/channelManagement/richTextEditor/richTextEditor.tsx
+++ b/public/src/components/channelManagement/richTextEditor/richTextEditor.tsx
@@ -9,12 +9,11 @@ import {
   createMarkPositioner,
 } from 'remirror/extensions';
 import {
-  ComponentItem,
+  CommandButtonGroup,
   EditorComponent,
   FloatingToolbar,
   FloatingWrapper,
   Remirror,
-  ToolbarItemUnion,
   useActive,
   useAttrs,
   useChainedCommands,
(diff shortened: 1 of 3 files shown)


public/src/components/channelManagement/richTextEditor/richTextEditor.tsx
```diff
@@ -239,34 +236,36 @@ const FloatingLinkToolbar = () => {
...
-        positioner={linkPositioner}
-        placement="bottom"
-        enabled={!isEditing && empty}
-      />
+      <FloatingToolbar placement="top">
+        <CommandButtonGroup>
+          {activeLink ? (
+            <>
+              <button
+                className={`remirror-button ${active.bold() && 'remirror-button-active'}`}
```
tomrf1, line 246: this is my fault - but this can just be:
`className={'remirror-button'}`
I think this was copied from the bold button!

sookburt: Removed, good spot!

tomrf1 (approved):

PR #660, opened 2024-12-11
support more than 1 bandit methodology

tomrf1:
for showing the `BanditAnalyticsButton`


sookburt (approved):
The button shows fine - nice
On running locally the loading of the data took a very long time (assume that is _because_ I'm running it locally tho

PR #661, opened 2024-12-12
Use methodology test name for analytics data

tomrf1:
If there is more than one methodology then we give each methodology a separate test name, so that we can track them separately.
Currently the bandit analytics data feature uses the main test name, but it should use the methodology test name if available.
Currently this means that if there is more than one methodology then it queries for the wrong test name, and no data is displayed


charleycampbell (approved):
Great catch 🎣

PR #662, opened 2024-12-13
Make bandit sample count configurable

tomrf1:
[Related SDC PR](https://github.com/guardian/support-dotcom-components/pull/1260).

Marketing would like to be able to configure a test to only query the most recent N hours.
This PR adds an optional `sampleCount` field to the methodology model, to configure the number hourly samples.

In the tool this is displayed as a "Window" toggle with a numeric input. If the toggle is off, the input is disabled and there is no limit on the data. If the toggle is on, the number of hours is configurable, with a minimum of 6.

I've also changed how the test name is displayed in the methodologies part of the tool. Previously it displayed the entire name, which can get very long. Now it's just a button that copies the test name to clipboard.

The data returned by the "analytics" button also changes, because it needs to reflect scores with the configured sampleCount.

### Before:
Screenshot 2024-12-13 at 14 57 44

### After:
Screenshot 2024-12-13 at 14 56 50
Screenshot 2024-12-13 at 14 56 54

### DIFF
--- a/app/services/DynamoBanditData.scala
+++ b/app/services/DynamoBanditData.scala
@@ -42,7 +42,7 @@ object BanditData {
 
 class DynamoBanditData(stage: String, client: DynamoDbClient) extends StrictLogging {
   // No DEV table for bandit data
-  private val tableName = s"support-bandit-${if (stage == "PROD") "PROD" else "PROD"}"
+  private val tableName = s"support-bandit-${if (stage == "PROD") "PROD" else "CODE"}"
 
   private def query(testName: String, channel: String): ZIO[ZEnv, DynamoGetError, java.util.List[java.util.Map[String, AttributeValue]]] = {
     effectBlocking {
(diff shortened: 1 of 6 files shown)


app/services/DynamoBanditData.scala
```diff
@@ -41,7 +41,8 @@ object BanditData {
 }
 
 class DynamoBanditData(stage: String, client: DynamoDbClient) extends StrictLogging {
-  private val tableName = s"support-bandit-$stage"
+  // No DEV table for bandit data
+  private val tableName = s"support-bandit-${if (stage == "PROD") "PROD" else "CODE"}"
```
tomrf1, line 45: we're getting errors with the bandit analytics button when running locally

app/services/DynamoBanditData.scala
```diff
@@ -108,11 +110,11 @@ class DynamoBanditData(stage: String, client: DynamoDbClient) extends StrictLogg
               logger.error(s"Failed to decode item from Dynamo: ${error.getMessage}")
               None
           }
-          .toList
+          .toArray
```
tomrf1, line 113: Switched to using an array internally here because we now need to get subarrays (see the `.slice` above), which would be very inefficient with a List

PR #663, opened 2024-12-13
Add override for commons io vulnerability

sookburt:
## What does this change?

Adds a fix for the vulnerability on Apache commons-io that introduces a high vulnerabillity - [Apache Commons IO](https://github.com/guardian/support-admin-console/security/dependabot/130) which is introduced via org.playframework.  

Unfortunately, upgrading org.playframework to the latest version does not remove the vulnerable transient dependency, hence the override.

Also fixes an issue with the .gitignore file which ignored the plugins.sbt.

## How to test

Run sbt test

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,5 +9,6 @@ node_modules/
 .bloop/
 .metals/
 .bsp/
-project
+project/project
 dynamodb-local/
+project/metals.sbt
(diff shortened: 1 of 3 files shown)


shtukas (approved):

tomrf1 (approved):
👍

PR #666, opened 2024-12-20
upgrade eslint to 8.57.0 and some associated plugins

sookburt:
### What's done?

Upgrades eslint from 6.8 to 8.57 and bumps some associated plugins.
Works as it did before the upgrade.
NOTE: Two new rules that were added by the versions introduced in the upgrade have been switched off in the config file - these need to be discussed within the team and can perhaps be reintroduced if desired (remove the override in the rules object in the .eslintrc.js file and fix the resulting issues).

### What's left to bring it up to latest?

The eslint-plugin-prettier plugin depends on a prettier upgrade which has a dependency on remirror v3 so this has been held back for now but this doesn't seem to cause any issues.
The upgrade to eslint v9+ requires a big change to the configuration file (to a 'flat' file) as well as the relevant plugin bumps so will be done separately.

## How to test

Run locally - add a known linting issue and run npm run lint or npm run fix to see the output/changes.


.eslintrc.js
```diff
@@ -3,7 +3,7 @@ module.exports = {
   extends: [
     'plugin:@typescript-eslint/recommended',
     'plugin:react/recommended',
-    'prettier/@typescript-eslint',
+    // 'prettier/@typescript-eslint', // merged with prettier/recommended
```
tomrf1, line 6: shall we remove this line?

tomrf1 (approved):
great!

PR #667, opened 2025-01-09
Update play-googleauth to enable caching

tomrf1:


app/wiring/AppComponents.scala
```diff
@@ -28,7 +28,7 @@ class AppComponents(context: Context, stage: String) extends BuiltInComponentsFr
   }
 
   // https://github.com/guardian/play-googleauth#implement-googlegroups-based-access-control-using-the-directory-api
-  override def groupChecker = {
+  override val groupChecker = {
```
tomrf1, line 31: This change is also important.
When it's a `def` the `groupChecker` is recreated each time!
Changing it to a `val` means a single instance of the `groupChecker` is used for the lifetime of the server, which means all requests use the cache.

shtukas: Nice one!

shtukas (approved):

PR #672, opened 2025-01-20
Remove analytics button from epic tool

tomrf1:
Currently the epic tool has an "Analytics" button, near the top:
Screenshot 2025-01-20 at 15 32 45

This uses AWS Athena to query impressions data in S3. We're moving away from using this data source and it will be decommissioned.

We _could_ migrate it to use BigQuery instead.
But - we've recently added a separate "Analytics" button to the "Experiment Methodology" section, for Multi-armed Bandit tests:
Screenshot 2025-01-20 at 15 36 24
This uses the bandit dynamodb table, which is a much better source of data.
So I think the older "Analytics" button is redundant and can just be removed.


app/services/Athena.scala
```diff
@@ -2,8 +2,6 @@ package services
 
 import com.typesafe.scalalogging.StrictLogging
 import io.circe.{Decoder, Encoder}
-import models.GroupedVariantViews.VariantName
```
tomrf1, line 5: Note that this Athena file is still needed for the Super Mode page, which also queries athena data. I'll deal with this separately

charleycampbell (approved):

PR #675, opened 2025-01-22
Remove super mode charts, Athena

tomrf1:


app/models/SuperMode.scala
```diff
@@ -3,7 +3,16 @@ package models
...
-case class SuperModeRow(url: String, region: String, startTimestamp: String, endTimestamp: String, avPerView: Double)
+case class SuperModeRow(
+  url: String,
+  region: String,
+  startTimestamp: String,
+  endTimestamp: String,
+  avPerView: Double,
+  totalAv: Double,
+  totalViews: Double,
+  totalAcquisitions: Option[Double]
```
tomrf1, line 14: `totalAcquisitions` is new to the super mode data, so it's optional here

shtukas (approved):

PR #682, opened 2025-02-06
simplify locations deprecation

tomrf1:
Instead of handling both fields and merging them, we can initialise `regionTargeting` to have the country groups from `locations`.
This means `TestEditorTargetAudienceSelector` and `TestEditorTargetRegionsSelector` don't need to handle the old field


LAKSHMIRPILLAI (approved):
This all looks good

PR #683, opened 2025-02-10
Update Target Audience Layout

LAKSHMIRPILLAI:
## What does this change?

PR to update the  "Target Audience" section layout in RRCP tests

## Why do we need this?

Currently, the different sections under "Target Audience" are in multiple levels .The associated components, specifically the headings and radio buttons, are not maintaining their proper alignment.  This results in the headings and radio buttons appearing on separate lines causing user confusion. This pull request adjusts the layout.

## Before 
image

## After

image


tomrf1 (approved):
much better!

PR #687, opened 2025-02-17
landing page tool improvements

tomrf1:
1. removes a stray "="
2. moves `regionTargeting` field to the root of the model. This is because the existing generic `Test` model already has the field there
3. formats the scala

I've also updated the default copy to match what we have in support-frontend currently


charleycampbell (approved):
Looks good to me 👍

LAKSHMIRPILLAI (approved):
👍🏼

PR #688, opened 2025-02-18
Update to netty-handler version

charleycampbell:
## What does this change?
Bumps netty-handler to 4.1.94.Final as earlier versions have a high security vulnerability and this fixes the issue detailed in the now removed code comment

## How to test
Tested in CODE and everything appears to be working as usual

There could be issues but having checked functionality on the code environment I am confident that this is a non breaking change. Should anything appear out of the ordinary we can revert this PR.

## Accessibility


tomrf1:
We're actually no longer using athena in this project (see https://github.com/guardian/support-admin-console/pull/675). I should have removed it from the dependencies there!
If netty-handler is still only coming via athena, then instead of bumping it we can just remove the athena dependency and this override

tomrf1 (approved):
👍

PR #698, opened 2025-03-06
Add audit table

tomrf1:
For storing the audit history of each channel test.
Each create/update in the tools will add a new item in the audit table. The item will contain all the data from the updated test.
The partition key is composed of the channel name and the test name. This is necessary because DynamoDb limits us to a single partition key and a sort key, and we need the sort key to be the timestamp of the update (we do something like this in the multi-armed bandit table).

This is just the cdk change. Follow-up PRs will actually use this table.


charleycampbell (approved):
Nice 👍

PR #699, opened 2025-03-12
Unhide gutter liveblog menu item

sookburt:
## What does this change?

Now that the Gutter Liveblog Ask development is complete, this PR removes the code that was hiding the menu item so that the MRR Team can now access the test and variant editors.

## How to test

Deploy to code and the Gutter Liveblog Ask menu item should appear.


tomrf1 (approved):
🎉

PR #700, opened 2025-03-12
Write audits

tomrf1:
Uses the new channel tests audit table.
- adds the `DynamoChannelTestsAudit` service, which we pass into each channel tests Controller.
- changes the create and update handlers in the `ChannelTestsController` class. These now also write to the audit table. The entire channel test config is included in the audit record, along with the user's email and a timestamp.
- adds the `ttlInSecondsSinceEpoch` field to the table, which tells Dynamodb when to auto-expire items
- adds the `ChannelTestsAuditController`, which has a single endpoint for getting the audit history of a test. For now nothing will use this endpoint

This PR does not handle changes to the `status` field when going between draft/live. I'll do this in the separate PR as it involves fetching the current data.

### DIFF
--- a/app/controllers/ChannelTestsController.scala
+++ b/app/controllers/ChannelTestsController.scala
@@ -10,7 +10,7 @@ import models.DynamoErrors._
 import play.api.libs.circe.Circe
 import play.api.mvc._
 import services.S3Client.{S3ClientError, S3ObjectSettings}
-import services.{DynamoArchivedChannelTests, DynamoChannelTests, S3Json, VersionedS3Data}
+import services.{DynamoArchivedChannelTests, DynamoChannelTests, DynamoChannelTestsAudit, S3Json, VersionedS3Data}
 import utils.Circe.noNulls
 import zio.{IO, UIO, ZEnv, ZIO}
 import com.typesafe.scalalogging.LazyLogging
@@ -39,6 +39,7 @@ abstract class ChannelTestsController[T <: ChannelTest[T] : Decoder : Encoder](
   runtime: zio.Runtime[ZEnv],
   dynamoTests: DynamoChannelTests,
   dynamoArchivedTests: DynamoArchivedChannelTests,
+  dynamoTestsAudit: DynamoChannelTestsAudit
 )(implicit ec: ExecutionContext) extends AbstractController(components) with Circe with LazyLogging {
 
   private val lockObjectSettings = S3ObjectSettings(
@@ -162,6 +163,7 @@ abstract class ChannelTestsController[T <: ChannelTest[T] : Decoder : Encoder](
       logger.info(s"${request.user.email} is updating $channel/'${test.name}'")
       dynamoTests
         .updateTest(test, channel, request.user.email)
+        .flatMap(_ => dynamoTestsAudit.createAudit(test, request.user.email))
         .map(_ => Ok("updated"))
         .catchSome { case DynamoNoLockError(error) =>
           logger.warn(s"Failed to save $channel/'${test.name}' because user ${request.user.email} does not have it locked: ${error.getMessage}")
@@ -176,6 +178,7 @@ abstract class ChannelTestsController[T <: ChannelTest[T] : Decoder : Encoder](
       logger.info(s"${request.user.email} is creating $channel/'${test.name}'")
       dynamoTests
         .createTest(test, channel)
+        .flatMap(newTest => dynamoTestsAudit.createAudit(newTest, request.user.email))
         .map(_ => Ok("created"))
         .catchSome { case DynamoDuplicateNameError(error) =>
           logger.warn(s"Failed to create $channel/'${test.name}' because name already exists: ${error.getMessage}")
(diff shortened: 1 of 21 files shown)


app/controllers/ChannelTestsController.scala
```diff
@@ -162,6 +163,7 @@ abstract class ChannelTestsController[T <: ChannelTest[T] : Decoder : Encoder](
       logger.info(s"${request.user.email} is updating $channel/'${test.name}'")
       dynamoTests
         .updateTest(test, channel, request.user.email)
+        .flatMap(_ => dynamoTestsAudit.createAudit(test, request.user.email))
```
tomrf1, line 166: This means after the update has been written to the main table, we then write to the audit table.
`flatMap` ([from zio](https://zio.dev/overview/basic-operations/#chaining)) chains these operations together in sequence

app/services/DynamoArchivedBannerDesigns.scala
```diff
@@ -15,16 +15,6 @@ class DynamoArchivedBannerDesigns(stage: String, client: DynamoDbClient) extends
 
   protected val tableName = s"support-admin-console-archived-banner-designs-$stage"
 
-  private def put(putRequest: PutItemRequest): ZIO[ZEnv, DynamoError, Unit] =
```
tomrf1, line 18: Moved into the shared DynamoService class as it's used in several places

app/services/DynamoChannelTests.scala
```diff
@@ -227,11 +217,12 @@ class DynamoChannelTests(stage: String, client: DynamoDbClient) extends DynamoSe
       .map(allTests => allTests.flatMap(_.priority).maxOption.getOrElse(0))
 
   // Creates a new test, with bottom priority
-  def createTest[T <: ChannelTest[T] : Encoder : Decoder](test: T, channel: Channel): ZIO[ZEnv, DynamoError, Unit] =
+  def createTest[T <: ChannelTest[T] : Encoder : Decoder](test: T, channel: Channel): ZIO[ZEnv, DynamoError, T] =
```
tomrf1, line 220: we now return the item that we're creating, so that it can be written to the audit table as well

app/services/DynamoChannelTestsAudit.scala
```diff
@@ -0,0 +1,54 @@
...
+import zio.{ZEnv, ZIO}
+import models.DynamoErrors._
+import services.DynamoChannelTestsAudit.ChannelTestAudit
+import utils.Circe.jsonToDynamo
+
+import java.time.OffsetDateTime
+
+object DynamoChannelTestsAudit {
+  // The model that we write to the audit table
+  case class ChannelTestAudit[T <: ChannelTest[T] : Encoder : Decoder](
```
tomrf1, line 19: Here we support the creation of audits for any type of ChannelTest, and it must have a JSON encoder + decoder

cdk/lib/admin-console.ts
```diff
@@ -140,6 +140,7 @@ export class AdminConsole extends GuStack {
         name: 'timestamp',
         type: AttributeType.STRING,
       },
+      timeToLiveAttribute: 'ttlInSecondsSinceEpoch',
```
tomrf1, line 143: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/TTL.html

conf/routes
```diff
@@ -250,6 +250,9 @@ GET   /frontend/super-mode                             controllers.SuperModeCont
 # ----- analytics ----- #
 GET   /frontend/bandit/:channel/:testName              controllers.BanditDataController.getDataForTest(channel: String, testName: String)
 
+# ----- audit ----- #
+GET   /frontend/audit/:channel/:testName              controllers.ChannelTestsAuditController.getAuditsForChannelTest(channel: String, testName: String)
```
tomrf1, line 254: nothing uses this endpoint yet

app/services/DynamoChannelTestsAudit.scala
```diff
@@ -0,0 +1,93 @@
...
+import DynamoChannelTestsAudit.{ChannelTestAudit, getTimeToLive}
+import utils.Circe.{dynamoMapToJson, jsonToDynamo}
+import zio.blocking.effectBlocking
+
+import java.time.OffsetDateTime
+import scala.jdk.CollectionConverters.{ListHasAsScala, MapHasAsJava}
+
+object DynamoChannelTestsAudit {
+  // The model that we write to the audit table
+  case class ChannelTestAudit[T : Encoder : Decoder](
```
tomrf1, line 21: We can create an audit entry for any item of type `T` which has a JSON encoder + decoder

shtukas (approved):

cdk/lib/admin-console.ts
charleycampbell: Thanks for sharing the resource!

charleycampbell (approved):
Looks good to me - great unit test addition! Thanks for adding the relevant documentation to the PR 👍

PR #703, opened 2025-03-21
Write audits for status updates

tomrf1:
A [previous PR](https://github.com/guardian/support-admin-console/pull/700) began writing to the audit table for each create + update request.

There was still one case that was not audited: updating the status of tests (between draft/live).
This is a bit more complex because we have to first fetch the full data for the test before writing to the audit table.
Also, the status update endpoint can take a set of test names, so we have to fetch and write in batch.

### DIFF
--- a/app/controllers/ChannelTestsController.scala
+++ b/app/controllers/ChannelTestsController.scala
@@ -246,6 +246,21 @@ abstract class ChannelTestsController[T <: ChannelTest[T] : Decoder : Encoder](
 
         case Some(status) =>
           dynamoTests.updateStatuses(testNames, channel, status)
+            .flatMap(_ => {
+              /**
+               * Fetch the full data for all the updated tests and then write audits.
+               * We use .forkDaemon here to avoid delaying the response to the client
+               * and instead run this task in the background.
+               */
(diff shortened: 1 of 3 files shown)


app/controllers/ChannelTestsController.scala
```diff
@@ -246,6 +246,21 @@ abstract class ChannelTestsController[T <: ChannelTest[T] : Decoder : Encoder](
 
         case Some(status) =>
           dynamoTests.updateStatuses(testNames, channel, status)
+            .flatMap(_ => {
+              /**
+               * Fetch the full data for all the updated tests and then write audits.
+               * We use .forkDaemon here to avoid delaying the response to the client
+               * and instead run this task in the background.
+               */
```
tomrf1, line 254: See https://zio.dev/reference/fiber/fiber.md/#fork-in-global-scope-daemon
We want this task to run in the background, independent of the response to the client.
"So if the parent fiber terminates, the daemon fiber will not be terminated. It will only will be terminated when the global scope is closed, or its life end naturaly."

LAKSHMIRPILLAI (approved):

PR #704, opened 2025-03-24
Display change audit

LAKSHMIRPILLAI:
## What does this change?

This PR is to create a new audit page for viewing the audit history

[Trello card](https://trello.com/c/ndi5MCWL/983-rrcp-display-change-audit)

### DIFF
--- a/public/src/components/drawer.tsx
+++ b/public/src/components/drawer.tsx
@@ -210,6 +210,11 @@ export default function NavDrawer(): React.ReactElement {
             <span className={classes.super}>🎨</span>
           </ListItem>
         </Link>
+        <Link key="Audit Tests" to="/audit-tests" className={classes.link}>
+          <ListItem className={classes.listItem} button key="Audit Tests">
+            <ListItemText primary="Audit Tests" />
+          </ListItem>
+        </Link>
       </div>
 
       <div>
--- a/public/src/main.tsx
+++ b/public/src/main.tsx
@@ -41,6 +41,7 @@ import DefaultPromos from './components/defaultPromos';
 import { StyledEngineProvider } from '@mui/material';
 import { LinkTrackingBuilder } from './components/linkTracking/LinkTrackingBuilder';
 import { SupportLandingPageTestsForm } from './components/channelManagement/supportLandingPage/supportLandingPage';
+import { AuditTestsDashboard } from "./components/channelManagement/auditTests/auditTestsDashboard";
 
 declare module '@mui/styles' {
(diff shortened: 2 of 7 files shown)


public/src/components/drawer.tsx
```diff
@@ -210,6 +210,11 @@ export default function NavDrawer(): React.ReactElement {
             <span className={classes.super}>🎨</span>
           </ListItem>
         </Link>
+        <Link key="Audit Tests" to="/audit-tests" className={classes.link}>
+          <ListItem className={classes.listItem} button key="Audit Tests">
+            <ListItemText primary="Audit Tests" />
```
tomrf1, line 215: I think "Test audits" would be a better name, since this is where we display the audit history of tests

LAKSHMIRPILLAI: Done 👍🏼

public/src/main.tsx
```diff
@@ -231,6 +232,10 @@ const AppRouter = () => {
             path="/banner-designs/:name?"
             element={createComponent(<BannerDesigns />, 'Banner Designs')}
           />
+          <Route
+            path="/audit-tests"
+            element={createComponent(<AuditTestsDashboard />, 'Audit Tests')}
```
tomrf1, line 237: this string sets the page title, so let's also make this "Test Audits"

pnpm-lock.yaml
```diff
@@ -2340,6 +2343,14 @@ packages:
       prosemirror-inputrules: 1.5.0
       prosemirror-keymap: 1.2.2
       prosemirror-model: 1.25.0
+      prosemirror-paste-rules: 3.0.0(prosemirror-model@1.25.0)(prosemirror-state@1.4.3)(prosemirror-view@1.39.1)
```
tomrf1, line 2346: there's some duplication here - try running `pnpm install`

public/src/components/channelManagement/auditTests/auditTestsTable.tsx
```diff
@@ -0,0 +1,132 @@
...
+        const onClick = () => {
+          const version: number = +params.row.index;
+          const jsonDiff = getJsonDiff(version);
+          setOpen(true);
+          setJsonDiff(jsonDiff);
+          setVersionToCompare(sortedRows.length - version);
+        };
+        //To hide and disable  the button for the last row
+        if (params.row.index === (sortedRows.length - 1).toString()) {
+          return <Button disabled></Button>;
```
tomrf1, line 94: could just return null if it's the first version

tomrf1 (approved):
great work!

PR #706, opened 2025-03-26
Field validation in the landing page tool

tomrf1:
This PR introduces react-hook-form to the landing page tool. We use this library elsewhere in the RRCP for field validation. This means the library now has control over the value and validation of the fields in the tool.

We now validate:
1. heading + subheading length
2. product card pill length
3. benefit copy length
4. number of benefits (max 8). We do this by disabling the button for adding benefits

Screenshot 2025-03-26 at 11 01 23

Screenshot 2025-03-26 at 11 46 28
Screenshot 2025-03-26 at 11 46 57


charleycampbell (approved):
👍

PR #711, opened 2025-04-03
remirror v3

tomrf1:
Migrates to the latest remirror version, which has the stated aim:

> This version aims to make it easier to use Remirror in existing applications, by not imposing our architectural decisions on to you.

And it is a simple upgrade!


charleycampbell (approved):
👍

PR #712, opened 2025-04-03
Replace react-beautiful-dnd

tomrf1:
We currently use react-beautiful-dnd for the drag+drop behaviour of the tests list.
This library hasn't been supported for a while and is deprecated.
After reviewing the options listed [here](https://github.com/atlassian/react-beautiful-dnd/issues/2672), I chose https://docs.dndkit.com/ as a replacement. It's relatively similar and simple to use.

The behaviour of the test list editor is the same with this branch.

### DIFF
--- a/public/src/components/channelManagement/testList.tsx
+++ b/public/src/components/channelManagement/testList.tsx
@@ -1,7 +1,9 @@
 import React from 'react';
 import { List } from '@mui/material';
 import { makeStyles } from '@mui/styles';
-import { DragDropContext, Droppable, Draggable, DropResult } from 'react-beautiful-dnd';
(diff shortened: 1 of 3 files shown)


public/src/components/channelManagement/testList.tsx
```diff
@@ -17,6 +19,48 @@ const useStyles = makeStyles(() => ({
...
+
+const TestListItem = <T extends Test>({
+  test,
+  isSelected,
+  isInEditMode,
+  onTestSelected,
+}: TestListItemProps<T>) => {
+  const { setNodeRef, attributes, listeners, transform, transition } = useSortable({
+    id: test.name,
+  });
```
tomrf1, line 37: this hook makes the item draggable and sorted

shtukas (approved):

PR #713, opened 2025-04-03
Use pnpm

tomrf1:
Replaces the use of npm and a package-lock.json with pnpm, which is now the department standard.
Also replaces yarn in the cdk directory.


package.json
```diff
@@ -27,6 +27,7 @@
   "dependencies": {
     "@dnd-kit/core": "^6.3.1",
     "@dnd-kit/sortable": "^10.0.0",
+    "@dnd-kit/utilities": "^3.2.2",
```
tomrf1, line 30: needs to be an explicit dependency with pnpm

package.json
```diff
@@ -40,6 +41,7 @@
     "immutability-helper": "^3.0.1",
     "lodash": "^4.17.21",
     "lz-string": "^1.5.0",
+    "prosemirror-state": "^1.4.3",
```
tomrf1, line 44: needs to be an explicit dependency with pnpm

package.json
```diff
@@ -62,7 +64,7 @@
     "build-dev": "webpack --config webpack.dev.js",
     "watch": "webpack --config webpack.dev.js --watch",
     "lint": "eslint .",
-    "fix": "npm run lint -- --fix",
```
tomrf1, line 65: removing `fix` as I think we're all used to running `pnpm lint --fix`

shtukas (approved):

PR #715, opened 2025-04-07
Add permissions table

tomrf1:
We want to add more granular permissioning to the RRCP.

We use Google auth to handle authentication (who a user is).
We will now build a simple bespoke solution for authorization (which features a user has access to).

The new table will have one row per user, keyed by email address.
This PR just adds the new table to the cdk script.


charleycampbell (approved):
👍

PR #718, opened 2025-04-09
Permissioning of landing page tool endpoints

tomrf1:
Uses the data from the permissions table to restrict access to the landing page tool endpoints.
It does this by introducing a new Play [Action](https://www.playframework.com/documentation/3.0.x/ScalaActions).

1. Adds the `PermissionsAction` Action, which ensures a user has the right permission, and returns a 403 if not.
2. Updates the `ChannelTestsController`, which is used by the landing page tool and all of the channel test tools. Previously this class received a single `authAction`, for google auth checks on all the endpoints. It now takes an instance of `AuthAndPermissionActions`, which contains separate Actions for `read` and for `write`. This means endpoints for reading and writing data now have separate Actions, with different permissions checks.
3. For now, only the landing page tool has permissions checks. The other channel tools do not

Note - this is all server-side for now. If a user without edit permission tries to edit a landing page test they'll get an error popup.
In a follow-up PR we'll update the client to disable the Edit button in the UI.

### DIFF
--- a/app/actions/PermissionAction.scala
+++ b/app/actions/PermissionAction.scala
@@ -43,6 +43,8 @@ object PermissionAction extends Results with LazyLogging {
     Forbidden(message)
   }
 
+  // In the filter function we return None if the user has permission to access the resource,
+  // otherwise we return a 403 Forbidden to the client
   def checkPermission(
     email: String,
     page: String,
@@ -73,9 +75,6 @@ class PermissionAction(
   val parse: PlayBodyParsers,
   val executionContext: ExecutionContext,
 ) extends ActionFilter[UserIdentityRequest] {
-
-  // In the filter function we return None if the user has permission to access the resource,
-  // otherwise we return a 403 Forbidden to the client
   override protected def filter[A](request: UserIdentityRequest[A]): Future[Option[Result]] = Future.successful {
     checkPermission(
       request.user.email,
--- a/app/actions/PermissionAction.scala
+++ b/app/actions/PermissionAction.scala
@@ -72,7 +72,7 @@ class PermissionAction(
   permissionsService: DynamoPermissionsCache,
   val parse: PlayBodyParsers,
   val executionContext: ExecutionContext,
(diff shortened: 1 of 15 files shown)


shtukas (approved):

app/controllers/SupportLandingPageController.scala
```diff
@@ -21,9 +22,24 @@ class SupportLandingPageController(
...
+    // all users have read access
+    readPermissionAction = None,
+    // users must have write access to make changes
+    writePermissionAction = Some(
+      new PermissionAction(
+        page = SupportLandingPageController.name,
+        requiredPermission = Permission.Write,
+        permissionsService,
+        components.parsers,
+        ec
```
charleycampbell, line 39: I know the naming of this wasn't part of this change but It would be clearer to have `ExecutionContext`...maybe one to refactor at a later date

charleycampbell (approved):
Looks good - great to see some unit tests! Would we consider adding a section on the RRCP to manage this group? Or would we edit the dynamo table?

app/actions/PermissionAction.scala
```diff
@@ -0,0 +1,86 @@
...
+          case Some(userPermission) =>
+            userPermission.permission match {
+              case Permission.Write => None // user has full access
+              case Permission.Read =>
+                if (requiredPermission != Permission.Write) None
+                else Some(logAndForbid(s"Invalid permission for user ${email}, for page $page"))
+            }
+          case None => Some(logAndForbid(s"No permission found for user ${email}, for page $page"))
+        }
+      case None => Some(logAndForbid(s"No permissions found for user ${email}"))
```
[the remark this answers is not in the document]
tomrf1: Hopefully the unit tests help here, but we return 403s in these cases:
1. no permissions at all for this user in the dynamodb table
2. no permissions for this resource, for this user, in the table
3. user has incorrect permission for this resource

app/controllers/SupportLandingPageController.scala
tomrf1: 👍 updated

PR #721, opened 2025-04-11
Disable buttons if user does not have write permission

tomrf1:
For now, only the Landing Page tool requires `Write` permission.
On the backend the relevant endpoints are already behind permissioning, so users without `Write` permission are already unable to make changes.

This PR updates the client to check the user's permission, and if they don't have write permission then the various buttons for editing are disabled.
We do this by passing through the user's permissions on the `window.guardian` object, for the client to use.

No permission check (Epic tool):
Screenshot 2025-04-11 at 08 34 12

With permission check and no write permission (Landing Page tool):
Screenshot 2025-04-11 at 08 34 16

### DIFF
--- a/public/src/components/channelManagement/stickyTopBar/stickyTopBar.tsx
+++ b/public/src/components/channelManagement/stickyTopBar/stickyTopBar.tsx
@@ -61,12 +61,14 @@ const useStyles = makeStyles(({ palette, spacing }: Theme) => ({
     },
     marginLeft: spacing(1),
   },
-  buttonText: {
-    fontSize: '14px',
-    fontWeight: 500,
-    textTransform: 'uppercase',
-    letterSpacing: '1px',
+  button: {
     color: palette.grey[800],
+    '& > p': {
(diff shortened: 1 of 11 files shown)


public/src/components/channelManagement/stickyTopBar/stickyTopBar.tsx
```diff
@@ -61,12 +61,14 @@ const useStyles = makeStyles(({ palette, spacing }: Theme) => ({
     },
     marginLeft: spacing(1),
   },
-  buttonText: {
-    fontSize: '14px',
-    fontWeight: 500,
-    textTransform: 'uppercase',
-    letterSpacing: '1px',
+  button: {
```
tomrf1, line 64: I've moved the color up to the <button> element, rather than the inner <p>, so that the color is automatically overridden when the button is disabled

charleycampbell (approved):
This looks really great Tom. Such a good (and important) feature to have in the RRCP

PR #722, opened 2025-04-11
Fetch audit history immediately if from url

tomrf1:
A small UX improvement.
If the user clicks through to the audit page then the channel + test name are in the url (e.g. `/audit-tests/Epic/TESTEPIC`).
Currently the user still has to click "Get audit" before the data is fetched.
This PR changes the audit page to immediately fetch the data in this case.


LAKSHMIRPILLAI (approved):
👍🏼

PR #725, opened 2025-04-22
Add ticker Config  in Support Landing Page Tests

LAKSHMIRPILLAI:
## What does this change?
This PR is to add ticker to the landing page tests

## Images

We will have a new checkbox for ticker 

image

On selection, the ticker configuration would allow the user to set the headline copy, campaign name and the currency symbol to be set in LP tests

image

NOTE:

Ticker campaign "GLOBAL" won't allow used to set a currencySymbol as it show the supporter count instead of the money raised
image

The related changes in the support landing page are in  the [Support Frontend PR] (https://github.com/guardian/support-frontend/pull/6970)

### DIFF
--- a/app/models/Ticker.scala
+++ b/app/models/Ticker.scala
@@ -2,19 +2,10 @@ package models
 import io.circe.generic.extras.Configuration
 import io.circe.generic.extras.semiauto.{deriveEnumerationDecoder, deriveEnumerationEncoder}
 
-sealed trait TickerEndType
-object TickerEndType {
-  case object unlimited extends TickerEndType
-  case object hardstop extends TickerEndType
-
-  implicit val customConfig: Configuration = Configuration.default.withDefaults
-  implicit val encoder = deriveEnumerationEncoder[TickerEndType]
-  implicit val decoder = deriveEnumerationDecoder[TickerEndType]
-}
-
 sealed trait TickerCountType
 object TickerCountType {
   case object money extends TickerCountType
+  case object supporterCount extends TickerCountType
 
   implicit val customConfig: Configuration = Configuration.default.withDefaults
   implicit val encoder = deriveEnumerationEncoder[TickerCountType]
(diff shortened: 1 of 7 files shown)


public/src/components/channelManagement/tickerEditor.tsx
```diff
@@ -170,7 +158,7 @@ const TickerEditor: React.FC<TickerEditorProps> = ({
             fullWidth
           />
 
-          {tickerSettings.countType === 'money' && (
+          {(tickerSettings.name === 'US' || tickerSettings.name === 'AU') && (
```
tomrf1, line 161: 👍 this is good. If in the future we want to count supporters for US or AU then we can adjust this logic

app/models/Ticker.scala
```diff
@@ -33,13 +16,9 @@ object TickerName {
 
 case class TickerCopy(
   countLabel: String,
-  goalReachedPrimary: Option[String] , //deprecated
-  goalReachedSecondary: Option[String] //deprecated
 )
 
 case class TickerSettings(
-  endType: TickerEndType,
-  countType: TickerCountType,
```
tomrf1, line 42: `endType` and `countType` are unused by the clients, but I think this may cause validation to fail in SDC and DCR as they're still required there -
https://github.com/guardian/support-dotcom-components/blob/main/src/shared/types/props/shared.ts#L101

I think your options are:
1. first remove them in SDC (and update the dependency in DCR)
2. leave them in for now and tidy up later

LAKSHMIRPILLAI: PR  to remove endType and countType from TickerSettings  in SDC https://github.com/guardian/support-dotcom-components/pull/1335

tomrf1 (approved):

LAKSHMIRPILLAI:
@tomrf1 :I identified a small bug yesterday while doing testing and have updated the PR with fix. It was showing decoding failure while updating the heading for global.

PR #726, opened 2025-04-29
Changes for making goal copy in  ticker configurable

LAKSHMIRPILLAI:
## What does this change?

The PR is to make the goal copy in ticker configurable ie the word "goal" after 50,000 on this ticker
image

Here we are adding a new field  in ticker configuration for the 'goal' copy 
[Trello card](https://trello.com/c/bDmwMfwB/1069-make-goal-in-tickercopy-configurable) 
## Why?

 This is to change word "goal" after 50,000 on this ticker to "readers" to show that it's for supporter count

## Images

image

### DIFF
--- a/app/models/Ticker.scala
+++ b/app/models/Ticker.scala
@@ -16,6 +16,7 @@ object TickerName {
 
 case class TickerCopy(
   countLabel: String,
+  goalCopy: String = "goal" ,
 )
 
 case class TickerSettings(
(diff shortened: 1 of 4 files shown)


app/models/Ticker.scala
```diff
@@ -16,6 +16,7 @@ object TickerName {
 
 case class TickerCopy(
   countLabel: String,
+  goalCopy: String = "goal" ,
```
LAKSHMIRPILLAI, line 19: Added this so as to not break any tests  in Dynamo DB with ticker enabled

PR #729, opened 2025-05-12
Add countdown in support landing pages tests

LAKSHMIRPILLAI:
## What does this change?

This PR is to add countdown to the landing page tests

[Trello card](https://trello.com/c/ZSSk3jdO/1061-landing-page-tool-countdown-config)

## How to test

Tested in CODE

## Images

We will have a new checkbox for countdown

image

On selection, the countdown configuration would allow the user to set the label to overwrite the support site heading , start date, deadline date , counter background colour  and the foreground colour  to be set in LP tests

Note : Previously we used  the below hardcoded values in support frontend for countdown timer
background colour :#1e3e72
foreground colour: #ffffff 

Also the user would be provided with a toggle switch to select  either UTC or Local browser time
We have added the below helper text to guide the user

Local time: E.g. in a US campaign, to end at midnight in the user's local time. This means users on the east coast stop seeing the countdown before users on the west coast. And if a user on the west coast and a user on the east coast view the page at the same moment, they will each see a different number of hours remaining on the countdown.

UTC: E.g. for a promo code with a midnight expiry, which is always in UTC. This means all users will stop seeing the countdown at the same time, and it will show the same hours/minutes remaining. It also means e.g. if a user on the US east coast views the page at 19:00 local time, then the countdown says 1 hour remaining.

This will be saved in Dynamo DB table as below

image

The related changes in the support landing page are in the [Support Frontend PR](https://github.com/guardian/support-frontend/pull/7018)


app/models/Countdown.scala
```diff
@@ -0,0 +1,13 @@
+package models
+
+case class CountdownTheme(
+  backgroundColor: String,
+  foregroundColor: String,
+)
+
+case class CountdownSettings(
+  label: String,
+  countdownStartInMillis: String,
```
tomrf1, line 10: I don't think this name is correct?
e.g. from CODE `countdownStartInMillis: "2025-05-13T10:43:31"`

LAKSHMIRPILLAI: Oh , I used the same  values as we had used in support frontend before,  but  yes, it's misleading.I will update the names in both PRs ,thanks

PR #731, opened 2025-05-15
Improve timezone toggle UI

tomrf1:
Shows both options either side of the toggle, and moves the explanation into an `Alert` component
Screenshot 2025-05-15 at 15 44 10


LAKSHMIRPILLAI (approved):
👍🏼

PR #732, opened 2025-05-19
Upgrade react-hook-form to v7

tomrf1:
Migration guide: https://legacy.react-hook-form.com/migrate-v6-to-v7/
I've grouped similar changes by commit. The main changes are:

1. The `register` function is now used differently. Instead of `ref={register(...`, we now have to spread the register function: `{...register(...`
2. The `errors` field now lives under the `formState` object
3. Because calling `...register` gives us an `onBlur` function, we have to add our own `onBlur` after calling register, otherwise it is overridden
4. The `render` function parameters have changed, so instead of `render={data => ...`, we now use `render={({ field }) => ...`
5. Referencing field arrays has changed from e.g. `benefits[${index}].copy` to `benefits.${index}.copy`

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "react": "^18.2.0",
     "react-colorful": "^5.6.1",
     "react-dom": "^18.2.0",
-    "react-hook-form": "^6.15.8",
+    "react-hook-form": "^7.56.3",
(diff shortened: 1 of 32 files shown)


shtukas (approved):

sookburt (approved):
👍

PR #733, opened 2025-05-21
Choice cards override in epic variants

tomrf1:
Adds the ability to configure the choice cards per epic variant.

Before, there was just a checkbox to enable choice cards:
Screenshot 2025-06-03 at 14 49 50

And SDC would use the hardcoded "default" choice cards config ([defined here](https://github.com/guardian/support-dotcom-components/blob/main/src/server/lib/choiceCards/defaultChoiceCardSettings.ts)).

Now, in the epic tool's variant editor there are 3 choices:
Screenshot 2025-06-03 at 12 31 35

If the 3rd option is selected then the user can add up to 3 choice cards:

Screenshot 2025-06-03 at 12 33 12

Each choice card is an expandable Accordion component:
Screenshot 2025-06-03 at 12 34 31

Screenshot 2025-06-03 at 12 36 34

### Limitations
- this is only for epics for now, but can be added to the banner tool later.
- the epic "live preview" does not currently support choice cards. This is because the SDC server builds the final config for the choice cards (e.g. pricing, substituting currency templates), and the live preview passes the epic variant config directly to storybook. In order to make this work I think we'd need to make a call to SDC, which is not ideal.
- while SDC does support mobile-specific choice cards config, this PR does not introduce that capability. For now we're keeping it simple.

PR #738, opened 2025-06-09
Enable choice cards config in banner tool

tomrf1:
A previous PR enabled Marketing to configure the choice cards in the epic tool:
https://github.com/guardian/support-admin-console/pull/733

This means it overrides the "default" settings, which are hardcoded in SDC.

This PR does the same for the banner tool.
The difference here is that choice cards are enabled in the banner from the _banner design tool_.
So, whereas the epic variant model has a boolean for enabling choice cards, banner variants get this from the associated banner design.
This means that the banner variant editor only displays the choice cards settings component if the banner design has choice cards enabled.


LAKSHMIRPILLAI (approved):
👍🏼 Great!

PR #756, opened 2025-07-04
Add scalafmt

tomrf1:
Adds the [scalafmt plugin](https://scalameta.org/scalafmt/) and applies it to the codebase. For consistent formatting of the scala code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ node_modules/
 project/project
 dynamodb-local/
 project/metals.sbt
+dynamodb-local-metadata.json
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -12,3 +12,4 @@ node_modules/
 project/project
 dynamodb-local/
 project/metals.sbt
+dynamodb-local-metadata.json
```
tomrf1, line 15: unrelated to scalafmt, but I noticed this file was being added when running tests locally

build.sbt
```diff
@@ -9,10 +9,18 @@ val circeVersion = "0.14.14"
 val awsVersion = "2.31.60"
 val zioVersion = "2.1.19"
 
+lazy val scalafmtSettings = Seq(
+  scalafmtFilter.withRank(KeyRanks.Invisible) := "diff-dirty",
```
tomrf1, line 13: this means that when running the tests it will check the formatting of any files that have changed

PR #759, opened 2025-07-10
Set jackson version

tomrf1:
Play is still pulling in an old version of jackson with vulnerabilities.
See https://github.com/orgs/playframework/discussions/11222.
A number of our codebases are having to fix all jackson deps, e.g. https://github.com/guardian/frontend/pull/27447

Prompted by a broken scala steward PR - https://github.com/guardian/support-admin-console/pull/752

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,7 +7,7 @@ scalaVersion := "2.13.16"
 val circeVersion = "0.14.14"
 val awsVersion = "2.31.78"
 val zioVersion = "2.1.19"
-val jacksonVersion = "2.17.2"
+val jacksonVersion = "2.17.3"
 
 lazy val scalafmtSettings = Seq(
   scalafmtFilter.withRank(KeyRanks.Invisible) := "diff-dirty",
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -48,8 +49,14 @@ libraryDependencies ++= Seq(
 
 dependencyOverrides ++= List(
   // Play still uses an old version of jackson-core which has a vulnerability - https://security.snyk.io/vuln/SNYK-JAVA-COMFASTERXMLJACKSONCORE-7569538
-  "com.fasterxml.jackson.core" % "jackson-core" % "2.17.3",
-  "com.fasterxml.jackson.core" % "jackson-databind" % "2.14.3",
+  "com.fasterxml.jackson.core" % "jackson-core" % jacksonVersion,
+  "com.fasterxml.jackson.core" % "jackson-annotations" % jacksonVersion,
```
tomrf1, line 53: without all these overrides we get compilation errors like:
`com.fasterxml.jackson.databind.JsonMappingException: Scala module 2.14.3 requires Jackson Databind version >= 2.14.0 and < 2.15.0 - Found jackson-databind version 2.17.3`

johnduffell: FYI I think 2.15.4 is enough to get rid of the vulnerability.  But if it works then that's all good.

PR #762, opened 2025-07-17
Add choice cards configurable destination

juabara:
## What does this change?

Adds an optional `destinationUrl` field to the Choice Card configuration in `support-admin-console`.  
This enables designable banners to specify a custom URL for each individual choice card.  
All related work in the consuming repositories has already been completed.

## How to test

1. Go to the support-admin-console.
2. Add a new Choice Card or edit an existing one.
3. Enter a URL in the new `destinationUrl` field.
4. Save and verify that the value is stored correctly and appears in the form.

- Editors are able to set custom URLs per choice card without issues.
- The custom URLs appear correctly in the form.
- No regressions in existing banners that don’t use this field.

- Minor risk of breaking existing configurations if the field is mishandled.
- Mitigated by keeping the field optional and verifying backward compatibility.
- No impact on personal data or legal/compliance concerns.


tomrf1 (approved):
👍

PR #764, opened 2025-07-23
Increase cache duration for GoogleGroupChecker following timeouts

sookburt:
## What does this change?

This increases the duration of the cache for GoogleGroupChecker from the default of 1 minute to 1 hour.  We are doing this because we have had a series of alerts recently from support-admin-console which relate to timeouts and the common item in each stack trace is the GoogleGroupChecker.  This is a strategy to see if this is what is timing out.

## How to test

Deploy to CODE and attempt to use the RRCP UI - it should authenticate as normal

We should see a reduction in the number of timeout errors in alerts that we have been experiencing lately

### DIFF
--- a/app/wiring/AppComponents.scala
+++ b/app/wiring/AppComponents.scala
@@ -12,23 +12,10 @@ import play.api.libs.ws.ahc.AhcWSComponents
 import play.api.mvc.AnyContent
 import play.api.{BuiltInComponentsFromContext, NoHttpFiltersComponents}
 import router.Routes
-import services.{
-  Aws,
-  BigQueryService,
-  CapiService,
-  DynamoArchivedBannerDesigns,
-  DynamoArchivedChannelTests,
-  DynamoBanditData,
-  DynamoBannerDesigns,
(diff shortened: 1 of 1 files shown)


app/wiring/AppComponents.scala
```diff
@@ -12,23 +12,10 @@ import play.api.libs.ws.ahc.AhcWSComponents
...
-  DynamoBanditData,
-  DynamoBannerDesigns,
-  DynamoCampaigns,
-  DynamoChannelTests,
-  DynamoChannelTestsAudit,
-  DynamoPermissionsCache,
-  DynamoSuperMode,
-  S3
-}
+import services.{Aws, BigQueryService, CapiService, DynamoArchivedBannerDesigns, DynamoArchivedChannelTests, DynamoBanditData, DynamoBannerDesigns, DynamoCampaigns, DynamoChannelTests, DynamoChannelTestsAudit, DynamoPermissionsCache, DynamoSuperMode, S3}
```
sookburt, line 15: I assume this is just a linting change

PR #785, opened 2025-09-16
Shorter policy for dynamodb permissions

tomrf1:
We're unable to add new permissions to the IAM policy because AWS has a max length. The error in cloudformation is:
`Maximum policy size of 10240 bytes exceeded for role support-admin-console-COD-InstanceRoleAdminconsole-C23CXSH2RPY1 `
See https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html#reference_iam-quotas-entity-length

This happened when I tried to add permissions for new dynamodb tables.

The solution here is to change how we construct dynamodb permissions.
Currently we build two `PolicyDocument`s for each table, using the gu-cdk `GuDynamoDBReadPolicy` and `GuDynamoDBWritePolicy` constructs. Each of these `PolicyDocument`s is quite long, and so we end up with an incredibly long policy.
Instead, this PR introduces a custom `DynamoPolicy` construct which builds a single `PolicyDocument` for _all_ tables. This makes the generated cloudformation much more concise.

### DIFF
--- a/cdk/lib/admin-console.ts
+++ b/cdk/lib/admin-console.ts
@@ -21,7 +21,7 @@ import {
 import type { Policy } from 'aws-cdk-lib/aws-iam';
 import { AccountPrincipal, Role } from 'aws-cdk-lib/aws-iam';
 import { ParameterDataType, ParameterTier, StringParameter } from 'aws-cdk-lib/aws-ssm';
-import {DynamoPolicy} from "./dynamo-policy";
+import {MultiDynamoTableReadPolicy, MultiDynamoTableWritePolicy} from "./dynamo-policy";
 
 export interface AdminConsoleProps extends GuStackProps {
   domainName: string;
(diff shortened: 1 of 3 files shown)


cdk/lib/admin-console.ts
```diff
@@ -269,16 +244,19 @@ export class AdminConsole extends GuStack {
...
+      bannerDesignsDynamoTable,
       archivedBannerDesignsDynamoTable,
-    );
-    const permissionsDynamoPolicies = this.buildDynamoPolicies(permissionsTable);
+      permissionsTable,
+    ]);
+
+    const channelTestsIndexPolicy = new GuDynamoDBReadPolicy(this, `DynamoRead-${channelTestsDynamoTable.node.id}/index/campaignName-name-index`, {
+      tableName: `${channelTestsDynamoTable.tableName}/index/campaignName-name-index`,
+    });
```
tomrf1, line 259: we still have to separately create this policy for an index on the channel-tests table

cdk/lib/dynamo-policy.ts
```diff
@@ -0,0 +1,20 @@
+import type {GuStack} from "@guardian/cdk/lib/constructs/core";
+import {GuAllowPolicy} from "@guardian/cdk/lib/constructs/iam";
+import type {Table} from "aws-cdk-lib/aws-dynamodb";
+
+const readActions = ["BatchGetItem", "GetItem", "Scan", "Query", "GetRecords"];
+const writeActions = ["BatchWriteItem", "PutItem", "DeleteItem", "UpdateItem"];
```
tomrf1, line 6: these permissions are taken from the gu-cdk construct here:
https://github.com/guardian/cdk/blob/main/src/constructs/iam/policies/dynamodb.ts#L31

cdk/lib/admin-console.ts
```diff
@@ -229,31 +229,6 @@ export class AdminConsole extends GuStack {
     return table;
   }
 
-  buildChannelTestsDynamoPolicies(table: Table): GuAllowPolicy[] {
-    return [
-      new GuDynamoDBReadPolicy(this, `DynamoRead-${table.node.id}`, {
-        tableName: table.tableName,
-      }),
-      new GuDynamoDBReadPolicy(this, `DynamoRead-${table.node.id}/index/campaignName-name-index`, {
-        tableName: `${table.tableName}/index/campaignName-name-index`,
```
tomrf1, line 238: this extra policy on the index is redundant as we grant access on `/index/*`

PR #793, opened 2025-10-02
Product catalog service

tomrf1:
This is in preparation for the new promos tool in the RRCP.

When creating a promo the user must define which product rate plans it should apply to.
We use the product catalog to find out which rate plans are available for each product. We store the selected rate plan IDs in the table.

This PR adds scala models for the parts of the product catalog that are relevant.
It also adds a `ProductCatalogCache`, which polls the product catalog API every 5 mins.
This data will eventually be used by the client to know which rate plans are available.

I think there's a case for _not_ defining a server-side model, and passing the entire JSON blob as-is to the client. This would work. But validating and refining the data server-side is more in keeping with the design of this app, and it enables us to catch and log any data issues more formally.

### DIFF
--- a/app/models/promos/ProductCatalog.scala
+++ b/app/models/promos/ProductCatalog.scala
@@ -1,0 +1,57 @@
+package models.promos
+
+case class Pricing(
+  aud: Option[Double] = None,
+  cad: Option[Double] = None,
+  eur: Option[Double] = None,
+  gbp: Option[Double] = None,
+  nzd: Option[Double] = None,
+  usd: Option[Double] = None
+)
+
+case class Charge(id: String)
+case class Charges(items: Map[String, Charge])
+
+case class RatePlan(
+  billingPeriod: String,
+  charges: Charges,
+  id: String,
+  pricing: Pricing,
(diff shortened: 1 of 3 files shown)


app/models/promos/ProductCatalog.scala
```diff
@@ -0,0 +1,141 @@
...
+ *         },
+ *         "billingPeriod": "Annual"
+ *       }
+ *     }
+ *   },
+ *   ...
+ * }
+ */
+
+case class ProductCatalog(
```
tomrf1, line 45: The products defined here by the catalog do not precisely map to the set of products available in the tool.
For example, `Newspaper` in the tool will encompass `HomeDelivery` + `SubscriptionCard` + `NationalDelivery` from the catalog.

rupertbates: We also need Digital Subscription in this list

tomrf1: good spot!

rupertbates (approved):
Looks good once digisub is added 👍

PR #1108, opened 2026-04-15
Hides the hero image section and validation for now

sookburt:
## What does this change?

We are not currently in a position to be able to implement the hero image in support-frontend and it is not likely to be needed in the short term, so we are hiding it in RRCP for the moment in order to enable MRR to be able to start using the  tooling in the meantime.

This has been deployed to CODE, created a test/variant combo and checked that the URL created behaved as expected.  Also checked that there was no impact on the existing UTS and student beans pages.

## Images

Screenshot 2026-04-15 at 17 26 37

### DIFF
--- a/public/src/components/channelManagement/studentLandingPage/utils/defaults.ts
+++ b/public/src/components/channelManagement/studentLandingPage/utils/defaults.ts
@@ -32,10 +32,13 @@ export const getDefaultInstitution = () => {
 };
 
 const DEFAULT_IMAGE: ResponsiveImage = {
-  mobileUrl: '',
-  tabletUrl: '',
-  desktopUrl: '',
-  altText: '',
+  mobileUrl:
+    'https://i.guim.co.uk/img/media/811f456e9786d119d766e55f2df821c056d415b0/0_0_2588_1276/master/2588.jpg',
+  tabletUrl:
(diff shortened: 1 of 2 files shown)


public/src/components/channelManagement/studentLandingPage/utils/defaults.ts
```diff
@@ -32,10 +32,13 @@ export const getDefaultInstitution = () => {
 };
 
 const DEFAULT_IMAGE: ResponsiveImage = {
-  mobileUrl: '',
-  tabletUrl: '',
-  desktopUrl: '',
-  altText: '',
+  mobileUrl:
+    'https://i.guim.co.uk/img/media/811f456e9786d119d766e55f2df821c056d415b0/0_0_2588_1276/master/2588.jpg',
```
sookburt, line 36: Using the existing image URLs for now.

public/src/components/channelManagement/studentLandingPage/variantEditor.tsx
```diff
@@ -87,7 +90,7 @@ export const VariantEditor: React.FC<StudentLandingPageVariantEditorProps> = ({
   const [sectionValidity, setSectionValidity] = useState<VariantSectionValidity>({
     copy: false,
     institution: false,
-    image: false,
+    image: true, // change this back to false when we add the hero image bits back in.
```
sookburt, line 93: Otherwise the save fails without useful feedback about where the problem exists.

tomrf1 (approved):

PR #1129, opened 2026-04-24
fix ESLint violations - part 002

juabara:
- Fix some import ordering
- Refactor state initialization to avoid setting state in useEffect


sookburt (commented):
When running pnpm lint, I'm getting the following warning:
```
/Users/_/code/support-admin-console/public/src/components/channelExclusions/useExclusionRuleHandlers.ts
  58:20  warning  Compilation Skipped: Use of incompatible library

This API returns functions which cannot be memoized without leading to stale UI. To prevent this, by default React Compiler will skip memoizing this component/hook. However, you may see issues if values from this API are passed to other components/hooks that are memoized.

/Users/_/code/support-admin-console/public/src/components/channelExclusions/useExclusionRuleHandlers.ts:58:20
  56 |     defaultValues: { rule },
  57 |   });
> 58 |   const formRule = watch('rule');
     |                    ^^^^^ React Hook Form's `useForm()` API returns a `watch()` function which cannot be memoized safely.
  59 |
  60 |   useEffect(() => {
  61 |     const serverRuleMatchesLastSaved = rulesAreEqual(rule, savedRule);  react-hooks/incompatible-library

✖ 1 problem (0 errors, 1 warning)
```

PR #1186, opened 2026-05-26
Enforce article count restriction for mParticle audience targeting of signed-out users in banner tests

juabara:
When a banner test targets an mParticle audience and can target signed-out users, it must have a minimum article count of 5 views. This prevents excessive mParticle traffic for low-engagement browsers.

Changes:
- Auto-enforce `articlesViewedSettings` with `minViews=5` when mParticle audience is set and `signedInStatus !== 'SignedIn'`
- Add validation to block saving if article count is manually reduced below 5 when the constraint applies
- Display error message below mParticle audience ID field when constraint is violated
- Display informational note in Article count section when constraint is active
- Fix bug where "Do not target by user's article count" wouldn't persist due to template-based re-calculation

### DIFF
--- a/public/src/components/channelManagement/bannerTests/bannerTestEditor.tsx
+++ b/public/src/components/channelManagement/bannerTests/bannerTestEditor.tsx
@@ -1,4 +1,4 @@
-import { Typography } from '@mui/material';
+import { Alert, Typography } from '@mui/material';
 import React, { useEffect, useRef, useState } from 'react';
 import {
   BannerContent,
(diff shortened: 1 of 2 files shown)


public/src/components/channelManagement/bannerTests/bannerTestEditor.tsx
```diff
@@ -155,7 +155,19 @@ const BannerTestEditor: React.FC<ValidatedTestEditorProps<BannerTest>> = ({
   };
 
   const onSignedInStatusChange = (signedInStatus: SignedInStatus): void => {
-    onTestChange((current) => ({ ...current, signedInStatus }));
+    onTestChange((current) => {
+      const updatedTest = { ...current, signedInStatus };
+      if (current.mParticleAudience !== undefined && signedInStatus !== 'SignedIn') {
+        const minViews = current.articlesViewedSettings?.minViews ?? 0;
+        if (!current.articlesViewedSettings || minViews < 5) {
+          const enforced = current.articlesViewedSettings
```
tomrf1: why do we need this logic? The validation prevents the user from saving

tomrf1 (approved):

PR #1207, opened 2026-06-11
remirror empty value fix and console warning for multiselect fix

AnastasiiaBalenko:
[Ticket link](https://app.asana.com/1/1210045093164357/project/1213309681835463/task/1215499216707078)
## What does this change?
This change fixes the bug with <br> tag instead of empty string in RichTextEditor component.
Also it fixes the console warning from MUI (see screenshot)

Deployed to CODE. Fields that use RichTextEditor were updated to empty string and the saves data was verified

## Images
**BEFORE**
Screenshot 2026-06-08 at 15 49 55
Screenshot 2026-06-12 at 09 53 25

## Accessibility

### DIFF
--- a/public/src/components/channelManagement/MutliSelectTagEditor.tsx
+++ b/public/src/components/channelManagement/MutliSelectTagEditor.tsx
@@ -70,6 +70,7 @@ const MultiselectAutocomplete: React.FC<MultiselectAutocompleteProps> = ({
         disabled={disabled}
         options={options}
         getOptionLabel={(option) => option.label}
+        isOptionEqualToValue={(option, value) => option.value === value.value}
         value={tagIds?.map<Option>((tag) => ({ label: tag, value: tag }))}
         inputValue={inputValue}
         componentsProps={{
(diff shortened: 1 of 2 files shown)


public/src/components/channelManagement/MutliSelectTagEditor.tsx
```diff
@@ -70,6 +70,7 @@ const MultiselectAutocomplete: React.FC<MultiselectAutocompleteProps> = ({
         disabled={disabled}
         options={options}
         getOptionLabel={(option) => option.label}
+        isOptionEqualToValue={(option, value) => option.value === value.value}
```
tomrf1, line 73: why do we need this?

AnastasiiaBalenko: There was the console warning from MUI: The value provided to Autocomplete is invalid. On each render the value is a new object so event if option and value are the same, they are still not the same object. This line checks if option and value are actually equal and eliminates the console warning
Screenshot 2026-06-12 at 09 53 25

tomrf1 (approved):

PR #1222, opened 2026-06-18
Re-authenticate when session expires

tomrf1:
## The problem
If the page is left open in the browser for too long then the session expires. When the page makes an API call (e.g. when the user clicks save) the server redirects for re-authentication, which fails.
The user sees:
Screenshot 2026-06-18 at 15 31 50
And the network calls look like this:
Screenshot 2026-06-18 at 15 32 20

The [advice](https://github.com/guardian/play-googleauth#caveats) in the play-googleauth library is to use an iframe to trigger a reauth. But this fails when a request is made to google. Claude suggested this is because google prevents requests from an iframe, and proposed instead doing it in a popup window.

## Solution
The client periodically checks if a reauth is needed by hitting a new `/isValid` endpoint. If 419 is returned then the client hits `/reauth` in a popup window. This should immediately close when the google request completes, and the client can continue with any API calls.

## Testing
Tested by deleting the `PLAY_SESSION` cookie, waiting 1 minute, then clicking save.
Previously this would trigger the error, but now it triggers the reauth and then succeeds:
Screenshot 2026-06-18 at 15 50 57

The popup window opens and closes instantly.


juabara (approved):

AnastasiiaBalenko (approved):
👍

PR #1239, opened 2026-06-30
get LP tests for destination for recurring products

AnastasiiaBalenko:
## What does this change?
This change updates destinationTests options for recurring products in Choice cards (all products except OneOff products)
Previously if "checkout" destination was selected, tests for One-time-checkout were shown up in destinationTests options. But it is not correct for recurring products.

Deployed to CODE and verified that correct tests are shown up for correct products.

## Images
Screenshot 2026-07-01 at 07 16 20

## Accessibility

### DIFF
--- a/public/src/components/channelManagement/choiceCards/ChoiceCardDestinationFields.tsx
+++ b/public/src/components/channelManagement/choiceCards/ChoiceCardDestinationFields.tsx
@@ -101,6 +101,14 @@ export const ChoiceCardDestinationFields: React.FC<ChoiceCardDestinationFieldsPr
     [clearVariantSelection],
   );
 
+  const getDestination = useCallback(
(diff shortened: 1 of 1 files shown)


public/src/components/channelManagement/choiceCards/ChoiceCardDestinationFields.tsx
```diff
@@ -101,6 +101,14 @@ export const ChoiceCardDestinationFields: React.FC<ChoiceCardDestinationFieldsPr
     [clearVariantSelection],
   );
 
+  const getDestination = useCallback(
```
tomrf1, line 105: it would be helpful to have a comment here explaining why

tomrf1 (approved):

PR #1273, opened 2026-07-24
Remove jackson overrides

tomrf1:
scala steward wants to bump the jackson version, but actually our current jackson overrides are no longer necessary.
Confirmed by running `sbt dependencyTree` without the override, and checking that we're getting at least v2.21.4.


tomrf1:
actually, not all jackson deps are covered

PR #1290, opened 2026-07-31
Return 404 if item not found in DynamoDB

AnastasiiaBalenko:
[Ticket link](https://app.asana.com/1/1210045093164357/project/1213309681835463/task/1217032428374914)
## What does this change?
This change updated code to return 404 instead of 500 when promo code is not found

When a promo code doesn't exist in DynamoDB, the GET /promo/:promoCode endpoint was returning a 500 and logging a spurious InternalServerError alert. This was caused by the not-found case falling through to the generic error handler in run().

Side effect: other services that use DynamoService.get (DynamoBannerDesigns, DynamoChannelTests, DynamoPromoCampaigns) are functionally unchanged (still 500 on not-found) but their not-found log messages are now cleaner — single-wrapped instead of double-wrapped.

Deployed to CODE env

## Images
Screenshot 2026-07-31 at 15 27 09

## Accessibility

### DIFF
--- a/app/controllers/promos/PromosController.scala
+++ b/app/controllers/promos/PromosController.scala
@@ -3,7 +3,7 @@ package controllers.promos
 import actions.{AuthAndPermissionActions, PermissionAction}
 import com.typesafe.scalalogging.LazyLogging
 import io.circe.syntax._
-import models.DynamoErrors.{DynamoDuplicateNameError, DynamoNoLockError}
+import models.DynamoErrors.{DynamoDuplicateNameError, DynamoGetError, DynamoNoLockError, DynamoNotFoundError}
 import models.promos.Promo
 import play.api.libs.circe.Circe
 import play.api.mvc.{AbstractController, Action, ActionBuilder, AnyContent, ControllerComponents, Result}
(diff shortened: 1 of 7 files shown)


juabara (approved):

app/controllers/promos/PromosController.scala
```diff
@@ -67,6 +67,7 @@ class PromosController(
           ).asJson
           Ok(noNulls(response))
         }
+        .catchSome { case DynamoGetError(_: DynamoNotFoundError) => ZIO.succeed(NotFound(s"Promo $promoCode not found")) }
```
tomrf1: should this be `case DynamoNotFoundError(_) =>`?
why is it wrapped in a `DynamoGetError`?

tomrf1: ah I think I see why - in DynamoService it's followed by
`.mapError(error => DynamoGetError(error))`
and we end up with nested `DynamoError`s

tomrf1: An idea for improving this in DynamoService.get -
We'd need to change the return error type to `DynamoError`, but I think this might be neater:
```
attemptBlocking {
    ...
    }.mapError(DynamoGetError(_))
      .flatMap {
        case Some(item) => ZIO.succeed(item)
        case None       => ZIO.fail(DynamoNotFoundError(query.keyConditionExpression()))
      }
```

This first handles any error from the `attemptBlocking` aws call in `mapError`, then handles the possibility of the item not being found. This would mean that in PromosController we can simply match on `case DynamoNotFoundError(_) =>`

tomrf1 (approved):

PR #1308, opened 2026-08-07
Reduce webpack bundle via source map removal and code splitting

juabara:
## Summary

- Remove `eval-cheap-module-source-map` devtool from `webpack.prod.js` — this was a dev-only setting that embedded full source maps in the production bundle, accounting for ~18MB of the 20MB bundle
- Implement route-level code splitting using `React.lazy` + dynamic `import()` with a `<Suspense>` boundary, so each page loads its JS on demand instead of bundling all 25+ routes into one file
- Change `tsconfig.json` `module` from `es6` to `esnext` to enable dynamic `import()` syntax
- Add `chunkFilename` and `publicPath: '/assets/build/'` to webpack configs so async chunks are emitted and served correctly via Play's asset pipeline

## Results

| Metric | Before | After |
|--------|--------|-------|
| Main bundle | 20.4 MB | 389 KiB |
| Initial load | 20.4 MB (everything) | 389 KiB + 1 route chunk |
| Total build | 20.4 MB | 2.9 MB (64 chunks) |
| Webpack warnings | 3 | 2 |

The initial page load is now **98% smaller**. Users only download the JS for the page they're visiting, with the rest loaded on navigation.

### DIFF
--- a/public/src/main.tsx
+++ b/public/src/main.tsx
@@ -6,44 +6,99 @@ import { Theme, ThemeProvider } from '@mui/material/styles';
 import Toolbar from '@mui/material/Toolbar';
 import Typography from '@mui/material/Typography';
 import { makeStyles, ThemeProvider as StylesThemeProvider } from '@mui/styles';
-import React from 'react';
+import React, { lazy, Suspense } from 'react';
 import { createRoot } from 'react-dom/client';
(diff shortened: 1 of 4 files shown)


AnastasiiaBalenko (approved):
👍

